PR #282, opened 2021-09-14
Don't load remote plugins when destination is disabled

danieljackins:
image

This PR avoids loading remote plugins when the corresponding destination action has been disabled, to prevent errors like above/save time loading plugins.

## Testing

Tested locally:
image

No more braze errors/no braze plugins in the queue when braze is disabled.

### DIFF
--- a/src/plugins/remote-loader/index.ts
+++ b/src/plugins/remote-loader/index.ts
@@ -5,6 +5,8 @@ import { asPromise } from '../../lib/as-promise'
 import { loadScript } from '../../lib/load-script'
 
 export interface RemotePlugin {
+  /** The name of the action destination. */
+  name: string
   /** The url of the javascript file to load */
   url: string
   /** The UMD/global name the plugin uses. Plugins are expected to exist here with the `PluginFactory` method signature */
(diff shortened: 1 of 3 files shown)


src/plugins/remote-loader/index.ts
```diff
@@ -5,6 +5,8 @@ import { asPromise } from '../../lib/as-promise'
 import { loadScript } from '../../lib/load-script'
 
 export interface RemotePlugin {
+  /** The name of the action destination. */
```
nettofarah, line 8: Action destination is one type of remote plugin, this change creates a contract between the two that will likely make it harder for us to evolve the architecture of RemotePlugins

What happens if we want to have edge functions as remote plugins?

src/plugins/remote-loader/index.ts
```diff
@@ -42,27 +44,28 @@ export async function remoteLoader(
   settings: LegacySettings
 ): Promise<Plugin[]> {
   const allPlugins: Plugin[] = []
-
   const pluginPromises = (settings.remotePlugins ?? []).map(
     async (remotePlugin) => {
-      try {
-        await loadScript(remotePlugin.url)
+      if (settings.integrations[remotePlugin.name]) {
```
nettofarah, line 49: This creates an implicit dependency between `settings.integrations` and `remotePlugins`, which should not be the case. We can't guarantee that remote plugins have an integrations object counterpart because remote plugins are not exclusively destinations

nettofarah (changes requested):
The implementation looks good, but I'm not sure this is the direction we want to take with remote plugins

mshwery:
@danieljackins curious where this is coming from – only enabled destination's remote plugins should be included by ajs-renderer in the cdn settings.

PR #283, opened 2021-09-23
Ensure document is loaded before appending (IE 11)

danieljackins:
This PR makes sure the document is loaded before appending script to the body, which in rare cases was causing an error in IE 11.

## Testing

Testing was done on a windows computer with IE 11, the error no longer appears and a.js loads properly

### DIFF
--- a/src/standalone.ts
+++ b/src/standalone.ts
@@ -60,7 +60,14 @@ if (shouldPolyfill()) {
     'src',
     'https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.7.0/polyfill.min.js'
   )
-  document.body.appendChild(script)
+
+  if (document.readyState === 'loading') {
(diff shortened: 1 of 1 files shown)


src/standalone.ts
```diff
@@ -60,7 +60,14 @@ if (shouldPolyfill()) {
...
   )
-  document.body.appendChild(script)
+
+  if (document.readyState === 'loading') {
+    document.addEventListener('DOMContentLoaded', () =>
+      document.body.appendChild(script)
+    )
+  } else {
+    document.body.appendChild(script)
+  }
```
juliofarah, line 70: is there a difference if we just do  this?
```js
    document.addEventListener('DOMContentLoaded', () =>
      document.body.appendChild(script)
    )
```

nettofarah: Yeah, the event does not trigger retroactively

nettofarah (approved):

PR #284, opened 2021-09-24
Allow integrations that exist as an object

danieljackins:
AJS classic sent events to integrations that weren't explicitly true but existed in the integrations object when All: false is set. For example,
`integrations: {
  All: false,
  Intercom: {
    user_hash: 'foo'
  }
}`

should send data to Intercom despite it not being set to true.

## Testing

Data being sent to intercom when using the above integrations format:
image


src/core/queue/__tests__/event-queue.test.ts
```diff
@@ -422,6 +422,35 @@ describe('Flushing', () => {
...
+
+      const evt = {
+        type: 'track' as const,
+        integrations: {
+          All: false,
+          Amplitude: {
+            amplitudeKey: 'foo',
+          },
+        },
+      }
```
juliofarah, line 438: do we need to update any type definitions?

danieljackins: i don't believe so

juliofarah: Just checked that `Integrations` already accepts `JSONObjects` https://github.com/segmentio/analytics-next/blob/c13a3053afb8f2f2ce96e6d617a357a1cea83c1d/src/core/events/interfaces.ts#L11-L12

src/core/queue/event-queue.ts
```diff
@@ -214,7 +214,7 @@ export class EventQueue extends Emitter {
     const available =
       denyList.All === false
         ? this.plugins.filter(
-            (p) => denyList[p.name] === true || p.type !== 'destination'
+            (p) => denyList[p.name] || p.type !== 'destination'
```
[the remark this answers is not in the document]
danieljackins: I think this would make any unspecified integration satisfy the `||`

juliofarah (approved):

PR #286, opened 2021-09-29
Respect anonymousId being set in events

danieljackins:
This PR makes the segment.io normalize function respect anonymousId being set in events, as outlined in our docs here https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/#override-the-anonymous-id-using-the-options-object

## Testing

image
image


nettofarah (approved):

pooyaj (approved):
🙌

PR #290, opened 2021-10-06
Allow QueryString to parse hashed urls

juliofarah:
We realized that Analytics JS did not consider the parameters for urls like `https://localhost/#/?ajs_id=123` (see the `#` symbol). Although this is not necessarily a bug (as values present after the `#` symbol are not fundamentally query parameters), we have customers passing URLs like those and want to support broader cases.

This PR allows AnalyticsBrowser to parse query parameters passed into the `window.location.hash` object.

## Testing
Screen Shot 2021-10-05 at 5 52 41 PM
Screen Shot 2021-10-05 at 5 53 23 PM


danieljackins (approved):

src/browser.ts
```diff
@@ -224,8 +224,13 @@ export class AnalyticsBrowser {
       analytics.page().catch(console.error)
     }
 
-    if (window.location.search.includes('ajs_')) {
-      analytics.queryString(window.location.search).catch(console.error)
+    const search = window.location.search ?? ''
+    const hash = window.location.hash ?? ''
+
+    const term = search.length ? search : hash.replace(/#\//, '')
```
pooyaj, line 230: is the replacement needed? If so, do we need to cover something like this: `www.test.com/#about?id=123` ?

juliofarah: for this example: `http://localhost:3000/#/?ajs_id=123`, the replacement is needed so we have the same format for both search and hash (`?` + `key` + `=` + `val`).

for `www.test.com/#about?id=123` we'd need to remove everything that's after the `#` and before the `?`

I can rework the regex to cover that

juliofarah: this `/(?=#).*(?=\?)/` should do the trick

src/__tests__/integration.test.ts
```diff
@@ -858,3 +858,85 @@ describe('.Integrations', () => {
     `)
   })
 })
+
+describe('queryString', () => {
```
nettofarah, line 862: This file is getting **very** large.
How about you create a `query-string.integration.test.ts`?

PR #291, opened 2021-10-06
Allow Segment.io to retry offline calls

juliofarah:
In order to accomplish parity with AJS Classic, we need to retry failed `segment.io` events (for both TAPI fails and events fired while offline). This PR allows the Segment.io integration to retry calls.

## Testing
Single events:
https://user-images.githubusercontent.com/484013/136282914-c50069ba-721f-4d53-87c5-e88f30f471d3.mov

Batched events:
https://user-images.githubusercontent.com/484013/137005683-de426aa7-21ca-4569-8080-f97319700184.mov

### DIFF
--- a/src/plugins/segmentio/__tests__/retries.test.ts
+++ b/src/plugins/segmentio/__tests__/retries.test.ts
@@ -37,16 +37,12 @@ describe('Segment.io retries', () => {
     await analytics.register(segment, pageEnrichment)
   })
 
-  test('buffers offline events', async () => {
-    await analytics.track('event')
-    expect(scheduleFlush).toHaveBeenCalled()
-  })
-
   test('add events to the queue', async () => {
     jest.spyOn(queue, 'push')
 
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "23.5 KB"
+      "limit": "29.08 KB"
```
nettofarah, line 23: :rip:

src/plugins/segmentio/index.ts
```diff
@@ -63,6 +65,32 @@ export function segmentio(
       ? batch(apiHost, settings?.deliveryStrategy?.config)
       : standard()
 
+  async function flushQueue(xt: Plugin, queue: PriorityQueue<Context>) {
```
nettofarah, line 68: Make sure to test this behavior with batching enabled.
I'm curious to see what happens if you push 1000 events into the queue while offline

Issue #293, opened 2021-10-13
ajs_anonymous_id can't be read from local storage using JSON.parse


viktor-edgardcooper:
This package seems to throw an error when reading the ajs_anonymous_id from local storage.
See screenshots below which triggers the error.

Im using a next project (^11.1.2) with the lastest version of this package (^1.30.0)

Local Storage on initial load

![image](https://user-images.githubusercontent.com/92312223/137104621-546e460d-95cf-46ce-b3e0-58357f5364c8.png)

Cookies on initial load

![image](https://user-images.githubusercontent.com/92312223/137104733-1450b11f-c55e-4383-8f20-2ee416cb92c9.png)

Local Storage on refresh

![image](https://user-images.githubusercontent.com/92312223/137104889-76d7c375-7196-4a9e-9dad-e15c527a8080.png)

Which results in the following error

![image](https://user-images.githubusercontent.com/92312223/137105270-5c2d2532-4b92-4ca8-b52f-9212fccf3721.png)
![image](https://user-images.githubusercontent.com/92312223/137105411-26d231b9-8a83-4414-8e45-41b873c78e31.png)

danieljackins:
Hey @viktor-edgardcooper, thanks for opening the issue.

I'm having trouble replicating this, would you be able to share a link or some reproduction steps? It looks like your localstorage anonymous_id is losing its quotations, somehow

viktor-edgardcooper:
@danieljackins after a long debugging session we were able to find what was causing the ajs_anonymous_id to lose its quotations.

We are using Amplify for authentication and we rely on SSR for some pages.

`Amplify.configure({ ...config, ssr: true });`
[see docs](https://docs.amplify.aws/lib/ssr/q/platform/js/)

By providing `ssr: true`, Amplify persists credentials on the client in cookies so that subsequent requests to the server have access to them. In this process the ajs_anonymous_id loses its quotations.

Wouldn't it be better if Segment didn't rely on a localstorage anonymous_id with quotations, and just rely on the same value as the cookie. The value without quotations?

danieljackins:
Fixed! Thanks @viktor-edgardcooper !

PR #296, opened 2021-10-14
Read non-stringified localstorage values

danieljackins:
This PR makes sure we're able to read values from localstorage that weren't stringified. See https://github.com/segmentio/analytics-next/issues/293 for more context.

## Testing

Added unit test


mmckhalidmmc87-stack:
My account money premier bank

PR #300, opened 2021-10-21
Allow updateEvent to override integrations set to true

danieljackins:
This PR allows integrations that are set to true in the integrations object to be updated with integration specific settings.

## Testing

- Updated unit tests
- Loading analytics with Amplitude set to true:
image

- Loading analytics with Amplitude set to false:
image

Please, make sure to describe how you tested this change, include any gifs or screenshots you find necessary.

### DIFF
--- a/src/core/context/index.ts
+++ b/src/core/context/index.ts
@@ -97,6 +97,17 @@ export class Context implements AbstractContext {
   }
 
   public updateEvent(path: string, val: unknown): SegmentEvent {
+    // Allow integrations that are set to true to be overridden with integration settings.
+    // For example, allow {..., integrations: {Amplitude: true}} to be overwritten with
+    // {..., integrations: {Amplitude: {sessionId: 'foo'}}}.
+    if (path.split('.')[0] === 'integrations') {
+      const integrationName = path.split('.')[1]
+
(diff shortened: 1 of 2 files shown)


src/core/context/index.ts
```diff
@@ -97,6 +97,17 @@ export class Context implements AbstractContext {
...
 
   public updateEvent(path: string, val: unknown): SegmentEvent {
+    // Allow integrations that are set to true to be overridden with integration settings.
+    // For example, allow {..., integrations: {Amplitude: true}} to be overwritten with
+    // {..., integrations: {Amplitude: {sessionId: 'foo'}}}.
+    if (path.split('.')[0] === 'integrations') {
+      const integrationName = path.split('.')[1]
+
+      if (this._event.integrations?.[integrationName] === true) {
+        delete this._event.integrations?.[integrationName]
```
pooyaj, line 107: 🤔 one quick thing to check here. Looks like the [latest version ](https://github.com/lukeed/dset/releases/tag/v3.0.0) of `dset` now allows overriding existing non-object values. So in this case we don't have to delete the nested value before calling `dset`.

pooyaj: Can we check if upgrading `dset` works here without breaking any of the existing tests?

PR #307, opened 2021-10-29
fix args resolver for nulls

danieljackins:
Similar to https://github.com/segmentio/analytics-next/pull/302, this PR handles null input for identify and group calls, and properly sets traits instead of in the context.

## Testing

image

### DIFF
--- a/src/core/arguments-resolver/index.ts
+++ b/src/core/arguments-resolver/index.ts
@@ -87,12 +87,15 @@ export function resolvePageArguments(
 
 export const resolveUserArguments = (user: User): ResolveUser => {
   return (...args): ReturnType<ResolveUser> => {
-    const id =
-      args.find(isString) ?? args.find(isNumber)?.toString() ?? user.id()
+    let id: string | ID | null = null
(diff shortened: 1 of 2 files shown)


src/core/arguments-resolver/index.ts
```diff
@@ -87,12 +87,15 @@ export function resolvePageArguments(
...
-      args.find(isString) ?? args.find(isNumber)?.toString() ?? user.id()
-
-    const objects = args.filter(
-      (obj) => isPlainObject(obj) || obj === null
-    ) as Array<object | null>
+    let id: string | ID | null = null
+    id = args.find(isString) ?? args.find(isNumber)?.toString() ?? user.id()
+
+    const objects = args.filter((obj) => {
+      if (id === null) {
```
juliofarah, line 94: do we need to worry about empty strings here?

juliofarah (approved):

src/core/arguments-resolver/index.ts
danieljackins: empty strings shouldn't be a problem

PR #313, opened 2021-11-03
Ensure source middleware is added first in buffered operations

danieljackins:
This PR ensures that source middleware is added before other buffered operations are ran - for example on the initial page call in the segment snippet.

## Testing

MW ran on snippet page call: 
image

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -85,7 +85,7 @@ function hasLegacyDestinations(settings: LegacySettings): boolean {
   )
 }
 
-function flushBuffered(analytics: Analytics): void {
+async function flushBuffered(analytics: Analytics): Promise<void> {
   const wa = window.analytics
   const buffered =
     // @ts-expect-error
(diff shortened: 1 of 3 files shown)


src/browser.ts
```diff
@@ -98,11 +98,16 @@ function flushBuffered(analytics: Analytics): void {
       // @ts-expect-error
       typeof analytics[operation] === 'function'
     ) {
-      // flush each individual event as its own task, so not to block initial page loads
-      setTimeout(() => {
+      if (operation === 'addSourceMiddleware') {
         // @ts-expect-error
-        analytics[operation].call(analytics, ...args)
-      }, 0)
+        await analytics[operation].call(analytics, ...args)
```
pooyaj, line 103: Just for simplicity, I wonder if this still works fine without using async/await. For example, just run `addSourceMiddleware` without the `setTimeout`, and then let the rest of operations run async in their own task.

pooyaj (commented):
Also possible to write a test to verify this?

src/browser.ts
danieljackins: That was my first thought as well but calling addSourceMiddleware without the setTimeout was still resulting in the page call being fired first 🤔

pooyaj: Oh, I see what's going on. Since we have a separate chunk for the `middleware` plugin, then the `analytics.addSourceMiddleware` happens asynchronously, because it has to lazy-load the middleware first.

pooyaj (approved):
I think this is the best implementation given the lazy loading of the middleware!

PR #318, opened 2021-11-07
refactor: change persisted storage constant to function to avoid execution on time of import

peeyushsingla1:
**Description :-**
- We are building an application that is hosted as an iframe with in another application.
- We would like to check if cookie access is enabled and then call `await AnalyticsBrowser.load({ writeKey })` but just importing AnalyticsBrowser crashes the application.

**Steps to reproduce:-**
- `git clone https://github.com/peeyushsingla1/testing-segment-analytics`
- `yarn && yarn start`
- Block all cookies from chrome(as our application lives in iframe setting block third party cookies caused this issue).

Screenshot 2021-11-07 at 12 16 02

- Visit application and it would break
Screenshot 2021-11-07 at 12 09 17

**Technical Details:-**
* Issue is happening because of [persisted](https://github.com/segmentio/analytics-next/blob/4d04b4813dd7ba539d86c31b83ccb1f96dd08d21/src/lib/priority-queue/persisted.ts#L5)

## Testing

* Its internal implementation of persisted so I guess no tests required.
* One problem after the refactoring is below test start to fail, unsure why

```
 FAIL  src/plugins/legacy-video-plugins/__tests__/index.test.ts
  ● loadLegacyVideoPlugins › attaches video plugins to ajs

    SecurityError: localStorage is not available for opaque origins
```

### DIFF
--- a/src/lib/priority-queue/persisted.ts
+++ b/src/lib/priority-queue/persisted.ts
@@ -2,13 +2,19 @@ import { PriorityQueue } from '.'
 import { Context, SerializedContext } from '../../core/context'
 import { isBrowser } from '../../core/environment'
 
-const loc = isBrowser()
-  ? window.localStorage
-  : {
-      getItem() {},
-      setItem() {},
-      removeItem() {},
-    }
+let loc:
+  | Storage
+  | { getItem: () => void; setItem: () => void; removeItem: () => void } = {
+  getItem() {},
+  setItem() {},
+  removeItem() {},
+}
+
+try {
+  loc = isBrowser() ? window.localStorage : loc
+} catch (err) {
+  console.warn('Unable to access localStorage', err)
+}
 
 function persisted(key: string): Context[] {
   const items = loc.getItem(key)


pooyaj (changes requested):
@peeyushsingla1 thanks for the PR! Looks like as you pointed out a test is failing. Did you narrow down the original exception to accessing the `window.localStorage` ? What do you think of modifying the solution so we wrap the localStorage access in a `try` `catch`, and default to the stub object if localStorage is not accessible? 

Something along this line: 

```typescript
let loc:
  | Storage
  | { getItem: () => void; setItem: () => void; removeItem: () => void } = {
  getItem() {},
  setItem() {},
  removeItem() {},
}

try {
  loc = isBrowser() ? window.localStorage : loc
} catch (err) {
  console.warn('Unable to access localStorage', err)
}
```

peeyushsingla1:
@pooyaj Good suggestion updated PR with the suggested code. tests are passing as well.

pooyaj (approved):
Looks great, thank you!

peeyushsingla1:
@pooyaj thanks a lot for merging this, Wanted to ask if there would be a release with this fix soon so we can use newer version?

pooyaj:
Just published `1.31.1` 🎉

Issue #303, opened 2021-10-28
process is not defined


gmathieu:
Hello,

Leveraging the example in the README raises the following error with v1.30.0:

![image](https://user-images.githubusercontent.com/7855/139306172-eb4782fb-9737-4d62-9ac8-720bcfab76f2.png)

It looks like `process.env` variables aren't being replaced during the `pkg` build: https://github.com/segmentio/analytics-next/blob/35bd0501a9130645802e953b03ec5414e83a1138/webpack.config.js#L16-L21

Great job on the new library! I'm working on replacing our use of https://github.com/segmentio/analytics.js-core with it. 

Thanks

Tip: you might also want to add `files: ["dist/", "src/"]` to `package.json` so only runtime files are published to NPM (see [files](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files)) as opposed to the whole repo
![image](https://user-images.githubusercontent.com/7855/139308233-d1a74a25-a621-44cc-a6f9-6c2ce39b50c1.png)

danieljackins:
Hey @gmathieu thanks for opening the issue!

I'm having trouble replicating this, do you have any further info or replication steps? (following the example in the README works for me)

gmathieu:
Hi @danieljackins,

It's a tricky one. Here's a sandbox with the error https://codesandbox.io/s/webpack-segment-analytics-next-5y33h?file=/src/App.tsx

![image](https://user-images.githubusercontent.com/7855/139508209-5fcbc561-bd49-47c1-b700-35953d175c73.png)
Note: Codesandbox doesn't capture these types of errors so you'll have to open the native console to see it.

Here's what I've been able to gather:
- this only affects `dist/pkg` because it is compiled using `tsc`. As opposed to `dist/umd` which is bundled using Webpack
  - For example, `process.env.VERSION` should be replaced with `"1.30.0"` in `dist/pkg/plugins/segmentio/normalize.js`
- #305 partially fixe the issue for bundlers that leverage CommonJS but ESM will likely be affected
- big assumption: the e2e tests run a local web server which leverages the same Webpack config. Because that config defines the same `EnvironmentPlugin`, all `process.env.*` were replaced with proper values

I hope this helps 🤞

danieljackins:
Hey @gmathieu sorry for the delay and thanks for the PR. We addressed this issue with https://github.com/segmentio/analytics-next/pull/331, as we wanted to avoid adding another bundler.

[...]

PR #319, opened 2021-11-07
Procces env undefined

gmathieu:
This PR addresses the issue in #303 by introducing [Rollup.js](https://rollupjs.org/) to compile CommonJs and ESM bundles.

### Approach

Previously, `dist/cjs` and `dist/pkg` were compiled using `tsc` leaving `process.env.*` in the bundle untouched. `process.env` is not defined in the browser which throws when importing `@segment/analytics-next`. I needed to find a bundler that supports injecting global variables, supported declaration files, and preserved the file structure (Webpack does not support the latter). Rollup seemed like the obvious choice.

My primary goal was to change as little of the source as possible (ideally none). But in order to get Rollup working, I had to upgrade TypeScript which added a few more dependencies:
1. upgrade `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin`
1. upgrade `eslint@7` (note: a lot of plugins don't support eslint@8 yet)
1. disable newly introduced rules (again, to keep the source unchanged)
1. deal with deprecated rule `@typescript-eslint/camelcase`
1. add `useUnknownInCatchVariables: true` to `tsconfig.json` to preserve existing behavior

### Global variables

Here's the list of variables being replaced. Webpack's values are meant to bundle the hosted version (https://cdn.segment.com/analytics-next) so these values are slightly different. I kept the behavior as `webpack.EnvironmentPlugin` to help with local development (not sure it's needed though).

```js
{
ASSET_PATH: '', // removes Webpack's public path dynamic assignment (which would break other Webpack builds)
DEBUG: false, // only used in test so doesn't matter
LEGACY_INTEGRATIONS_PATH: '', // leverage default in src
NODE_ENV: 'production', // only one if statement which checks for `!== 'test'`
VERSION: version, // imported from package.json
}
```
The code is then minified so any dead code is removed (notably  `if (process.env.ASSET_PATH) {...}`).

### How was this missed?

I was surprised to find that `yarn run-example` imports `dist/pkg` (the ESM bundle) directly but never reported any `process is not defined`. I believe Next.js automatically replaces `process.env.*` with an object like previous versions of Webpack did (see [environment](https://nextjs.org/docs/api-reference/edge-runtime#environment)).

## Testing

- I ran as many `make` commands as I could (a lot of QA ones require env vars I don't have access to). This led me to fix a couple things (see 47e4e767834f22426448589877ffc613864745d1, ab7448bf5deebf4ce99c304110dfd6d1cc5d6fd7)
- I used [yalc](https://github.com/wclr/yalc) to include a local build in our web-app and loaded the library

## Future proofing

As a follow-up, I would replace `process.env.*` with global variables and enable this ESLint rule [no-process-env](https://eslint.org/docs/rules/no-process-env) for `.ts` files. Without this enforcement, it's too easy to introduce a new process.env and forget to update Webpack and Rollup's configs.

For example, `global.d.ts`:
```ts
declare const ASSET_PATH: string;
declare const DEBUG: boolean;
declare const LEGACY_INTEGRATIONS_PATH: string;
declare const VERSION: string;
```


pooyaj:
@gmathieu highly appreciate the PR 🙏 and sorry for the delay in the review. We are discussing internally whether we should move from `tsc` to `rollup`, but will prioritize the review of this.

BorePlusPlus:
Just wanted to confirm that we're hitting the same issue when trying to use the module. It would be great if it could be addressed. 🤞

danieljackins:
Addressed with https://github.com/segmentio/analytics-next/pull/331

PR #331, opened 2021-11-23
Clean up environment variables

danieljackins:
- Removes dependency on environment variables other than webpack specific behaviour.
- Tracks if users are using npm version or web version for better analytics

## Testing

Version being populated in call:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   ],
   "sideEffects": false,
   "scripts": {
-    "build-prep": "sed -i '' \"s/export const version: .*$/export const version: string | undefined = '$npm_package_version'/\" build/build.ts",
+    "build-prep": "sed -i '' \"s/export const version: .*$/export const version = '$npm_package_version'/\" build/build.ts",
     "umd": "webpack",
     "pkg": "tsc",
     "cjs": "tsc -p tsconfig.cjs.json",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   ],
   "sideEffects": false,
   "scripts": {
-    "build-prep": "sed -i '' \"s/export const version: .*$/export const version = '$npm_package_version'/\" build/build.ts",
     "umd": "webpack",
     "pkg": "tsc",
     "cjs": "tsc -p tsconfig.cjs.json",
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   ],
   "sideEffects": false,
   "scripts": {
+    "build-prep": "sed -i '' \"s/export const version: .*$/export const version: string | undefined = '$npm_package_version'/\" build/build.ts",
     "umd": "webpack",
     "pkg": "tsc",
     "cjs": "tsc -p tsconfig.cjs.json",
--- a/src/core/stats/remote-metrics.ts
+++ b/src/core/stats/remote-metrics.ts
@@ -77,10 +77,9 @@ export class RemoteMetrics {
 
     const type = getVersion()
     if (type === 'web') {
-      formatted['library_version'] = `next-${version}` ?? 'next-undefined'
+      formatted['library_version'] = `next-${version}`
     } else {
-      formatted['library_version'] =
-        `npm:next-${version}` ?? 'npm:next-undefined'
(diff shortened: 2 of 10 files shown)


src/core/stats/remote-metrics.ts
```diff
@@ -72,8 +73,14 @@ export class RemoteMetrics {
     }, {} as Record<string, string>)
 
     formatted['library'] = 'analytics.js'
-    formatted['library_version'] =
-      `next-${process.env.VERSION}` ?? 'next-undefined'
+
+    if (typeof process.env.VERSION !== 'undefined') {
```
pooyaj, line 77: Would reading `process.env.VERSION` create a similar issue to https://github.com/segmentio/analytics-next/issues/303 for non-webpack users? I think if `process.env` is undefined this may cause exception 🤔

pooyaj: Also in order to follow DRY ( we have this check in two places), might want to consolidate this logic somewhere, like a `getVersion` method that does this check and returns `npm` vs. `web` versions.

src/core/stats/remote-metrics.ts
```diff
@@ -72,8 +73,14 @@ export class RemoteMetrics {
...
+      formatted['library_version'] =
+        `web:next-${version}` ?? 'web:next-undefined'
```
pooyaj, line 79: let's remove the prefix for `web:` ... although cleaner, concerned about CDN users already relying on the format for something 😭

package.json
```diff
@@ -12,6 +12,7 @@
   ],
   "sideEffects": false,
   "scripts": {
+    "build-prep": "sed -i '' \"s/export const version: .*$/export const version = '$npm_package_version'/\" build/build.ts",
```
pooyaj, line 15: My bad for not suggesting this alternative earlier 😅 😅 we don't need the `sed` magic to build the `build/build.ts` file. We can just do: 
```
import { version } from '../../../package.json'
```
whenever we have 
```
import { version } from '../../../build/build.ts'
```

pooyaj (commented):
Perfect 💯  Added suggested improvements

src/core/stats/remote-metrics.ts
danieljackins: typeof will return 'undefined' if the variable doesn't exist, not an exception

package.json
danieljackins: `import { version} from ./package.json` will import the entire package.json into the webpack build, though

pooyaj: Oh! I see, then probably we have to go with `sed` magic 👍

src/core/stats/remote-metrics.ts
pooyaj: I think for those users, the `process` is undefined, so `typeof process.env.VERSION` will throw a 
`Uncaught ReferenceError: process is not defined`

src/core/stats/remote-metrics.ts
```diff
@@ -72,8 +74,14 @@ export class RemoteMetrics {
     }, {} as Record<string, string>)
 
     formatted['library'] = 'analytics.js'
-    formatted['library_version'] =
-      `next-${process.env.VERSION}` ?? 'next-undefined'
+
+    const type = getVersion()
+    if (type === 'web') {
+      formatted['library_version'] = `next-${version}` ?? 'next-undefined'
```
pooyaj, line 80: nit, but I think `next-${version}` will always be defined, so the `?? `next-undefined` never happens, and we can safely remove it

src/core/stats/remote-metrics.ts
```diff
@@ -72,8 +74,14 @@ export class RemoteMetrics {
...
+    } else {
+      formatted['library_version'] =
+        `npm:next-${version}` ?? 'npm:next-undefined'
```
pooyaj, line 83: same as above

pooyaj (approved):

BorePlusPlus:
Nice one 🎉 

How long does it usually take for a fix like this to make it into a release?

eliashassing154:
Now that it's merged, when will this be live? Really waiting for this to be fixed. Thank you

PR #333, opened 2021-11-25
Catch localstorage quota errors

danieljackins:
This PR makes sure we're catching any localStorage errors, which some users have been encountering when localStorage is full.

## Testing

- Unit test added
- Tested locally
image


pooyaj (approved):
🙌

PR #335, opened 2021-11-29
updates module and types fields in package.json to reference correct build path

chrisradek:
This change fixes the `module` and `types` fields in `package.json` to reference the correct paths: `dist/pkg/src`. The change is needed because we recently started importing `package.json` directly in our code, so `package.json` now shows up in `dist/pkg` with the JS output under the `src` folder.

I ran into this (and another problem related to the `nextjs` update to 11) when running `make dev` since the imports used in the examples directory pointed to the incorrect files.

Alternatively, we could stop importing `package.json` and reference a generated as mentioned [here](https://github.com/segmentio/analytics-next/pull/331#discussion_r756535573).

@danieljackins and @pooyaj since you have more insight into why we import `package.json` directly instead of generating the version file.

As a bonus, this also fixes `make dev`!

### DIFF
--- a/example/next-env.d.ts
+++ b/example/next-env.d.ts
@@ -1,2 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
+/// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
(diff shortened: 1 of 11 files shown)


pooyaj (approved):

chrisradek:
Closing. Will open a new PR that doesn't import package.json instead.

PR #336, opened 2021-11-29
replaces package.json imports with generated version constant

chrisradek:
- Replaces the package.json imports with a `version` constant that is generated at package build time. This code is heavily borrowed from @danieljackins work in https://github.com/segmentio/analytics-next/pull/331. Webpack was giving warnings that eventually we will have to use the default export when importing `package.json`, which would mean importing the entire file rather than just the version field. This strategy should be safe regardless of which bundler is used.
- Updates the version of `react`/`react-dom`/`@types/react` in the `example` project to meet the peer dependency requirements for next v11. This in addition to the above change allows `make dev` to work.

## Testing

- Added a unit test to ensure that the version constant is always in sync with what's in package.json.
- Tested with bundler (npm) example:
npm-version
- Tested with standalone (web) example:
web-version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ node_modules: package.json yarn.lock
 	@touch $@
 
 build: ## Builds typescript files and UMD library
-	yarn clean && yarn concurrently "yarn umd" "yarn pkg"
+	yarn clean && yarn build-prep && yarn concurrently "yarn umd" "yarn pkg"
 .PHONY: build
 
 build-browser:
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ build-browser:
 .PHONY: build-browser
 
 build-prod: ## Builds libraries in prod mode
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@ node_modules: package.json yarn.lock
 	@touch $@
 
 build: ## Builds typescript files and UMD library
-	yarn clean && yarn concurrently "yarn umd" "yarn pkg"
+	yarn clean && yarn build-prep && yarn concurrently "yarn umd" "yarn pkg"
```
pooyaj, line 16: We need to add a similar prep for the `build-prod` step

pooyaj (approved):
Looks great 👌

Makefile
chrisradek: Thanks for the tip, added!

eliashassing154:
Now that it's merged, when will this be live? Really waiting for this to be fixed. Thank you

PR #340, opened 2021-12-02
safely get global object

danieljackins:
This PR is to safely fetch the global object in environments where it isn't natively available.

## Testing

Please, make sure to describe how you tested this change, include any gifs or screenshots you find necessary.

### DIFF
--- a/src/lib/global-shim.ts
+++ b/src/lib/global-shim.ts
@@ -1,0 +1,14 @@
+export function getGlobal() {
+  // see https://mathiasbynens.be/notes/globalthis for more details
+  if (typeof globalThis === 'object') return
+  Object.defineProperty(Object.prototype, '__magic__', {
+    get: function () {
+      return this
+    },
+    configurable: true, // This makes it possible to `delete` the getter later.
+  })
+  // @ts-expect-error defined above
+  __magic__.globalThis = __magic__
+  // @ts-expect-error defined above
+  delete Object.prototype.__magic__
+}
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "24.1 KB"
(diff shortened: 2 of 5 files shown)


pooyaj:
Oops, bundle size increase by more than 1kb 😢 can we check if we can possibly only either use this for the `npm` package or maybe find a lighter weight implementation? ( The CDN version doesn't need the pollyfill because webpack already pollyfills the `global`, it is only the `npm` package that needs it )

src/lib/global-shim.ts
```diff
@@ -0,0 +1,14 @@
+export function getGlobal() {
```
yordis, line 1: Not following why this would be a function that you suppose to call before making calls to the global context, is error-prune from my experience, it is easy to forget and not sure how many times, and where the thing is supposed to be.

Instead of,

A) Be a ponyfill (less ideal since you will force people to bundle this in)
B) Be a opt-in polyfill (ideally, as long as the CDN versions must make sure that include it if the CDN wants to support it @pooyaj mentioned that this is cover)

Would you mind sharing about it?

> it is only the npm package that needs it

@pooyaj I don't think so thou, that will be making an assumption about my environment, which makes sense for the CDN, but I may want to bundle for modern platforms that support such feature (https://caniuse.com/?search=globalThis)

Documentation for those importing from NPM should be sufficient, here is an example: https://github.com/straw-hat-team/fetcher/blob/master/docs/polyfills.md

pooyaj: > it is only the npm package that needs it

@yordis we use webpack to bundle the CDN version, and webpack automatically takes care of polyfilling `globalThis`. So we are covered by webpack on CDN. 

I agree, you raise a good point here. If we include this in the npm package, we are forcing the npm users to use our way of polyfilling globalThis, and potentially impact non-segment usage of global as a side-effect.

PR #343, opened 2021-12-02
adds checks to ensure process is defined before accessing it

### LINKED ISSUE #303: process is not defined

chrisradek:
Fixes #303 

I added a new function - `getProcessEnv()` that returns either `process.env` if process and process.env are defined, or an empty object. Also updated all references to `process.env` to use the result from `getProcessEnv()`.

## Testing

I ran the react sample in our README with a webpack config that does not shim environment variables into the bundle. Before these changes, I saw `process is undefined` in the console. After this change, the console is clear of errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "24.1 KB"
+      "limit": "24.8 KB"
     }
   ],
   "prettier": {
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "24.8 KB"
+      "limit": "24.7 KB"
     }
   ],
   "prettier": {
--- a/src/browser-umd.ts
+++ b/src/browser-umd.ts
@@ -1,0 +1,18 @@
+import { getCDN } from './lib/parse-cdn'
+
+if (process.env.ASSET_PATH) {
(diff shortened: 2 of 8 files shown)


danieljackins (approved):

chrisradek:
@danieljackins @pooyaj 
I updated this PR so that the `process.env.ASSET_PATH` codepath is only present in `browser-umd.ts` and `standalone.ts`. Also updated the UMD webpack config to use `browser-umd.ts` as the entry point, so the cjs/pkg builds don't load this codepath at all.

My understanding is that `standalone.ts` isn't actually used by the cjs/pkg builds, so it is safe to keep the ASSET_PATH logic in that file but let me know if I'm incorrect there.

CI passes, and so does running the react readme sample with a webpack config that doesn't define environment variables. Any suggestions on other testing to do?

nettofarah (approved):
Looks awesome!

PR #344, opened 2021-12-04
remove process.env requirement from getting version type

chrisradek:
#343 made checking `process.env` safer, but it turned out this broke the UMD webpack config `EnvironmentPlugin`: it could no longer replace `process.env.VERSION` with a hard-coded value.

This change removes the need for `process.env.VERSION`. Instead, the `getVersion` (renamed to `getVersionType`) has been refactored to always return the type `npm` unless `setVersionType` is called. `setVersionType('web')` is now called from the `standalone.ts` and `browser-umd.ts` files which are only used for the hosted libraries.

This change also slightly reduces the size of the library!

## Testing

When running `make dev` and running the example, I see the `npm:` prefix on the version as expected since this example uses a bundler and not the UMD build.

When running `make standalone-example` and running the local-example, I don't see the `npm:` prefix on the version anymore (I did prior to this change).

I'd love to write a test for this but I didn't find any examples using our web build unless they were also running against classic. Is there a place to run UMD tests automatically already or is that something we can consider in the future?

### DIFF
--- a/src/browser-umd.ts
+++ b/src/browser-umd.ts
@@ -1,4 +1,5 @@
 import { getCDN } from './lib/parse-cdn'
+import { setVersionType } from './plugins/segmentio/normalize'
 
 if (process.env.ASSET_PATH) {
   if (process.env.ASSET_PATH === '/dist/umd/') {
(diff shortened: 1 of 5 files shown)


pooyaj (approved):
Looks great

PR #352, opened 2021-12-30
export AnalyticsNode

### LINKED ISSUE #351: Does not work with Next.js SSR or api routes
When trying to use analytics-next on backend window is not defined. The source code that breaks the initialization lies here:
https://github.com/segmentio/analytics-next/blob/master/src/lib/priority-queue/persisted.ts#L112

Am I doing anything wrong?

```
ReferenceError: window is not defined
    at new PersistedPriorityQueue (/Users/franzvonderlippe/Programming/ally/node_modules/@segment/analytics-next/dist/cjs/lib/priority-queue/persisted.js:88:9)
    at new Analytics (/Users/franzvonderlippe/Programming/ally/node_modules/@segment/analytics-next/dist/cjs/analytics.js:36:97)
    at getServerSideProps (webpack-internal:///./pages/bookings/[id]/index.tsx:237:27)
    at Object.renderToHTML (/Users/franzvonderlippe/Programming/ally/node_modules/next/dist/server/render.js:479:26)
    at async doRender (/Users/franzvonderlippe/Programming/ally/node_modules/next/dist/server/next-server.js:1389:38)
    at async /Users/franzvonderlippe/Programming/ally/node_modules/next/dist/server/next-server.js:1484:28
    at async /Users/franzvonderlippe/Programming/ally/node_modules/next/dist/server/response-cache.js:63:36
    ```

### PR DESCRIPTION
**Pull request by chrisradek**
Fixes #351 

`AnalyticsNode` isn't currently being exported from our package. This means the package won't work in node.js environment since it depends on the `PersistentPriorityQueue` which in turn relies on `window` being present.

One possible concern I had was that exporting `AnalyticsNode` from the `index.ts` file could add bloat when bundled. `make size` shows no change in bundle size. I also tested bundling with `webpack` and `vite` (the latter still needs the `global` fix) and tree-shaking appeared to be working properly.

## Testing

I created a simple node.js script using and confirmed I was able to see my `track` call in the segment debugger.

Sample script:
```js
const { AnalyticsNode } = require('@segment/analytics-next');

async function run() {
    const [analytics, ctx] = await AnalyticsNode.load({writeKey: 'key'});
    await analytics.track("test");
}

run().then(() => console.log('ran')).catch(console.error);
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "sideEffects": false,
   "scripts": {
     "build-prep": "echo \"// This file is generated.\\nexport const version = '$npm_package_version'\" > src/generated/version.ts",
+    "version": "npm run build-prep",
     "umd": "webpack",
     "pkg": "tsc",
     "cjs": "tsc -p tsconfig.cjs.json",
@@ -52,6 +53,7 @@
     "@segment/tsub": "^0.1.9",
     "dset": "^3.0.0",
     "js-cookie": "^2.2.1",
+    "node-fetch": "^2.6.1",
     "spark-md5": "^3.0.1",
     "tslib": "^2.1.0",
     "unfetch": "^4.1.0"
@@ -107,7 +109,6 @@
     "micro-memoize": "^4.0.9",
     "mime": "^2.4.6",
     "ngrok": "^3.3.0",
-    "node-fetch": "^2.6.1",
     "np": "^7.5.0",
     "p-map": "^4.0.0",
     "playwright": "^1.10.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
   "sideEffects": false,
   "scripts": {
     "build-prep": "echo \"// This file is generated.\\nexport const version = '$npm_package_version'\" > src/generated/version.ts",
+    "version": "npm run build-prep",
```
chrisradek, line 16: This change should cause `lerna` to update the version that's generated after bumping the package.

nettofarah (approved):
Awesome!!

pooyaj (approved):

PR #356, opened 2022-01-06
Resolve indirect sourcemaps for easier debugging

zikaari:
This patch will follow and include indirect sourcemaps using `sourceMappingURL` (if available at the end of a `.js` file).

**Before**

before

___

**After**

after


pooyaj (approved):
Awesome 🙌  please feel free to merge anytime

PR #357, opened 2022-01-10
[LIBWEB-979] Fix destination negation logic

zikaari:


pooyaj (approved):

src/core/queue/event-queue.ts
```diff
@@ -211,13 +211,10 @@ export class EventQueue extends Emitter {
   }
 
   private availableExtensions(denyList: Integrations): PluginsByType {
-    const available =
-      denyList.All === false
-        ? this.plugins.filter(
-            (p) => denyList[p.name] !== undefined || p.type !== 'destination'
-          )
-        : // !== false includes plugins not present on the denyList
-          this.plugins.filter((p) => denyList[p.name] !== false)
```
zikaari, line 220: @pooyaj I just noticed that this clause didn't consider the plugin `type`, whether it being a `destination` or lack thereof, not sure if that was intentional.

pooyaj: @zikaari Good callout. I think there is a small edge case that this change may cause a regression. The items in the `denyList` will be almost always `destination` plugins, except, `Segment.io` which is an `after` plugin. So if someone passes the following denyList, then the old code will not send data to `Segment.io` but the new code will send it: 
```
{
'Segment.io': false
}
```

Also as a side-node, I'm not sure which one will be a desirable behavior, but we might want to avoid the regression here.

zikaari: > I'm not sure which one will be a desirable behavior, but we might want to avoid the regression here.

The old behaviour is conflicting in itself. The first clause in the ternary, when `All: false`, specifically gives non-destination plugins a greenlight, however the latter had no check of any kind. It was inconsistent.

It's a tough spot, either continue with the inconsistency OR have an edge-case regression for some. The challenge is that it'll be difficult to properly document the former for public docs.

pooyaj: One rule of thumb in these cases would be to check the behavior of classic AJS in this instance: https://github.com/segmentio/analytics.js-core/blob/master/lib/analytics.ts ... As we claim behavior backwards compatibility the two behavior should match. I can't remember what was the classic behavior in this case tho 🤔

pooyaj: @zikaari just checked the old library behavior, and it is consistent with the new code. So we are all good here.

PR #358, opened 2022-01-10
fixes make release version generation

chrisradek:
The `make release` command was updating `src/generated/version.ts` but was not comitting it before pushing changes to the repo. This change ensures that the generated file is included in the staged changes prior to `np` comitting changes.

## Testing

Tested these changes with my fork and now see `src/generated/version.ts` being updated:
https://github.com/chrisradek/analytics-next/commit/83957a77407fff145a85dcaec2e2211c5f9e025b


pooyaj (approved):
💥

PR #360, opened 2022-01-11
Fix analytics on initialize

arielsilvestri:
This PR provides a fix to event handler registration during setup. By flushing event emitters from the buffer prior to a full buffer flush, we register the events early enough for user provided callbacks to fire as desired.

## Testing

Added a test for this change which mimics the test for `setAnoymousId`. Used `index-local`, registering `initialize` callbacks within the script tag in the example before and after load fires to ensure the callback would execute as intended.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -84,7 +84,7 @@ async function flushBuffered(analytics: Analytics): Promise<void> {
       // @ts-expect-error
       typeof analytics[operation] === 'function'
     ) {
-      if (operation === 'addSourceMiddleware') {
+      if (operation === 'addSourceMiddleware' || operation === 'on') {
         // @ts-expect-error
         await analytics[operation].call(analytics, ...args)
       } else {
@@ -220,9 +220,6 @@ export class AnalyticsBrowser {
       plugins
     )
 
-    analytics.initialized = true
-    analytics.emit('initialize', settings, options)
-
     if (options.initialPageview) {
       analytics.page().catch(console.error)
     }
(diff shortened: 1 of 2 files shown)


pooyaj (commented):
I wonder if can flush `on` method at the same time as `flushAnonymousUser` (i.e., earlier)  in order to avoid moving the code block with `analytics.initialized`/`analytics.emit`? That may also fixes the tests 🤔

arielsilvestri:
I'll give that a shot - seems like it would be a little less surface area being changed by doing so.

arielsilvestri:
@pooyaj that worked functionality wise but it sadly did not fix the QA tests. I'll look into it.

pooyaj:
@arielsilvestri actually looks like the unit tests are passing. `Snyk` is failing tho, but that's probably unrelated to the change.

arielsilvestri:
@pooyaj I'm still seeing the test-qa tests failing. Have those been failing all the time?

pooyaj:
Ah, those tests are flakey, so we remove them from the CI success/failure. They can behave differently between runs, so as far as most of them work we should be good 👍

src/browser.ts
```diff
@@ -116,6 +116,27 @@ function flushAnonymousUser(analytics: Analytics): void {
   }
 }
 
+/**
+ * Ensures events can be registered before library initialization.
+ * This is important so users can register to 'initialize' and any events that may fire early during setup.
+ */
+
+function flushEventEmitters(analytics: Analytics): void {
```
pooyaj, line 124: Can we merge the `flushAnonymousUser` and `flushEventEmitters` into one function to avoid repeating the shared logic?

pooyaj (approved):
💥

PR #363, opened 2022-01-13
Use decodeURIComponent for querystring api params

### LINKED ISSUE #342: Querystring `ajs_uid` is not decoded when passed to identify
Hi, we use email as User ID and want to use querystring api, but what happens if I send customers to `https://example.com/?ajs_uid=some@example.org` it often ends on `https://example.com/?ajs_uid=some%40example.org` because web-server or app redirect to encoded url, it results that identify is fired for `some%40example.org` however the url parameter is still `some@example.org` if accessed properly using `URLSearchParams`.

I expect that analytics.js should be able to decode url params properly.

zikaari:
Fixes #342

Note: I switched the utm param processing to also use the graceful version of `decodeURIComponent`, please think about any side-effects this may have that I don't know of. I think this is preferable since it'll prevent an error like that from breaking the end-user's page.

### DIFF
--- a/src/core/query-string/__tests__/gracefulDecodeURIComponent.test.ts
+++ b/src/core/query-string/__tests__/gracefulDecodeURIComponent.test.ts
@@ -1,0 +1,17 @@
+import { gracefulDecodeURIComponent } from '../gracefulDecodeURIComponent'
+
+describe('gracefulDecodeURIComponent', () => {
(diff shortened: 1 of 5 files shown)


pooyaj (approved):
looks good to me 🙌

pooyaj:
Can we add a test case to the query-string suite for what the customers reported around emails?

zikaari:
Done

PR #364, opened 2022-01-13
Pull request advisory visible only to the PR author

zikaari:
When creating a pull request through CLI tools the pull request template also makes into the final comment.

This patch makes the advisory visible only to the author while the comment is being composed, it won't be visible when saved.

Let me know if anyone also wants to touch up the phrasing


pooyaj (approved):
👌👌

PR #368, opened 2022-01-27
fix braze debounce middleware unwanted loading

zikaari:
This patch only fixes Braze classic. Action based Braze is also faulty
and will be fixed as a different patch as it is caused by a different problem
somewhere else.


pooyaj (approved):
Looks great 💥

Issue #370, opened 2022-02-01
Allow overriding messageId


jflayhart:
We have other tracking libraries we use on our site that we wish to use the same messageId for the same event sent to different sources, for example:

**User clicks button (Event A):**
```js
const messageId = generateMsgId() // 456
const event = 'Event A'

segment.track(event, {}, { messageId }) // sends Event A with messageId 456
anotherSDK.track(event, { messageId }) // sends Event A with messageId 456
```

Instead, this is the problem we're running into:

**User clicks button (Event A):**
```js
const messageId = generateMsgId() // 456
const event = 'Event A'
const properties = ...

segment.track(event, properties, { messageId }) // sends Event A with different messageId `ajs-next-*`
anotherSDK.track(event, properties, { messageId }) // sends Event A with messageId 456
```

Is there not a way to override the messageId per event, so we can use our own format in order to aggregate the same event across different sources? I see no way to do it with this library, but [in the docs it says we can](https://segment.com/docs/guides/duplicate-data/#99-deduplication) and you can in [your other libs](https://github.com/segmentio/analytics-python/issues/66):

> Keep in mind that Segment’s libraries all generate message_ids for each event payload, with the exception of the Segment HTTP API, which assigns each event a unique message_id when the message is ingested. **You can override these default generated IDs and manually assign a message_id if necessary.**

If there isn't a way in JS, I would propose Segment's JS lib allows it according to your docs.

jflayhart:
It seems [this is why](https://github.com/segmentio/analytics-next/blob/v1.33.4/src/core/events/index.ts#L234) `messageId` can't be overridden because of the spread operator placement.

Can we simply allow people to override it at their own risk? Again, docs state "You can override these default generated IDs and manually assign a message_id if necessary."

So the change could be:

```

```

jflayhart:
For anyone else running across this "issue" it wasn't very clear but I found some middleware docs for Analytics 2.0 that allows us to mutate the payload object. Yes, feels messy but it works.

https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/middleware/

pooyaj:
Closing since this is possible through a middleware!

troex:
agree with @jflayhart!

@pooyaj it is still a bummer to do with a middleware, other segment libraries allow to pass messageId and it is in a spec. In our case we want to send events with exact messageId to make sure they are not tracked twice, and we don't have a status tracking if event happened before and if it was delivered but we only want it once.

PR #371, opened 2022-02-01
[RFC] Allow messageId override

jflayhart:
Please view issue #370. Mainly just want to make this more easily overridden like userId and anonymousId can be per event.

tl;dr addSourceMiddleware doesn't cut it because I need to regen a messageId in my own track middleware, which is where all our events are fired and those events (outside segment's use case) need to use the same messageId.

- [ ] Test `tracking` events with messageId override
- [ ] Test `page` events with messageId override

### DIFF
--- a/src/core/events/index.ts
+++ b/src/core/events/index.ts
@@ -169,7 +169,7 @@ export class EventFactory {
    * are provided in the `Options` parameter for an Event
    */
   private context(event: SegmentEvent): [object, object] {
-    const optionsKeys = ['integrations', 'anonymousId', 'timestamp', 'userId']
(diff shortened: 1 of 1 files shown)


pooyaj:
Closing this one, as there is workaround via middleware

troex:
agree with @jflayhart!

@pooyaj it is still a bummer to do with a middleware, other segment libraries allow to pass `messageId` and it is in a spec. In our case we want to send events with exact `messageId` to make sure they are not tracked twice, and we don't have a status tracking if event happened before and if it was delivered but we only want it once.

PR #373, opened 2022-02-02
[bugfix] honor blocking unplanned events in tracking plan

chrisradek:
This PR updates the schema-filter and ajs-destination plugins to properly support blocking uplanned track calls.

Our documentation currently states that a source's schema configuration can be setup to disallow unplanned track events: [documentation](https://segment.com/docs/protocols/enforce/schema-configuration/#:~:text=IMPORTANT%3A%20Unplanned%20event%20blocking%20is%20supported%20across%20all%20device%2Dmode%20and%20cloud%2Dmode%20Destinations.) 
The analytics library correctly drops events to destinations if a specific destination is disabled for an event, but the library wasn't reading the `__default` tracking plan settings that is updated when unplanned events are blocked at the source level.

I verified that the `__default` field is supposed to be present in the `settings.plan.track` object (the `__` prefix is to prevent collisions with user-provided event names, not to indicate it is private).

I tested a source using the Braze device mode (actions) destination and the classic Amplitude device mode destination, and both allow expected events and drop unplanned events after this change.

### DIFF
--- a/src/core/events/interfaces.ts
+++ b/src/core/events/interfaces.ts
@@ -155,9 +155,13 @@ export interface SegmentEvent {
  * A Plan allows users to specify events and which destinations they would like them to be sent to
  */
 export interface Plan {
-  track?: {
-    [key: string]: PlanEvent
-  }
+  track?: TrackPlan
+}
+
+export interface TrackPlan {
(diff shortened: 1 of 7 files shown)


pooyaj (approved):
Looks great 🙌

PR #375, opened 2022-02-08
Recursively scan and convert any classes in event.properties

arielsilvestri:
An edge case for when we're processing instances of classes with our cloning utilities occurs where the clone won't be properly made. The code introduced here is the code from the `klona/lite` package that `Facade` depends on, but changing this line to spread the properties of the class onto a new object instance, instead of a new class instance: https://github.com/lukeed/klona/blob/master/src/lite.js#L8

I'm wondering if there are other ways to address this issue and I'm open to feedback. We'll effectively be cloning objects twice now, and I'm wondering if we're better off doing some in-place replacement of any class instances with objects instead of full copies.

### DIFF
--- a/src/lib/to-facade.ts
+++ b/src/lib/to-facade.ts
@@ -14,7 +14,55 @@ export type SegmentFacade = Facade<SegmentEvent> & {
   obj: SegmentEvent
 }
 
+interface Properties {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any
+}
+
+interface ClonedObj {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any
+}
+
+const convertClassesToObjects = (
+  properties: Properties,
+  tmp: ClonedObj = {}
+) => {
+  if (typeof properties !== 'object') return properties
+  if (Object.prototype.toString.call(properties) === '[object Object]') {
+    if (
+      properties.constructor !== Object &&
+      typeof properties.constructor === 'function'
+    ) {
+      tmp = { ...properties }
+      for (const key in properties) {
+        // eslint-disable-next-line no-prototype-builtins
+        if (properties.hasOwnProperty(key) && tmp[key] !== properties[key]) {
+          tmp[key] = convertClassesToObjects(properties[key])
+        }
+      }
+    } else {
+      for (const key in properties) {
+        if (key === '__proto__') {
+          Object.defineProperty(tmp, key, {
+            value: convertClassesToObjects(properties[key]),
+            configurable: true,
+            enumerable: true,
+            writable: true,
+          })
+        } else {
+          tmp[key] = convertClassesToObjects(properties[key])
+        }
+      }
+    }
+  }
+  return tmp
+}
+
 export function toFacade(evt: SegmentEvent, options?: Options): SegmentFacade {
+  if (evt.properties) {
+    evt.properties = convertClassesToObjects(evt.properties)
+  }
   let fcd = new Facade(evt, options)
 
   if (evt.type === 'track') {
--- a/src/plugins/segmentio/__tests__/index.test.ts
+++ b/src/plugins/segmentio/__tests__/index.test.ts
@@ -13,7 +13,10 @@ jest.mock('unfetch', () => {
 
 class StubbedClass {
   name: string
-  constructor(name: string) {
+  constructor(name?: string) {
+    if (!name) {
+      throw new TypeError('Missing argument: name')
+    }
     this.name = name
   }
 }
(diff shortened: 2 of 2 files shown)


src/plugins/segmentio/__tests__/index.test.ts
```diff
@@ -105,17 +112,22 @@ describe('Segment.io', () => {
 
   describe('#track', () => {
     it('should enqueue an event and properties', async () => {
```
chrisradek, line 114: This test appears to be passing even without your change. I think that's because your stubbed class doesn't throw any errors in the constructor - so even the current implementation will be able to instantiate the class and then copy the properties over.

In the example I saw, the instantiation of the new class would fail due to an error (I believe a `TypeError` due to trying to access a field on an `undefined` object.) For testing, you can mimic that by throwing an error in the constructor if the parameters aren't passed.

arielsilvestri: We had suggested to the client to clone their class instance into a new object, because it appeared that using a class instance did not play well with the data flow.  It sounds like the problem may be elsewhere if a unit test trying to mimic this experience ends up green. I'll take a closer look at things on my end.

arielsilvestri: This was a good catch, @chrisradek. I do think there's also an element of something you mentioned here going wrong, in tandem with the way that the data being passed as a class is incongruent with the cloning utility.

chrisradek: In one user's case, their class was something like:
```js
class Foo {
  constructor(props) {
    Bar.call(this, props.a, props.b, props.c);
  }
}
```
Since the cloning utility doesn't have access to `props` to pass in when instantiating the class, the constructor threw a TypeError since it was trying to access fields on `props` while `props` was undefined. I think copying the fields to a new plain object would have worked fine (like your change does).

Anyway, that's why I think we could reproduce the issue in a test easily enough by just throwing a TypeError if nothing is passed to the constructor.

arielsilvestri: Yep, that's what I noticed as well. I'll get a test up and running to reflect the issue and the fix better.

src/lib/to-facade.ts
```diff
@@ -14,7 +14,55 @@ export type SegmentFacade = Facade<SegmentEvent> & {
...
+interface ClonedObj {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any
+}
+
+const convertClassesToObjects = (
+  properties: Properties,
+  tmp: ClonedObj = {}
+) => {
+  if (typeof properties !== 'object') return properties
```
chrisradek, line 31: It looks like this function doesn't properly handle arrays. 
I ran a test where one of the properties was an array:
```js
await analytics.track(
        'event',
        { prop: true, prop2: stubbedClass, list: ['a', 2, {foo: 'bar'}] },
        { opt: true }
      )
```
Before this change, the array was copied over correctly, but after this change I see `list: {}` instead. I think an `Array.isArray` check where you call `convertClassesToObjects` on each element _should_ work.

arielsilvestri:
@pooyaj I rather do that as well tbh. I think everything would be fixed if that one line where `klona` calls the constructor would pass proper arguments. I'm just not sure what the timeline would look like re: merge and deploy

src/lib/to-facade.ts
arielsilvestri: noticing the same with the qa tests. I'll look into it

arielsilvestri:
Here's what I'm thinking - let's port over the `klona/lite` code to `facade`, with the fix applied where we change [this line](https://github.com/lukeed/klona/blob/master/src/lite.js#L8) to:

`tmp = new x.constructor({...x})`

That should fix the issue as well as keep all the rest of the functionality in place. I think per @chrisradek's points, we would need the whole library locally in order to account for every case, and per your point @pooyaj , this will probably add too much to the bundle size.

Simultaneously, I can open up a PR against `klona` to fix the issue on their end and we won't have to depend on a third party library's schedule to get this fix out to our client.

chrisradek:
I'm for removing `klona` alotgether. It's setting `__proto__` which we really shouldn't rely on (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto).

I think instead of instantiating a new class based on the constructor, we can create a new object using Object.create() and then copy the props over.

Something like:
```js
function crazyCopyIdea(properties) {
  if (Object.prototype.toString.call(properties) === '[object Object]') {
    let temp = {};
    if (
      properties.constructor !== Object &&
      typeof properties.constructor === 'function'
    ) {
      temp = Object.create(properties);
    }
    for (const key in properties) {
      temp[key] = crazyCopyIdea(properties[key]);
    }
    return temp;
  } else if (Array.isArray(properties)) {
    return properties.map(crazyCopyIdea);
  } else {
    return properties;
  }
}
```

arielsilvestri:
`Object.create(properties)` and `{...properties}` are functionally similar iirc, I'm not entirely sure if there's perf differences anymore (I believe Object.create used to be _slightly_ better), but tbh there's nothing very wild about your suggestion, @chrisradek. It's just an easier-to-read, simplified version of the big chunk of logic in `klona`, with our fix applied.

arielsilvestri:
Closing this PR in favor of https://github.com/segmentio/facade/pull/168, can move discussion there.

PR #382, opened 2022-02-24
deps: update js-cookie to v3

chambo-e:
Hello 👋 

A small PR to bump js-cookie to v3


src/core/user/index.ts
```diff
@@ -87,32 +87,46 @@ export class Cookie extends Store {
...
+
+      if (!value) {
+        return null
+      }
+
+      try {
+        return JSON.parse(value)
+      } catch (e) {
+        return (value as unknown) as T
+      }
```
arielsilvestri, line 111: Just curious - how do we handle JSON we can't parse later on, or is it not a big deal?

arielsilvestri (approved):
lgtm

src/core/user/index.ts
chambo-e: I assumed that JSON that can't be parsed is "not" JSON 😄 

I kinda stole the logic from the previous implementation in js-cookie ([here](https://github.com/js-cookie/js-cookie/blob/54962f884e9ae33f93e13ac903ffaf1d5a523598/src/js.cookie.js#L125))

pooyaj:
merging to a local branch to allow the CI run the test suites

PR #385, opened 2022-03-02
Dont load disabled action destinations

danieljackins:
This PR makes action destinations respect the same logic as legacy destinations in the integrations object

Testing completed locally with a source with multiple action destinations enabled. Setting All: false and Fullstory (Actions): true properly only loads the fullstory actions:
image
image


pooyaj (approved):
💥 🙌

PR #386, opened 2022-03-03
Load obfuscated destinations

arielsilvestri:
In tandem with https://github.com/segmentio/analytics.js-integrations/pull/665, this will enable bundle obfuscation to be served to our customers who opt in. There's no strict timing on when one or the other needs to be merged as we have not yet provided documentation on how to use this feature.

The idea here is that consumers can pass `obfuscate: true` to the second argument of `analytics.load`, and we will fetch the obfuscated bundle files for the users.

Originally I thought I would need to translate the obfuscated file names using `atob`, but it turns out we pull in from workspace settings where those destinations are already in plain text in order to kick off other loading processes and setup, so that work was not needed.

Tested by pointing the paths for integrations and middleware to the staging bucket where I uploaded the obfuscated paths, where I was able to successfully load the obfuscated bundles.

### DIFF
--- a/src/plugins/ajs-destination/loader.ts
+++ b/src/plugins/ajs-destination/loader.ts
@@ -13,7 +13,7 @@ function normalizeName(name: string): string {
   return name.toLowerCase().replace('.', '').replace(/\s+/g, '-')
 }
 
-function obfuscatePathName(pathName: string, obfuscate = false): string {
+function obfuscatePathName(pathName: string, obfuscate = false): string | void {
   return obfuscate ? btoa(pathName).replace(/=/g, '') : undefined
 }
 
--- a/src/plugins/ajs-destination/loader.ts
+++ b/src/plugins/ajs-destination/loader.ts
@@ -33,14 +33,17 @@ export async function loadIntegration(
   analyticsInstance: Analytics,
   name: string,
   version: string,
-  settings?: { [key: string]: any }
+  settings?: { [key: string]: any },
+  obfuscate?: boolean
 ): Promise<LegacyIntegration> {
   const pathName = normalizeName(name)
   let obfuscatedPathName = ''
-  if (settings?.obfuscate) {
+  if (obfuscate) {
     obfuscatedPathName = btoa(pathName).replace(/=/g, '')
   }
-  const fullPath = `https://cdn.segment.build/next-integrations/integrations-2.0/integrations/b3B0aW1pemVseQ/${version}/b3B0aW1pemVseQ.dynamic.js.gz`
+  const fullPath = `${path}/integrations/${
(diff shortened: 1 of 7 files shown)


src/plugins/ajs-destination/loader.ts
```diff
@@ -33,10 +33,17 @@ export async function loadIntegration(
   analyticsInstance: Analytics,
   name: string,
   version: string,
-  settings?: object
+  settings?: { [key: string]: any },
+  obfuscate?: boolean
 ): Promise<LegacyIntegration> {
   const pathName = normalizeName(name)
-  const fullPath = `${path}/integrations/${pathName}/${version}/${pathName}.dynamic.js.gz`
+  let obfuscatedPathName = ''
```
pooyaj, line 40: Looks like we are repeating the same logic around `load`/`unload`. Should we consolidate in a function?

pooyaj: Also nit, but might be a bit cleaner: 
```
const obfuscatedPathName = obfuscate ? btoa(pathName).replace(/=/g, '') : ''; 
```

pooyaj (approved):
Looks great. Few nits otherwise lgtm

arielsilvestri:
@pooyaj In addition to some of the feedback here, I'm going to roll an e2e test into these changes. As far as a unit test goes, I think a proper e2e test would be a better choice for these changes and should cover all our needs.

src/plugins/ajs-destination/loader.ts
```diff
@@ -76,12 +85,17 @@ export async function loadIntegration(
...
 ): Promise<void> {
   const pathName = normalizeName(name)
-  return unloadScript(
-    `${path}/integrations/${pathName}/${version}/${pathName}.dynamic.js.gz`
-  )
+  const obfuscatedPathName = obfuscatePathName(name, obfuscate)
+
+  const fullPath = `${path}/integrations/${
+    obfuscate ? obfuscatedPathName : pathName
+  }/${version}/${obfuscate ? obfuscatedPathName : pathName}.dynamic.js.gz`
```
chrisradek, line 96: Same comment here wrt replacing the ternary expressions with the nullish expressions.

chrisradek (commented):
Looking good! I'm wondering if we should also obfuscate the action-destination names as well. Since `obfuscate` is set for a.js and not individual destinations, it _might_ break some CSP rules if we wanted `obfuscate` to apply to other bundles beyond what we allow with these changes.

arielsilvestri:
@chrisradek Just to clarify here, the obfuscation is set to individual destination bundles, not ajs. Would the CSP issues be if we served an obfuscated destination from something that didn't have the client's name in the bundle, as the unobfuscated destinations currently have?

PR #387, opened 2022-03-03
Upload builds to shadow pipeline as well

danieljackins:
This PR uploads our builds to the shadow pipeline buckets as well so it can properly build AJS 2.0 projects. The env variables have been added to our buildkite.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -13,7 +13,10 @@ const bucket =
     ? process.env.PROD_BUCKET
     : process.env.STAGE_BUCKET
 
-const shadowBucket = process.env.SHADOW_BUCKET
+const shadowBucket =
+  process.env.NODE_ENV == 'production'
+    ? process.env.PROD_SHADOW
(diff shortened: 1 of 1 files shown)


scripts/release.js
```diff
@@ -68,6 +73,14 @@ async function upload(meta) {
         mime.getType(filePath.replace('.gz', '')) || 'application/javascript',
     }
 
+    const shadowOptions = {
+      Bucket: shadowBucket,
+      Key: path.join(`analytics-next`, meta.branch, meta.sha, f),
+      Body: await fs.readFile(filePath),
+      ContentType:
+        mime.getType(filePath.replace('.gz', '')) || 'application/javascript',
+    }
```
chrisradek, line 82: Looks like this can be simplified and remove an IO operation!
```suggestion
    const shadowOptions = {
      ...options,
      Bucket: shadowBucket,
    }
```

chrisradek (approved):

PR #388, opened 2022-03-03
adds disableClientPersistance field to allow turning off client-side storage

chrisradek:
This PR adds a new field to `InitOptions`: `disableClientPersistance`. When set to `true`, this will turn off storing data from a.js in cookies or localstorage. This is useful for users that do not want any client-side storage to be used by a.js.

### Current cookies usage
Cookies are used in 2 places:
- identity resolution in `User` and `Group`
- session cookie for `referrer` used by the `normalize` method from the segment.io plugin.

### LocalStorage usage
LocalStorage is used in 2 places:
- user/group traits in `User` and `Group`
- Storing events that aren't sent during `beforeunload`, handled by `PersistedPriorityQueue`

### Limitations

If a user sets `disableClientPersistance: true` but explicitly passes in a `PersistedPriorityQueue`, `User`, or `Group` that uses client-side storage, then the library will not be able to disable client-side storage for those objects. However, those can only be passed if using the `Analytics` class directly - most users should be using `AnalyticsBrowser`. Users would also need to explicitly define those objects so it's reasonable to defer to their judgement in this case.


pooyaj (commented):
Looks great! There are few other `PersistedPriorityQueue`s ( for legacy destinations and segmentio ) that need an update 😬

chrisradek:
> There are few other PersistedPriorityQueues ( for legacy destinations and segmentio ) that need an update

Woops, good point!

src/core/user/tld.ts
```diff
@@ -47,3 +47,11 @@ export function tld(url: URL): string | undefined {
     }
   }
 }
+
+export function tld(url: URL): string | undefined {
```
silesky, line 51: Can we rewrite this to be more explicit about why and where we're silently swallowing errors? For example, just wrapping the specific part of tld that we expect to be unsafe in a try/catch, leaving a comment, etc.

src/plugins/ajs-destination/index.ts
```diff
@@ -359,7 +361,7 @@ export async function ajsDestinations(
         name,
         version,
         integrationOptions[name],
-        options as object
+        options
```
silesky, line 364: nice

danieljackins (approved):
LGTM!

src/core/user/tld.ts
chrisradek: I added a comment on why we throw the error. Unfortunately `cookie.get()` itself will throw an error if the library is ran in a sandboxed environment. Unfortunately `Navigator.cookieEnabled` doesn't give the correct behavior in IE/Safari/old Edge inside iframes so we can't rely on that either.

PR #394, opened 2022-03-15
[bugfix] Honor Integration["Segment.io"]: false to API calls

chrisradek:
### What is this change:
This PR updates the plugin filtering logic to support denying the `Segment.io` plugin when `{Integrations: {"Segment.io": false}}` is passed to API calls (e.g. track/identify/etc).

### Why this change is needed:
Prior to this update, setting the `Segment.io` integration to false was only honored when passed to `analytics.load()`. All other destinations can already be disabled via the `integrations` object passed to API calls. This meant users explicitly trying to disable `Segment.io` for individual API calls were still seeing events sent to Segment.

### Difference compared to `analytics.load()` behavior
Passing `{ integrations: { All: false } }` to `analytics.load()` will cause events to __not__ be sent to Segment.

For individual event API calls (e.g. track/identify/etc), `Segment.io` has to be explicitly set to `false` in order to prevent events from being sent to Segment.

Behavior before this PR:
| A.JS Call | { Integrations: { All: false } } | { Integrations: { 'Segment.io': false } } |
| --- | --- | --- |
|analytics.load() | Not sent | Not sent |
| analytics.track() | Sent | Sent |

Behavior with this PR:
| A.JS Call | { Integrations: { All: false } } | { Integrations: { 'Segment.io': false } } |
| --- | --- | --- |
|analytics.load() | Not sent | Not sent |
| analytics.track() | Sent | Not sent |

### Risks

With this change, any code that currently passes `{ integrations: { 'Segment.io': false } }` to an event-generating API call will stop sending those events to Segment. This seems reasonable to treat as a bugfix since the user is explicitly disabling Segment in this case.

Passing only `{ integrations: { All: false } }` will __NOT__ prevent events from being sent to Segment. While this means the behavior of `All: false` differs between `analytics.load()` and event-generating methods, I don't think it's worth the risk of unintentionally disabling `Segment.io` if users are currently relying on the (incorrect) behavior where `All: false` only disables destination plugins.

### DIFF
--- a/src/__tests__/integration.test.ts
+++ b/src/__tests__/integration.test.ts
@@ -352,6 +352,37 @@ describe('Dispatch', () => {
     expect(segmentSpy).not.toHaveBeenCalled()
   })
 
+  it('does dispatch events to Segment.io when All is false', async () => {
+    const [ajs] = await AnalyticsBrowser.load({
+      writeKey,
+      plugins: [amplitude, googleAnalytics],
+    })
+
+    const segmentio = ajs.queue.plugins.find((p) => p.name === 'Segment.io')
+    expect(segmentio).toBeDefined()
+
+    const ampSpy = jest.spyOn(amplitude, 'track')
+    const gaSpy = jest.spyOn(googleAnalytics, 'track')
+    const segmentSpy = jest.spyOn(segmentio!, 'track')
+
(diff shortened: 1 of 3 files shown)


danieljackins (approved):
:shipit:

PR #397, opened 2022-03-16
Tests for obfuscated bundles

arielsilvestri:
Introduces unit and integration tests for the bundle obfuscation feature.

### DIFF
--- a/qa/__tests__/smoke.test.ts
+++ b/qa/__tests__/smoke.test.ts
@@ -137,12 +137,14 @@ describe('Smoke Tests', () => {
     expect(classicCookies['ajs_user_id']).toContain('Test')
 
     compareSchema(results)
+  })
 
-    const [url2, chrome2] = await Promise.all([server(true), browser()])
+  test.concurrent.each(samples)(`obfuscated smoke test`, async (writekey) => {
+    const [url, chrome] = await Promise.all([server(true), browser()])
     const obfuscatedresults = await run({
-      browser: chrome2,
+      browser: chrome,
       script: code,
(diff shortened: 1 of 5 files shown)


qa/__tests__/smoke.test.ts
```diff
@@ -137,5 +137,19 @@ describe('Smoke Tests', () => {
     expect(classicCookies['ajs_user_id']).toContain('Test')
 
     compareSchema(results)
+
+    const [url2, chrome2] = await Promise.all([server(true), browser()])
```
pooyaj, line 141: Should we extract this out into its own test? feels like a separate test with new setup etc.

qa/lib/runner.ts
```diff
@@ -34,18 +33,9 @@ export async function run(params: ComparisonParams) {
...
-          const obfuscatedUrl = request
-            .url()
-            .replace(new RegExp(key, 'g'), obfuscatedKey)
-          bundleRequests.push(request.url())
-          bundleRequests.push(obfuscatedUrl)
-        } else {
-          bundleRequests.push(request.url())
+        bundleRequests.push(request.url())
+        if (request.url().includes('next-integration')) {
+          bundleRequestFailures.push(request.failure())
```
pooyaj, line 38: This is cool 👌👌

pooyaj (approved):
Looks great! One minor comment otherwise lgtm

qa/__tests__/smoke.test.ts
arielsilvestri: It worked this time /shrug

PR #405, opened 2022-03-18
updates event-queue to only scheduleFlush when items exist in queue

### LINKED ISSUE #402: Using AnalyticsNode causes node process to hang
Hello, I'm currently working on a library that wraps this library to support both Node.js and React (as well as providing some custom validation logic).  As I've been working on the Node.js side, I noticed that running code similar to the sample script in https://github.com/segmentio/analytics-next/pull/352 causes the node process to hang.  I have two questions:
1. Is it intended for the node process to keep running after calling `AnalyticsNode.load`?  If so, is there a way to gracefully stop the process?
2. Is this the best library to use for integrating Segment into backend Node.js applications?  I saw that there's https://github.com/segmentio/analytics-node, but using the same library for both frontend and backend would simplify the maintenance of our wrapper library.

Any help would be appreciated, thank you!

chrisradek:
Fixes #402 

This PR changes the way the `EventQueue` class schedules flush calls. The main motivation for this change is to prevent `AnalyticsNode` from preventing the process from exiting due to a persistent `setTimeout` from `EventQueue's` polling.

## Current state

`EventQueue` starts polling it's queue via `scheduleFlush` as soon as it is instantiated. If there are no events in the queue, it will continue to poll every 500 ms. As mentioned above, this prevents a node.js process from exiting.

## What do these changes do?

These changes get rid of the persistent polling in `EventQueue`. Instead, `scheduleFlush` will only be called under the following conditions:
- `dispatch()` is called - same as currently
- After the `flush` invoked by `scheduleFlush` resolves, if the queue still has items - same as currently
- When a Context object transitions from a PriorityQueue's `future` array to `queue` array.
  - This is the part that replaces the need for persistent polling

### PriorityQueue.future

The PriorityQueue's `future` array holds events that failed to be sent initially, and are being retried with some backoff. 

Currently the `EventQueue` polls continuously to check if `priorityQueue.length > 0` to handle retried events getting re-enqueued. This PR changes how this is handled from being poll-based to push-based via listening for a new event emitted by the `PriorityQueue`. This should give a small performance benefit since now the flush will be scheduled immediately when an event is re-enqueued versus having to wait up to 500 ms for the poller to see it.

### DIFF
--- a/src/core/queue/__tests__/event-queue.test.ts
+++ b/src/core/queue/__tests__/event-queue.test.ts
@@ -11,16 +11,20 @@ import { Plugin } from '../../plugin'
 import { EventQueue } from '../event-queue'
 
 async function flushAll(eq: EventQueue): Promise<Context[]> {
-  let flushed: Context[] = []
-
+  const flushSpy = jest.spyOn(eq, 'flush')
   await pWhile(
     () => eq.queue.length > 0,
     async () => {
-      const res = await eq.flush().catch(() => [])
-      flushed = flushed.concat(res)
+      return new Promise((r) => setTimeout(r, 0))
     }
   )
-  return flushed
+  const results = flushSpy.mock.results.map((r) => r.value)
+  flushSpy.mockClear()
+
+  const flushed = await Promise.all(results)
+  return flushed.reduce((prev, cur) => {
+    return prev.concat(cur)
+  }, [])
 }
 
 const testPlugin: Plugin = {
@@ -97,6 +101,10 @@ test('does not enqueue multiple flushes at once', async () => {
   expect(setTimeout).toHaveBeenCalledTimes(1)
   expect(eq.queue.length).toBe(2)
 
+  // Ensure already enqueued tasks are executed
+  jest.runAllTimers()
+
+  // reset the world to use the real timers
   jest.useRealTimers()
   await flushAll(eq)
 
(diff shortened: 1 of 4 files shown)


src/core/queue/__tests__/event-queue.test.ts
```diff
@@ -11,16 +11,20 @@ import { Plugin } from '../../plugin'
...
 async function flushAll(eq: EventQueue): Promise<Context[]> {
-  let flushed: Context[] = []
-
+  const flushSpy = jest.spyOn(eq, 'flush')
   await pWhile(
     () => eq.queue.length > 0,
     async () => {
-      const res = await eq.flush().catch(() => [])
-      flushed = flushed.concat(res)
+      return new Promise((r) => setTimeout(r, 0))
```
chrisradek, line 18: I had to change this test code because the previous implementation was relying on `scheduleFlush` to be called initially with a timeout of 500 ms, and that the `flush` calls invoked would complete in less than 500 ms.

Since I removed the initial `scheduleFlush` from the EventQueue constructor, the 1st time it's invoked is with a timeout of 0 ms, so we weren't capturing all the meaningful `flush` calls.

pooyaj (approved):
Looks great 🙌

danieljackins (approved):
lgtm!

PR #406, opened 2022-03-22
Ensure anonymous id is set before analytics calls are made if passed as a query string

arielsilvestri:
Fixes an issue where analytics calls would use generated anonymous ids instead of the id set via query string when sending track and identify calls as part of the query string also.

Tested using the standalone example, as well as added tests.


pooyaj (approved):
:shipit:

PR #407, opened 2022-03-24
feat: add support for configuring CDN

### LINKED ISSUE #362: Allow cdn in AnalyticsSettings
Hello 👋 

If I understand correctly the current way to set the cdn is this 

```ts
window.analytics._cdn = <cdn>
```

Because this variable can then be reused in the plugins 

We try to avoid touching the window object as much as possible 
Would it be possible to pass the cdn in the AnalyticsSettings which is then passed down to the plugins ? 

```ts
AnalyticsBrowser.load({ writeKey, cdn })
```

```ts
export function loadLegacySettings(writeKey: string, cdn: string): Promise<LegacySettings> {
  return fetch(`${cdn}/v1/projects/${writeKey}/settings`)
    .then((res) => ({ ...res.json(), cdn }))
    .catch((err) => {
      console.warn('Failed to load settings', err)
      throw err
    })
}
```

Thanks :)

yordis:
fixes #362

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -39,6 +39,7 @@ const _analytics = globalAny.analytics
 
 export interface AnalyticsSettings {
   writeKey: string
+  cdn?: string
   timeout?: number
   plugins?: Plugin[]
   [key: string]: unknown
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -62,9 +62,9 @@ export interface AnalyticsBrowserSettings extends AnalyticsSettings {
   cdnSettings?: LegacySettings & Record<string, unknown>
 }
 
-export function loadLegacySettings(writeKey: string): Promise<LegacySettings> {
-  const cdn = window.analytics?._cdn ?? getCDN()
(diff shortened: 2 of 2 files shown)


pooyaj:
awesome! thanks @yordis 🎉 There are few places that we need to propagate the CDN address to as well. For example, when loading destinations/remote-plugins/etc. ... searching the codebase for `cdn = window.analytics?._cdn ?? getCDN()` will give you pointers to that. Can we update those as well? Happy to tackle internally as well if it looks tedious to support.

yordis:
@pooyaj happy to fix everything, I wasn't aware of them. Let me dig into it.

silesky:
@yordis, thank you for your contribution, this is useful! I'm continuing your work in this PR https://github.com/segmentio/analytics-next/pull/414, and am working to bring it across the finish line.

yordis:
@silesky please feel free to take over if you finish first, I don't mind at all, I was about to work on this over the weekend. So if you have bandwidth before, keep going.

silesky:
Continued in https://github.com/segmentio/analytics-next/pull/414

PR #413, opened 2022-03-28
Add support for segment inspector

zikaari:
This patch adds the interface that allows inspector clients to connect
and monitor all Segment activity happening on the page. This API is not
related to the events API (over Emitter interface) by design, as both of
these API's serve a very distinct purpose and will be developed at their
own pace without breaking each other's behaviour. End-users of the library
are not advised to use this interface in their production code and should
continue to rely on the existing events API.

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -43,7 +43,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "25.6 KB"
+      "limit": "25.7 KB"
     }
   ],
   "dependencies": {
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -59,6 +59,7 @@
     "unfetch": "^4.1.0"
   },
   "devDependencies": {
+    "@segment/inspector-core": "^1.0.0",
     "@size-limit/preset-big-lib": "^7.0.8",
     "@types/flat": "^5.0.1",
     "@types/fs-extra": "^9.0.2",
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -43,7 +43,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "25.7 KB"
+      "limit": "25.8 KB"
     }
   ],
   "dependencies": {
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -29,6 +29,7 @@ import type {
 import { version } from '../../generated/version'
 import { PriorityQueue } from '../../lib/priority-queue'
 import { getGlobal } from '../../lib/get-global'
+import { inspectorHost } from '../inspector'
 
 const deprecationWarning =
   'This is being deprecated and will be not be available in future releases of Analytics JS'
@@ -113,6 +114,13 @@ export class Analytics extends Emitter {
     this.eventFactory = new EventFactory(this._user)
     this.integrations = options?.integrations ?? {}
     this.options = options ?? {}
+
+    inspectorHost.start({
+      user: {
+        id: this.user().id() || null,
+        traits: this.user().traits(),
+      },
+    })
 
     autoBind(this)
   }
@@ -320,6 +328,13 @@ export class Analytics extends Emitter {
     callback?: Callback
   ): Promise<DispatchedEvent> {
     const ctx = new Context(event)
+
+    inspectorHost.trace({
+      stage: 'triggered',
+      id: ctx.id,
+      event: event as any,
+      timestamp: new Date().toISOString(),
+    })
 
     if (isOffline() && !this.options.retryQueue) {
       return ctx
--- a/packages/browser/src/core/inspector/index.ts
+++ b/packages/browser/src/core/inspector/index.ts
@@ -1,0 +1,9 @@
+import { Inspector } from 'segment-inspector-core'
+
+// FIXME: Replace this with `globalThis` to support all environments
+const env = typeof window !== 'undefined' ? window : null
+
+export const inspectorHost: Inspector = {
+  start: (...args) => (env as any)?.['__SEGMENT_INSPECTOR__']?.start(...args),
+  trace: (...args) => (env as any)?.['__SEGMENT_INSPECTOR__']?.trace(...args),
+}
--- a/packages/browser/src/core/inspector/index.ts
+++ b/packages/browser/src/core/inspector/index.ts
@@ -1,4 +1,5 @@
 import type { Inspector } from '@segment/inspector-core'
+import { getGlobal } from '../../lib/get-global'
 
 declare global {
   interface Window {
@@ -6,8 +7,7 @@ declare global {
   }
 }
 
-// FIXME: Replace this with `globalThis` to support all environments
-const env = typeof window !== 'undefined' ? window : null
+const env = getGlobal()
 
 export const inspectorHost: Inspector = {
   start: (...args) => (env as any)?.['__SEGMENT_INSPECTOR__']?.start(...args),
--- a/packages/browser/src/core/inspector/index.ts
+++ b/packages/browser/src/core/inspector/index.ts
@@ -1,4 +1,10 @@
-import { Inspector } from 'segment-inspector-core'
+import type { Inspector } from '@segment/inspector-core'
+
+declare global {
+  interface Window {
+    __SEGMENT_INSPECTOR__?: Inspector
+  }
+}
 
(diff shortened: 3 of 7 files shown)


src/inspector/index.ts
```diff
@@ -0,0 +1,66 @@
+import { Context, Integrations, LegacySettings } from '..'
+
+// FIXME: Import these types directly from `@segment/inspector-core`
+type SegmentEvent = object
+type Inspector = {
+  start: (config: object) => void
+  trace: (eventPhase: object) => void
+}
+
+export const inspectorHost = (() => {
```
chrisradek, line 10: I'm curious why you used an IIFE to set the value of `inspectorHost`. Since this file is already a module you could just export `connectInspector` directly.

Thinking about this more, it might be useful to have `inspectorHost` be a function that takes an `Analytics` object. Then you can rely on analytics to resolve some of your values for you, for example: `analytics.user.id()` instead of `localStorage.getItem('ajs_user_id')`. The benefit for this example would be you'd get the correct user id from analytics itself, be it stored in cookies, localstorage, or in memory!

chrisradek (commented):
Looking good! Planning to keep working in this branch for now?

src/inspector/index.ts
zikaari: Those are actually pretty good points!

I'll experiment with binding the host directly to an Analytics instance when I get back from vacation.

_(You too are welcome to try it out! there are tests already which will ensure we don't mess up compliance)_

zikaari:
> Looking good! Planning to keep working in this branch for now?

Yes. Let's do that for some time. Once we have 2-3 sources operational with this addition, we can do a full roll-out by merging into master.

packages/browser/src/core/inspector/index.ts
```diff
@@ -0,0 +1,9 @@
+import { Inspector } from 'segment-inspector-core'
```
silesky, line 1: I think any isomorphism would be better baked into `segment-inspector-core` (core is the keyword), rather than the responsibility of the consumer

packages/browser/src/core/analytics/index.ts
```diff
@@ -115,6 +116,13 @@ export class Analytics extends Emitter {
     this.integrations = options?.integrations ?? {}
     this.options = options ?? {}
 
+    inspectorHost.start({
```
silesky, line 119: Can this be imported and loaded lazily (e.g. on devtools open / connection) in order that it doesn't affect bundle size?

packages/browser/src/core/inspector/index.ts
zikaari: I don't follow. Can you elaborate?

packages/browser/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@segment/analytics-next",
-  "version": "1.38.0",
+  "version": "1.39.0",
```
silesky, line 3: Per my understanding of the release process, we don't want to bump the version here.

packages/browser/src/core/inspector/index.ts
```diff
@@ -0,0 +1,9 @@
...
+
+// FIXME: Replace this with `globalThis` to support all environments
+const env = typeof window !== 'undefined' ? window : null
+
+export const inspectorHost: Inspector = {
```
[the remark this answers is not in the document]
zikaari: I added that wrapper that so that each caller wouldn't need to add optional chaining on every call eg: `inspectorHost?.trace(...)`, but on a second though, it's not a big deal. Will remove the wrapper.

```
if (typeof window !== 'undefined') {
   throw new Error("This can only be run in browser.")
}
```

We don't want to throw an errors that could be fatal for Analytics' operations. In case of Inspector, it is a "non-essential", so a silent fail is preferable.

packages/browser/package.json
zikaari: This is temporary. Will be reverted right before merge.

packages/browser/src/core/inspector/index.ts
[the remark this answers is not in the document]
silesky: I see, so it was a wrapper for the sake of a noop

silesky:
nice job @zikaari. I love that the extension injects an interface.

silesky:
I like the idea of the plugin exposing an interface, that's pretty neat!

packages/browser/src/core/inspector/index.ts
silesky:
```suggestion
import type { Inspector } from 'segment-inspector-core'

```

packages/browser/src/core/analytics/index.ts
silesky: @zikaari Maybe not worth investigating, I think the stakes are super low.

The way the module is being injected -- means that it actually won't be loaded unless you have the devtools extension installed (basically, you're a developer =)

Nice job.

packages/browser/src/core/inspector/index.ts
silesky: @zikaari ignore the isomorphism bit, realizing that the Inspector was just a type changed my p-o-v =)

silesky: > I don't follow. Can you elaborate?

I changed my mind =)

packages/browser/src/core/queue/event-queue.ts
```diff
@@ -265,6 +266,13 @@ export class EventQueue extends Emitter {
       }
     }
 
+    inspectorHost.trace({
+      stage: 'enriched',
+      id: ctx.id,
+      event: ctx.event as any,
```
pooyaj, line 272: Does it worth just passing in the whole `ctx` object instead. Given the `ctx` will have both `id` and the `event` ... also if in the future we add more to the context, we don't have to change the trace signature

pooyaj (approved):
Amazing 🙌 love the dependency injection

packages/browser/src/core/queue/event-queue.ts
chrisradek: `ctx` also has the `logs` and `metrics` for each plugin that's invoked which seems useful for inspectHost :smile:

zikaari: Good point, again! Will change that up.

packages/browser/src/core/analytics/index.ts
chrisradek: Not sure it's worth doing. I think to make this a plugin, you'd need to make it the `utility` type and make the `load` method start the host and attach event listeners for the API calls (track, identify, page, group, alias, screen), then call `inspectorHost.trace` with the context.

But I don't fully understand why `inspectorHost.trace` is being called in-between plugin lifecycles and/or if the context has enough data so those calls can be removed and triggered once when the event is delivered instead.

chrisradek: There might be a benefit to allowing the `inspectorHost` to be passed to the `Analytics` constructor. Right now it only works in browsers - and we're starting to break up our repo so that the `core` isn't specific to a single runtime.

silesky: Yep. Much of “core” is going to be moved into its own package, and what will remain will be browser-specific. Goal is to remove anything that references the window / document object from core.

silesky:
@zikaari
- If we do decide to publish inspector core to npm, Just want to make sure we publish it as @segment/package-name. Literally just ran into this issue with an old employee squatting on analytics-next.
- We might want to think about bringing in the entire inspector into this monorepo for all of the reasons that monorepos are nice, rather than publishing inspector core at all, since the SDK is the primary (only) dependent.

packages/browser/src/core/queue/event-queue.ts
zikaari: This will require extensive refactoring. We can postpone this until it becomes a necessity _(i.e when we do show extended metrics and logs in the UI)_

packages/browser/src/core/inspector/index.ts
```diff
@@ -0,0 +1,15 @@
+import type { Inspector } from '@segment/inspector-core'
+
+declare global {
+  interface Window {
+    __SEGMENT_INSPECTOR__?: Inspector
+  }
+}
+
+// FIXME: Replace this with `globalThis` to support all environments
```
silesky, line 9: Can we use this instead? https://github.com/segmentio/analytics-next/blob/master/packages/browser/src/lib/get-global.ts#L2

zikaari: cool. didn't know about this. fixed up!

silesky (approved):

PR #414, opened 2022-03-29
feature: Add support for specifying CDN URL during .load call

silesky:
This a continuation of https://github.com/segmentio/analytics-next/pull/407 which addresses https://github.com/segmentio/analytics-next/issues/362.

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -39,6 +39,7 @@ const _analytics = globalAny.analytics
 
 export interface AnalyticsSettings {
   writeKey: string
+  cdn?: string
   timeout?: number
   plugins?: Plugin[]
   [key: string]: unknown
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -39,10 +39,8 @@ const _analytics = globalAny.analytics
 
 export interface AnalyticsSettings {
   writeKey: string
-  cdn?: string
   timeout?: number
   plugins?: Plugin[]
-  [key: string]: unknown
 }
 
 export interface InitOptions {
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -62,9 +62,9 @@ export interface AnalyticsBrowserSettings extends AnalyticsSettings {
   cdnSettings?: LegacySettings & Record<string, unknown>
 }
 
-export function loadLegacySettings(writeKey: string): Promise<LegacySettings> {
-  const cdn = window.analytics?._cdn ?? getCDN()
-  return fetch(`${cdn}/v1/projects/${writeKey}/settings`)
+export function loadLegacySettings(writeKey: string, cdn?: string): Promise<LegacySettings> {
+  const baseUrl = cdn ?? window.analytics?._cdn ?? getCDN()
+  return fetch(`${baseUrl}/v1/projects/${writeKey}/settings`)
     .then((res) => res.json())
     .catch((err) => {
       console.warn('Failed to load settings', err)
@@ -232,7 +232,7 @@ export class AnalyticsBrowser {
     options: InitOptions = {}
   ): Promise<[Analytics, Context]> {
     const legacySettings =
-      settings.cdnSettings ?? (await loadLegacySettings(settings.writeKey))
+      settings.cdnSettings ?? (await loadLegacySettings(settings.writeKey, settings.cdn))
 
     const retryQueue: boolean =
       legacySettings.integrations['Segment.io']?.retryQueue ?? true
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1,5 +1,5 @@
 import { getProcessEnv } from './lib/get-process-env'
-import { getCDN } from './lib/parse-cdn'
+import { getCDN, setGlobalCDNUrl } from './lib/parse-cdn'
 
 import fetch from 'unfetch'
 import { Analytics, AnalyticsSettings, InitOptions } from './analytics'
@@ -60,10 +60,18 @@ export interface AnalyticsBrowserSettings extends AnalyticsSettings {
    * for the source.
    */
   cdnSettings?: LegacySettings & Record<string, unknown>
+  /**
+   * If provided, will override the default Segment CDN (https://cdn.segment.com) for this application.
+   */
+  cdnURL?: string
 }
 
-export function loadLegacySettings(writeKey: string, cdn?: string): Promise<LegacySettings> {
-  const baseUrl = cdn ?? window.analytics?._cdn ?? getCDN()
+export function loadLegacySettings(
+  writeKey: string,
+  cdnURL?: string
+): Promise<LegacySettings> {
+  const baseUrl = cdnURL ?? getCDN()
+
   return fetch(`${baseUrl}/v1/projects/${writeKey}/settings`)
     .then((res) => res.json())
     .catch((err) => {
@@ -231,8 +239,11 @@ export class AnalyticsBrowser {
     settings: AnalyticsBrowserSettings,
     options: InitOptions = {}
   ): Promise<[Analytics, Context]> {
+    // this is an ugly side-effect, but it's for the benefits of the plugins that get their cdn via getCDN()
+    if (settings.cdnURL) setGlobalCDNUrl(settings.cdnURL)
     const legacySettings =
-      settings.cdnSettings ?? (await loadLegacySettings(settings.writeKey, settings.cdn))
+      settings.cdnSettings ??
+      (await loadLegacySettings(settings.writeKey, settings.cdnURL))
 
     const retryQueue: boolean =
       legacySettings.integrations['Segment.io']?.retryQueue ?? true
--- a/src/lib/parse-cdn.ts
+++ b/src/lib/parse-cdn.ts
@@ -1,31 +1,48 @@
 import { embeddedWriteKey } from './embedded-write-key'
-const regex = /(https:\/\/.*)\/analytics\.js\/v1\/(?:.*?)\/(?:platform|analytics.*)?/
 
-export function getCDN(): string {
-  let cdn: string | undefined = undefined
-
+const analyticsScriptRegex = /(https:\/\/.*)\/analytics\.js\/v1\/(?:.*?)\/(?:platform|analytics.*)?/
+const getCDNUrlFromScriptTag = (): string | undefined => {
+  let cdn: string | undefined
   const scripts = Array.prototype.slice.call(
     document.querySelectorAll('script')
   )
-
   scripts.forEach((s) => {
     const src = s.getAttribute('src') ?? ''
-    const result = regex.exec(src)
+    const result = analyticsScriptRegex.exec(src)
 
     if (result && result[1]) {
       cdn = result[1]
     }
   })
+  return cdn
+}
 
-  // it's possible that the CDN is not found in the page because:
-  // - the script is loaded through a proxy
-  // - the script is removed after execution
-  // in this case, we fall back to the default Segment CDN
-  if (!cdn) {
+const getGlobalCDNUrl = (): string | undefined => {
+  return window.analytics?._cdn
+}
+
+export const setGlobalCDNUrl = (cdn: string) => {
+  if (window.analytics) {
+    window.analytics._cdn = cdn
+  }
+}
+
+export const getCDN = (): string => {
+  const globalCdnUrl = getGlobalCDNUrl()
+
+  if (globalCdnUrl) return globalCdnUrl
+
+  const cdnFromScriptTag = getCDNUrlFromScriptTag()
+
+  if (cdnFromScriptTag) {
+    return cdnFromScriptTag
+  } else {
+    // it's possible that the CDN is not found in the page because:
+    // - the script is loaded through a proxy
+    // - the script is removed after execution
+    // in this case, we fall back to the default Segment CDN
     return `https://cdn.segment.com`
   }
-
-  return cdn
 }
 
 /**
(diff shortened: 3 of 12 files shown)


src/lib/parse-cdn.ts
```diff
@@ -1,31 +1,48 @@
 import { embeddedWriteKey } from './embedded-write-key'
 const regex = /(https:\/\/.*)\/analytics\.js\/v1\/(?:.*?)\/(?:platform|analytics.*)?/
 
-export function getCDN(): string {
-  let cdn: string | undefined = undefined
+export class CDN {
```
yordis, line 4: I would push back in adding more classes to Segment, classes are not tree-shakeable nor do they allow minify method name.

Instead, export functions as it was before that works with a given CDN data structure (string)

src/lib/parse-cdn.ts
```diff
@@ -1,31 +1,51 @@
...
+
+  if (cdnFromScriptTag) {
+    setGlobalCDNUrl(cdnFromScriptTag)
+    return cdnFromScriptTag
+  } else {
+    // it's possible that the CDN is not found in the page because:
+    // - the script is loaded through a proxy
+    // - the script is removed after execution
+    // in this case, we fall back to the default Segment CDN
+    return `https://cdn.segment.com`
```
yordis, line 47: Adding such a fallback means extra bytes of a string that can't be minified.

src/lib/parse-cdn.ts
```diff
@@ -1,31 +1,51 @@
...
+
+export const getCDN = (): string => {
+  const globalCdnUrl = getGlobalCDNUrl()
+
+  if (globalCdnUrl) return globalCdnUrl
+
+  const cdnFromScriptTag = getCDNUrlFromScriptTag()
+
+  if (cdnFromScriptTag) {
+    setGlobalCDNUrl(cdnFromScriptTag)
```
yordis, line 40: Personally, keeping CQS in mind, I rather move this to `setGlobalCDNUrl` where the `cdn` could be `undefined` or `null`.

```ts
export const setGlobalCDNUrl = (cdn?: string) => {
  if (window.analytics) {
    window.analytics._cdn = cdn ?? getCDNUrlFromScriptTag()
  }
}
```

You deal with the command side once when you are trying to set things up.

Especially that, if I am not mistaken, ideally `getCDNUrlFromScriptTag` would be removed altogether.

For example,

```ts
import { getCDNUrlFromScriptTag } from '...';
AnalyticsBrowser.load({
      cdn: getCDNUrlFromScriptTag(), // notice that I am using when I initialize things
      writeKey
})
```

Or at least I hope Segment move to that, so the bundle sizes are taken into consideration.

yordis (commented):
@silesky one of the reasons why I didn't finish it right away the way you did it was because I wanted to figure out how to remove that `window.analytics._cdn`.

I am curious about the technical limitations but would be much better to avoid having that global object if possible. I am not sure.

Any thoughts on that?

src/lib/parse-cdn.ts
yordis: > I’ve gone back to explicitly setting the global


I love it!

src/lib/parse-cdn.ts
```diff
@@ -1,31 +1,50 @@
...
-  // - the script is loaded through a proxy
-  // - the script is removed after execution
-  // in this case, we fall back to the default Segment CDN
-  if (!cdn) {
-    return `https://cdn.segment.com`
+const getGlobalCDNUrl = (): string | undefined => {
+  if (window.analytics) {
+    return window.analytics._cdn
   }
+}
```
yordis, line 24:
```suggestion
const getGlobalCDNUrl = (): string | undefined => {
  return window?.analytics?._cdn
}
```

src/lib/parse-cdn.ts
silesky: Per the comment (which isn't mine, btw), we still need to have the fallback in case the script is removed after execution.

 If there's an easy way to keep this default constant in a way that you think is better for minification, I'm open to it. Otherwise, I would lean towards keeping the old behavior.

src/lib/parse-cdn.ts
silesky: > @silesky one of the reasons why I didn't finish it right away the way you did it was because I wanted to figure out how to remove that `window.analytics._cdn`.
> 
> I am curious about the technical limitations but would be much better to avoid having that global object if possible. I am not sure.
> 
> Any thoughts on that?

I agree that the global is ugly, and I was able to further minimize its footprint by removing  the setGlobal call from the load-browser call (after doing some digging, looking at tests, and being satisfied that there weren't any more implicit dependencies.)

The window.analytics._cdn is the current source of truth for standalone.ts.  and, while maybe undocumented, is part of the existing API (i.e. being able to set window.analytics._writeKey the same way). We couldn't get rid of it without breaking downstream consumers, so it's probably here to stay.

silesky:
> @silesky one of the reasons why I didn't finish it right away the way you did it was because I wanted to figure out how to remove that `window.analytics._cdn`.
>
> I am curious about the technical limitations but would be much better to avoid having that global object if possible. I am not sure.
>
> Any thoughts on that?

@yordis I agree that the global is ugly, and I was able to further minimize its footprint by removing the setGlobal call from the load-browser call (after doing some digging, looking at tests, and being satisfied that there weren't any more implicit dependencies.)

The window.analytics._cdn is the current source of truth for standalone.ts. and, while maybe undocumented, is part of the existing API (i.e. being able to set window.analytics._writeKey the same way). We couldn't get rid of it without breaking downstream consumers; so it's probably here to stay.

As long as we don't add any post-initialization mutation behavior and keep it a user-controlled configuration variable, I'm satisfied with it.

yordis:
> We couldn't get rid of it without breaking downstream consumers; so it's probably here to stay.

Related to, https://github.com/segmentio/analytics-next/pull/391#issuecomment-1085083103

It seems that `standalone.ts` is where we get to decide to avoid breaking changes or be a more singleton/global object/object-oriented. Which it makes sense, it is the CDN version I will assume.

While the other areas could avoid such problems altogether. 🤔

silesky:
> > We couldn't get rid of it without breaking downstream consumers; so it's probably here to stay.
> 
> Related to, [#391 (comment)](https://github.com/segmentio/analytics-next/pull/391#issuecomment-1085083103)
> 
> It seems that `standalone.ts` is where we get to decide to avoid breaking changes or be a more singleton/global object/object-oriented. Which it makes sense, it is the CDN version I will assume.
> 
> While the other areas could avoid such problems altogether. 🤔

The _cdn property is also something that can currently be used on the library as well.

For example:
```ts
import { AnalyticsBrowser, Analytics,  } from "@segment/analytics-next";

const useAnalytics = (writeKey: string, staging = true) => {
  const [analytics, setAnalytics] = React.useState<Analytics | undefined>(
    undefined
  );
  useEffect(() => {
    const loadAnalytics = async () => { // this is tested and works
       (globalThis as any).analytics = {
          _cdn: "https://cdn.segment.build",
        };
        }
        let [response, ctx] = await AnalyticsBrowser.load({
          writeKey,
        });
....
```

silesky:
> > I’m not against it, but it’s something we’d probably have to plan for / require a lot more discussion. I can see reasons why one would even argue that having both configurations options are useful.
> 
> I get it, please take all my feedback and thoughts with a grain of salt, remember that I don't work full-time on these things nor I am part of those conversations, trying my best to help.
> 
> Definitely excited about the incoming versions.

Great! I want to reiterate that your contribution and then participation in this code review has been fun and valuable, and I hope you’ll stick around!

src/standalone-analytics.ts
```diff
@@ -61,19 +61,20 @@ function getWriteKey(): string | undefined {
...
+  const options = window.analytics?._loadOptions ?? {}
   if (!writeKey) {
     console.error(
       'Failed to load Write Key. Make sure to use the latest version of the Segment snippet, which can be found in your source settings.'
     )
     return Promise.resolve()
   }
 
-  return AnalyticsBrowser.standalone(
+  const standaloneAnalytics = (await AnalyticsBrowser.standalone(
```
pooyaj, line 74: looks way cleaner ✨✨

pooyaj (approved):

src/analytics.ts
```diff
@@ -39,6 +39,7 @@ const _analytics = globalAny.analytics
 
 export interface AnalyticsSettings {
   writeKey: string
+  cdn?: string
```
chrisradek, line 42: I believe (but correct me if I'm wrong) that this field is only useful when using AnalyticsBrowser, not AnalyticsNode. There's an `AnalyticsBrowserSettings` interface in `browser.ts` that might make more sense to add this to then - I don't think the `Analytics` class actually uses this setting at all iirc.

chrisradek: Sorry, one more comment - can you add a doc tag for this as well to explain what this field is for? (We haven't been great about doing that thus far). I kind of prefer calling it `cdnEndpoint` but can be persuaded since it stays consistent with the `analytics._cdn` global.

src/browser.ts
```diff
@@ -62,9 +62,14 @@ export interface AnalyticsBrowserSettings extends AnalyticsSettings {
...
 }
 
-export function loadLegacySettings(writeKey: string): Promise<LegacySettings> {
-  const cdn = window.analytics?._cdn ?? getCDN()
-  return fetch(`${cdn}/v1/projects/${writeKey}/settings`)
+export function loadLegacySettings(
+  writeKey: string,
+  cdn?: string
+): Promise<LegacySettings> {
+  if (cdn) setGlobalCDNUrl(cdn) // this is an ugly side-effect, but it's for the benefits of the plugins that get their cdn via getCDN()
```
chrisradek, line 69: What do you think about moving the `setGlobalCDNUrl` to the `AnalyticsBrowser.load` method? I think the side-effect here of setting the global CDN on `analytics` makes more sense from the class that's actually creating that `analytics` object.

This function could be called multiple times with different CDN values (not something I'd ever expect to actually hapen), so removing the side-effect here makes this feel safer.

src/standalone-analytics.ts
chrisradek: Does it make sense to assign directly to window.analytics and skip creating the `standaloneAnalytics` variable?

chrisradek (commented):
Overall looks great! Appreciate all the tests you added as well!

src/browser.ts
silesky: Sure!

src/standalone-analytics.ts
silesky: @chrisradek Makes sense!

silesky: @chrisradek Seem kind of a smell  that AnalyticsBrowser.standalone returns an analytics type that we have to make a type assertion about? Is there an objection to me refactoring that?

chrisradek: No objection - not sure how worth it is though :smile: `AnalyticsBrowser.standalone` uses the correct return type, it's just we internally declare the global analytics object to be what you'd get from the Segment source snippet.

If you have really strong feelings about it you can probably just remove the Symbol.iterator from the `StandaloneAnalytics` type, clean up the unused types and everything _should_ just work then.

src/analytics.ts
```diff
@@ -41,7 +41,6 @@ export interface AnalyticsSettings {
   writeKey: string
   timeout?: number
   plugins?: Plugin[]
-  [key: string]: unknown
 }
```
silesky, line 44: Tightened up this type.

src/browser.ts
```diff
@@ -60,11 +61,20 @@ export interface AnalyticsBrowserSettings extends AnalyticsSettings {
    * for the source.
    */
   cdnSettings?: LegacySettings & Record<string, unknown>
+  /**
+   * If provided, will override the default Segment CDN (https://cdn.segment.com) for this application.
+   * Typically used to proxy client-side requests from analytics.js through a specific server.
```
chrisradek, line 66: `client-side requests` makes me think of requests hitting the tracking API which this doesn't actually include. It's just for requests that load source settings and the dynamically loaded pieces of the library.

Maybe something like:
> Typically used to proxy Source settings and analytics.js...

src/browser.ts
```diff
@@ -231,8 +241,10 @@ export class AnalyticsBrowser {
     settings: AnalyticsBrowserSettings,
     options: InitOptions = {}
   ): Promise<[Analytics, Context]> {
+    if (settings.cdnURL) setGlobalCDNUrl(settings.cdnURL) // this is an ugly side-effect, but it's for the benefits of the plugins that get their cdn via getCDN()
```
chrisradek, line 244: Nit: Can you move the comment above the line it's referring to? Keeps the column count down a bit :smile:

src/browser.ts
```diff
@@ -278,7 +290,9 @@ export class AnalyticsBrowser {
   static standalone(
     writeKey: string,
     options?: InitOptions
-  ): Promise<Analytics> {
-    return AnalyticsBrowser.load({ writeKey }, options).then((res) => res[0])
+  ): Promise<StandaloneAnalytics> {
```
chrisradek, line 293: I don't think the return type of this should be changed. If I recall, this function returns an instance of `Analytics`, not `StandaloneAnalytics`. The main difference being that `StandaloneAnalytics` is an iterable.

src/standalone-analytics.ts
```diff
@@ -2,23 +2,15 @@ import { Analytics, InitOptions } from './analytics'
...
 import { embeddedWriteKey } from './lib/embedded-write-key'
 
-type FunctionsOf<T> = {
-  [k in keyof T]: T[k] extends Function ? T[k] : never
-}
-
-type AnalyticsMethods = Pick<Analytics, keyof FunctionsOf<Analytics>>
-
-type StandaloneAnalytics = Analytics & {
+export type AnalyticsSnippet = Analytics & {
```
chrisradek, line 5: Just curious, does this need to be exported? Fine if we do need to for our tests, just would have thought the global declaration would have us covered.

chrisradek (approved):
Looks good to me!

PR #416, opened 2022-03-31
Fixes the QA job

chrisradek:
In our QA tests there was one source loading a destination in debug mode that was overwriting `Date.prototype.toJSON`. Unfortunately we want to overwrite `Date.prototype.toJSON` to return a simple string in our QA job so it's easier to compare the results from 2 browsers.

This change simply freezes the `Date.prototype` object after we modify so nothing else can.

### DIFF
--- a/qa/lib/runner.ts
+++ b/qa/lib/runner.ts
@@ -95,6 +95,7 @@ export async function run(params: ComparisonParams) {
     // This forces every timestamp to look exactly the same
     await page.evaluate('Date.prototype.toJSON = () => "<date>";')
     await page.evaluate('Date.prototype.getTime = () => 1614653469;')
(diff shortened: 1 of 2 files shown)


qa/lib/runner.ts
```diff
@@ -95,6 +95,7 @@ export async function run(params: ComparisonParams) {
     // This forces every timestamp to look exactly the same
     await page.evaluate('Date.prototype.toJSON = () => "<date>";')
     await page.evaluate('Date.prototype.getTime = () => 1614653469;')
```
silesky, line 97: Good find. We might as well freeze the prototype of getTime as well I guess :-)

silesky (approved):

PR #418, opened 2022-04-01
fixes qa ci job by making integrations checking more resilient

chrisradek:
QA job was failing again. The main cause was due to how we check the event's `integrations` object between classic and next. `analytics-next` can include action-destinations which classic doesn't, so we're seeing 1-2 sources where `integrations` contains more data with `next` due to these destinations than `classic`.

Since `next` should contain everything `classic` has, I made the `integrations` check ensure that `next` has __at least_ the same properties as `classic`.

### DIFF
--- a/qa/lib/schema.ts
+++ b/qa/lib/schema.ts
@@ -10,19 +10,41 @@ declare global {
   }
 }
 
+function injectAncestorKeys(keys: string[]): string[] {
+  const allKeys = new Set(keys)
+  for (const key of keys) {
+    const parts = key.split('.')
+    if (parts.length === 1) continue
+
+    do {
+      // remove the last item from the parts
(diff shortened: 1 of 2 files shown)


qa/lib/schema.ts
```diff
@@ -10,19 +10,41 @@ declare global {
   }
 }
```
silesky, line 12: Do/while -- I kind of forgot about that construct! Fancy! Maybe since it's a bit imperative (and we're not testing), I would add this comment for a future self so they don't have to follow the logic:
```suggestion
// Example:  ['hello', 'foo.bar.baz'] -> ['hello', 'foo', 'foo.bar', 'foo.bar.baz']
```

silesky (approved):

PR #421, opened 2022-04-06
Applies analytics.timeout to user-provided callbacks

### LINKED ISSUE #409: Timeout not working
I cloned this repository and started up the local dev environment for testing events.

I updated the track event to output the default timeout setting and add timing events to the console:
```js
    // @ts-ignore
    console.log(analytics.settings)
    console.time('track')
    const ctx = await analytics.track(evt?.event ?? 'Track Event', evt)
    console.timeEnd('track')
```
analytics.settings.timeout is set to the default 300ms, but the track call returns after 60ms:

`track: 59.259765625 ms`

If I add a callback to the track call, the time is essentially the same:

`const ctx = await analytics.track(evt?.event ?? 'Track Event', evt, null, () => { console.timeLog('track')})`

I also don't have any luck changing the timeout with `window.analytics.timeout(1000)` or passing it into `AnalyticsBrowser.load`

Is setting the default timeout supported?

chrisradek:
Fixes #409 

## Background

Analytics.js v1 allows callback functions to be provided to event calls (e.g. `track`, `page`, etc). These callbacks get invoked after 300ms, or a user-provided timeout, and are meant to give analytics.js a chance to send events to destinations before invoking the callback. A common scenario for this is navigating to another page after calling `track`.

Analytics.js v2 (next) still allows callback functions to be provided to event calls as well, however now the callback is invoked immediately after the event has been sent to all destinations.

## Is this change still useful?

I can think of 1 scenario where analytics-next sends an event to a client-side destination that does some sort of batching on its own. In this case having the timeout delay the callback may provide some utility though it's still somewhat fragile unless the user understands what the destination's flush interval is.

Creating this PR in draft since it does make the timeout work similarly to a.js classic, but would like to know if the above scenario is actually realistic before merging. This PR will make the timeout apply only if the user specifies a callback - if no callback is provided then the timeout is ignored like it is today.


src/analytics.ts
```diff
@@ -309,17 +309,21 @@ export class Analytics extends Emitter {
...
       dispatched = await this.queue.dispatch(ctx)
     }
+    const endTime = Date.now()
+    const elapsedTime = endTime - startTime
 
     if (callback) {
       dispatched = await invokeCallback(
         dispatched,
         callback,
+        Math.max((this.settings.timeout ?? 300) - elapsedTime, 0),
```
pooyaj, line 326: smart 🧠👌👌

pooyaj (approved):

silesky (approved):
LGTM

PR #424, opened 2022-04-08
fix cdn race condition

silesky:
- I'm still evaluating if it's safe to remove the `setGlobalCDNUrl` call from `brower-umd.ts`  and `standalone.ts` (if anyone has any thoughts / I'm interested!).

### DIFF
--- a/example/pages/index.tsx
+++ b/example/pages/index.tsx
@@ -7,6 +7,8 @@ import JSONTree from 'react-json-tree'
 import faker from 'faker'
 import { shuffle } from 'lodash'
 import Table from 'rc-table'
+import { useLocalStorage } from '../utils/hooks/useLocalStorage'
+import { useDidMountEffect } from '../utils/hooks/useDidMountEffect'
 
 import {
   AnalyticsBrowserSettings,
@@ -36,24 +38,20 @@ const jsontheme = {
   base0F: '#a3685a',
 }
 
-// This function runs on dependency change but not on initial render.
-const useDidMountEffect = (func, deps) => {
-  const didMount = useRef(false)
-
-  useEffect(() => {
-    if (didMount.current) func()
-    else didMount.current = true
-  }, deps)
-}
-
 export default function Home(): React.ReactElement {
   const [analytics, setAnalytics] = useState<Analytics | undefined>(undefined)
   const [settings, setSettings] = useState<
     AnalyticsBrowserSettings | undefined
   >(undefined)
   const [analyticsReady, setAnalyticsReady] = useState<boolean>(false)
-  const [writeKey, setWriteKey] = useState<string>('')
-  const [url, setURL] = useState<string>('https://cdn.segment.com')
+  const [writeKey, setWriteKey] = useLocalStorage(
+    'segment_playground_write_key',
+    ''
+  )
+  const [url, setURL] = useLocalStorage(
+    'segment_playground_cdn_url',
+    'https://cdn.segment.com'
+  )
 
   useDidMountEffect(() => {
     fetchAnalytics()
@@ -86,18 +84,23 @@ export default function Home(): React.ReactElement {
   const [ctx, setCtx] = React.useState<Context>()
 
   async function fetchAnalytics() {
-    const [response, ctx] = await AnalyticsBrowser.load({
-      ...settings,
-      writeKey,
-    })
-
-    if (response) {
+    try {
+      const [response, ctx] = await AnalyticsBrowser.load({
+        ...settings,
+        writeKey,
+      })
       setCtx(ctx)
       setAnalytics(response)
       setAnalyticsReady(true)
       setEvent(newEvent())
       // @ts-ignore
       window.analytics = response
+    } catch (err) {
+      console.error(err)
+      setCtx(undefined)
+      setAnalytics(undefined)
+      setAnalyticsReady(false)
+      setEvent('')
     }
   }
 
--- a/src/__tests__/integration.test.ts
+++ b/src/__tests__/integration.test.ts
@@ -12,6 +12,7 @@ import { isOffline } from '../core/connection'
 import * as SegmentPlugin from '../plugins/segmentio'
 import jar from 'js-cookie'
 import { AMPLITUDE_WRITEKEY, TEST_WRITEKEY } from './test-writekeys'
+import { getCDN } from '../lib/parse-cdn'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 let fetchCalls: Array<any>[] = []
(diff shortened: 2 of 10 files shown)


example/pages/index.tsx
```diff
@@ -31,13 +37,57 @@ const jsontheme = {
...
+  }
+  return [storedValue, setValue] as const
+}
+
 export default function Home(): React.ReactElement {
   const [analytics, setAnalytics] = useState<Analytics | undefined>(undefined)
-  const [settings, setSettings] = useState<AnalyticsSettings | undefined>(
-    undefined
-  )
+  const [writeKey, setWriteKey] = useLocalStorage<string>('__dev_writeKey', '')
```
silesky, line 82: This is a somewhat unrelated change that I'm sneaking in -- I wanted to be able to have the write key input persist between page refreshes.

src/plugins/ajs-destination/loader.ts
```diff
@@ -42,6 +39,7 @@ export async function loadIntegration(
 ): Promise<LegacyIntegration> {
   const pathName = normalizeName(name)
   const obfuscatedPathName = obfuscatePathName(pathName, obfuscate)
+  const path = getNextIntegrationsURL()
```
silesky, line 42: calling this method outside of the global scope make them easier to reason about, especially since it's a wrapper around a global, we don't need to worry about stale references.

example/pages/index.tsx
```diff
@@ -31,13 +37,57 @@ const jsontheme = {
...
+  const [settings, setSettings] = useState<
+    AnalyticsBrowserSettings | undefined
+  >({
+    writeKey,
+    cdnURL: 'https://cdn.segment.com',
+  })
```
yordis, line 88:
```suggestion
  const [writeKey, setWriteKey] = useLocalStorage<string>('__dev_writeKey', '')
  const [settings, setSettings] = useState<
    AnalyticsBrowserSettings | undefined
  >({
    writeKey: process.env.NEXT_PUBLIC_SEGMENT_WRITE_KEY,
    cdnURL: process.env.NEXT_PUBLIC_SEGMENT_CDN_URL,
  })
```

I would not like to teach people to do this ... it will cause issues like caching the wrong value or somebody blocking the `localStorage` ... you named it.

You do you, but I rather teach people to put such information in the `env` file since that is probably where people should be injecting such thing, and it will fix your issue

src/lib/parse-cdn.ts
```diff
@@ -17,14 +17,17 @@ const getCDNUrlFromScriptTag = (): string | undefined => {
   return cdn
 }
 
+let _globalCDN: string | undefined // set globalCDN as in-memory singleton
```
yordis, line 20: I LOVE IT

src/lib/parse-cdn.ts
```diff
@@ -45,6 +48,12 @@ export const getCDN = (): string => {
   }
 }
 
+export const getNextIntegrationsURL = () => {
+  const cdn = getCDN()
+  const path = `${cdn}/next-integrations`
```
yordis, line 53:
```suggestion
  return `${cdn}/next-integrations`
```

Interesting, I am curious to learn what made you decide to use anchor variables style rather than inlining the return 🤔 Nothing here btw, I am just curious of the habit.

silesky: That’s a good question!

In part, it was copypasta from the refactor. 

On the broad topic, I’m pretty inconsistent, but I do feel that raw return make debugging harder for “complex” expressions like “return pipe(doFoo, doBar)”, since setting a breakpoint to capture a function’s return value can be tricky if not impossible for many debuggers, so you often have to go to the caller to set the breakpoint. I have no hard and fast rules — I think intermediate variables are one of those things that I don’t attach hard and fast rules to, it’s a tool to communicate developer intent.

Here, I think we could definitely do without the return, as it’s quite simple.

yordis: > On the broad topic, I’m pretty inconsistent, but I do feel that raw return make debugging harder for “complex” expressions like “return pipe(doFoo, doBar)”,

Right right, I can see that some debuggers allow you to step into the return statement itself, while the part is to step out and see the value there rather than caring at the function call level ... 

I see I see ... just trying to learn about you in general.

Thank you for responding thou 💜

example/pages/index.tsx
silesky: - I didn't even consider that people might want to override cdnURL, but that's a good idea!
- on the subject of writeKey -- on a regular application, I agree. However, this is a playground for debugging (run `make dev` to see) -- so there's a writeKey input field that allows people to swap out writeKeys at runtime.

pooyaj:
@silesky I think the purpose of the `setGlobalCDNUrl` in the standalone etc. is to avoid consistently hitting the `getCDNUrlFromScriptTag` when calling `getCDN`, and only using `getCDNUrlFromScriptTag` once ( because it is a relatively costly DOM scan ).

example/pages/index.tsx
yordis: > I didn't even consider that people might want to override cdnURL, but that's a good idea!

Well, we have to do a proxy to Segment to bypass AdBlocker ... so probably most companies that know a bit about infra would do that most of the time 😄

yordis: > run make dev if you haven't yet to see what I mean!

I see I see, would call the directory "playground" because people are getting used to the idea of using `examples` directory to learn how to get up to speed with this ... which is my actually worry (not you, but others reading code out of context)

silesky: > > I didn't even consider that people might want to override cdnURL, but that's a good idea!
> 
> Well, we have to do a proxy to Segment to bypass AdBlocker ... so probably most companies that know a bit about infra would do that most of the time 😄

yep, I just meant the few of us who use the playground — I’d just seen it used to demo segment and develop locally directly against segment itself. Anyway, now it’s overridavle!

silesky: > > run make dev if you haven't yet to see what I mean!
> 
> I see I see, would call the directory "playground" because people are getting used to the idea of using `examples` directory to learn how to get up to speed with this ... which is my actually worry (not you, but others reading code out of context)

I think it’s meant to serve a duel role, but that’s a good point — it’s not typically what you think of as an ‘example’ in an examples repo!

silesky:
> @silesky I think the purpose of the `setGlobalCDNUrl` in the standalone etc. is to avoid consistently hitting the `getCDNUrlFromScriptTag` when calling `getCDN`, and only using `getCDNUrlFromScriptTag` once ( because it is a relatively costly DOM scan ).

Ah, If that's true that it's a cache, can we get rid of it? It seems like a premature optimization for a function that's only called basically during initialization

example/pages/index.tsx
danieljackins: love it!

src/__tests__/integration.test.ts
```diff
@@ -175,6 +176,26 @@ describe('Initialization', () => {
...
+            ...xt,
+            load: async () => {
+              expect(window.analytics).toBeUndefined()
+              expect(getCDN()).toContain(overriddenCDNUrl)
+            },
+          },
+        ],
+      })
+      expect(fetchCalls[0][0]).toContain(overriddenCDNUrl)
+      expect.assertions(3)
```
chrisradek, line 196: That's a neat trick :smile:

chrisradek (approved):
Looks great!

With regards to your question on if we need to setCdn for standalone/browser.umd, I think @pooyaj's explanation makes sense. If we can guarantee that we would only do the script tag lookup once with the changes you've made we can likely remove it - but I think to keep the scope of this PR focused I'd rather create a new issue or JIRA to track making that change/testing it.

PR #436, opened 2022-04-27
add ability to capture customer events before library init

silesky:
Any user method calls to .track / .identify / etc before analytics will be cached until analytics is finished loading.

One of the benefits of the snippet is that If a user calls analytics.track, we capture events for that user and then flush those events once the library has finished initializing. 

However, when using the npm library, a user must  currently wait for a .load in order to capture .track calls. This leads to 
 additional boilerplate as well as scenarios where user events might not get captured, as initialization can take a while. This feature offers a synchronous API that will cache events and then dispatch them when analytics finishes loading.

Other notes:
- this includes a fix for snippet users where both an `.on` event and its `callback` (e.g `on('track', callback)`  will fire twice if done before init.

This PR allows you to do
```tsx
import { useEffect, useState } from 'react'

// synchronous
const analytics = analyticsBrowser.load({writeKey}) // this loads an instance of AnalyticsBuffer 

const App = () => {
  return (
    <div>
        <button
          onClick={(e) => analytics.track("hello world", {prop: 1})}>
        >
          Track
        </button>
    </div>
  )
}
```

This change is backwards-compatible with the previous API.
Awaiting on AnalyticsBrowser will still return you [Analytics, Context] (like before).
```tsx
import { useEffect, useState } from 'react'

const App = () => {
  const [analytics, setAnalytics] = useState(undefined)
  useEffect(() => {
    const loadAnalytics = async () => {
      const [analytics] = await AnalyticsBrowser.load({ writeKey })
      setAnalytics(analytics)
    }
    loadAnalytics()
  }, [])

  return (
    <div>
        <button
          onClick={(e) => analytics?.track("hello world", {prop: 1})}>
        >
          Track
        </button>
    </div>
  )
}
```


src/__tests__/standalone-analytics.test.ts
```diff
@@ -220,26 +219,23 @@ describe('standalone bundle', () => {
...
   it('sets buffered event emitters before loading destinations', async (done) => {
-    // @ts-ignore ignore Response required fields
-    mocked(unfetch).mockImplementation((): Promise<Response> => fetchSettings)
+    mocked(unfetch).mockImplementation(() => fetchSettings as Promise<Response>)
 
     const operations: string[] = []
 
     track.mockImplementationOnce(() => operations.push('track'))
-    on.mockImplementationOnce(() => operations.push('on', 'on'))
+    on.mockImplementationOnce(() => operations.push('on'))
```
silesky, line 227: @arielsilvestri -- quick santity check here?  I had to update this test you wrote after fixing the double "window.analytics.on('foo') call" bug.

arielsilvestri: 👍

src/__tests__/pre-init.integration.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    AnalyticsBrowser._resetGlobalState()
+    mockFetchSettingsResponse()
+    delete window.analytics
+  })
+  test('load should instantiate an analytics object', async () => {
+    expect(AnalyticsBrowser.instance).toBeUndefined()
+    await AnalyticsBrowser.load({ writeKey })
+    expect(AnalyticsBrowser.instance).toBeDefined()
+  })
+  test('If a user sends a single preinitiozed track event, that event gets flushed', async () => {
```
arielsilvestri, line 36:
```suggestion
  test('If a user sends a single pre-initialized track event, that event gets flushed', async () => {
```

src/browser.ts
```diff
@@ -88,33 +95,118 @@ function hasLegacyDestinations(settings: LegacySettings): boolean {
...
+  | 'on'
+  | 'addSourceMiddleware'
+  | 'addIntegrationMiddleware'
+  | 'setAnonymousId'
+  | 'addDestinationMiddleware'
+
+/**
+ *  Represents a buffered method call that occurred before initialization.
+ */
+interface PreInitMethodCall<T extends PreInitMethodName = PreInitMethodName> {
```
arielsilvestri, line 120: I'm unfamiliar with the section that follows the `extends` keyword here. What's happening there?

silesky: Sure; it assigns the default type parameter of PreInitMethodName, so PreInitMethodCall can be called without an argument (if desired). 

Created a playground

https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwK7QLxQOQDMD2ONQA+mARnAE4YBQVokUAYgDwAqUEAHsBAHYAmAZ1iIUUdPGQQAfGKgBvKlCVQAbiIgAuKC0XKeSALZaAjFQC+NKgHoAVDdrhoAQVkLlq9VoBEuHF4Dcukr6RlDGgRZQNlYO9C7ozNh4GDI0tvZ00MxsnNz8QhKi4uoy6G7KapJaLIHuISYRVNGWAMY4PALAUADmEMAMeLKs7Fy8gsKSUgAUFN3VAJTVYjLlSuR9SOQ8ULPmre2dUCSyvf14U0n4OwXq8zRtHV1Ysj19AzgXvhh3QA

silesky: BTW, I just pushed a commit that simplified the types a bit!

chrisradek:
I think this feature has a lot of merit and I like the improvements, especially that track et al. now returns a promise in all cases.

I also like that this allows calling methods on a single object pre and post initialization. I think you could modify this a bit to avoid having to add a bunch of static properties to `AnalyticsBrowser` and treat it as a singleton.

If we exposed something like a `preload()` method, _that_ could return a stubbed `Analytics` that calls the real deal after `load()` is called on it. So instead of stubbing those methods on `AnalyticsBrowser` and then forwarding calls to the real `Analytics` when `instance` exists, you'd be stubbing a new object returned by `preload()` instead.

The benefits I see here are:
- Limit the surface of new APIs. This would be just the addition of 1-2 new APIs.
- Prevents treating `AnalyticsBrowser` as a singleton.
- Allows for multiple instances in the future - we've gone back and forth on this, but if we treat `AnalyticsBrowser` as a singleton we make it a lot harder to support if we decide to in the future.

I haven't gone through the code in depth, but I wanted to get your thoughts on this before digging deeper.

src/browser.ts
arielsilvestri: Oh I didn't know you could do default values in generics, neat!

src/browser.ts
```diff
@@ -88,33 +88,109 @@ function hasLegacyDestinations(settings: LegacySettings): boolean {
...
+
+function flushBuffered(analytics: Analytics, buffer: PreInitMethodCallBuffer) {
+  const callBufferedAnalyticsMethod = (methodCall: PreInitMethodCall) => {
+    const { method } = methodCall
+    // this guard is probably not needed.
+    if (typeof analytics[method] !== 'function') {
+      return console.warn(
+        `invariant error: method call "${method}" does not exist on analytics instance: ${analytics}`
+      )
+    }
```
arielsilvestri, line 171: Hmm, I'm also struggling to figure out how this could happen.

silesky: It was a guard in the existing code! I do think it's an impossible state; if the method doesn't exist, the user will get a reference error.

silesky:
> I think this feature has a lot of merit and I like the improvements, especially that track et al. now returns a promise in all cases.
> 
> I also like that this allows calling methods on a single object pre and post initialization. I think you could modify this a bit to avoid having to add a bunch of static properties to `AnalyticsBrowser` and treat it as a singleton.
> 
> If we exposed something like a `preload()` method, _that_ could return a stubbed `Analytics` that calls the real deal after `load()` is called on it. So instead of stubbing those methods on `AnalyticsBrowser` and then forwarding calls to the real `Analytics` when `instance` exists, you'd be stubbing a new object returned by `preload()` instead.
> 
> The benefits I see here are:
> 
> * Limit the surface of new APIs. This would be just the addition of 1-2 new APIs.
> * Prevents treating `AnalyticsBrowser` as a singleton.
> * Allows for multiple instances in the future - we've gone back and forth on this, but if we treat `AnalyticsBrowser` as a singleton we make it a lot harder to support if we decide to in the future.
> 
> I haven't gone through the code in depth, but I wanted to get your thoughts on this before digging deeper.

These are excellent concerns, and I'd love to talk more about the API. To respond to some of the smaller points (for anyone else reviewing)

- This feature looks a bit scary, but it's actually does not add any overhead to  implementing multi-instance in the future; I imagine that the static method API  is explicitly only for singleton users -- I could see a multi-instance API that requires analytics.load for the users who need it.

- The warning is not actually a stop sign, just a kind of "this is most likely a common bug" -- e.g., doing a .load directly in the render path.  
  - I could imagine we end up supporting multi instances with a single .load call (e.g., it returns a tuple), we could detect the multiple write keys and suppress that message, or make it a "development-build only" message, similar to development-only react library warnings. The model was the warning for google maps.

src/browser.ts
pooyaj: Yeah, probably almost impossible state. Maybe if someone uses a snippet with invalid methods, or manually add :shit: to the buffer.

src/browser.ts
```diff
@@ -123,26 +199,22 @@ async function flushBuffered(analytics: Analytics): Promise<void> {
...
-  const wa = window.analytics
-  const buffered =
-    // @ts-expect-error
-    wa && wa[0] ? [...wa] : []
 
-  const anon = buffered.find(([op]) => op === 'setAnonymousId')
-  if (anon) {
-    const [, id] = anon
-    analytics.setAnonymousId(id)
+async function flushPreBuffer(
```
pooyaj, line 203: I think some opportunity for optimization here (I know it is old code), but given `on` and `setAnonymousId` aren't async we can filter for both `setAnonymousId` and `on` in one go, and then call all of them sequentially in a `forEach`

pooyaj (commented):
Did the first pass, amazing Typescript 🧙‍♂️ two general comments: 

1- If we don't change the implementation, I wonder from usage standpoint we can suggest something like this as well ( in case someone doesn't like to use the class/static methods post-load or to do multi-instance, pretty much gives user the pre-buffer access instantly, and they eventually get access to the `instance` after load ): 
```
  const [analytics, setAnalytics] = useState(AnalyticsBrowser)
  useEffect(() => {
    const loadAnalytics = async () => {
      let [response] = await AnalyticsBrowser.load({ writeKey })
      setAnalytics(response)
      AnalyticsBrowser_resetGlobalState()
    }
    loadAnalytics()
  }, [writeKey])
```
(only concern, not sure how happy Typescript will be with the above ☝️)

2- Probably the only concern is bundle-size, not sure how much we can squeeze down 😬

src/browser.ts
```diff
@@ -235,13 +307,95 @@ async function registerPlugins(
...
+   * Analytics context instance initialized by AnalyticsBrowser.load.
+   */
+  static ctx?: Context = undefined
+
+  static preInitBuffer = new PreInitMethodCallBuffer()
+
+  /**
+   * Clear the global state. Useful mainly for testing.
+   */
+  static _resetGlobalState() {
```
yordis, line 378: Related to: https://github.com/segmentio/analytics-next/pull/391

I wish the codebase start to move away from these classes, especially if every method is a static method or variable.

Such practices make the codebase less tree-shakeable and treat the class as a bag of things (a file module) when you can achieve the same doing `import * as MyName from '...'` if you care about the aesthetic aspect of it.
Also, it is quite hard to follow, in the FP version you would have clear boundaries with data, state, and behavior.

I made the PR to showcase what it will look like and hopefully you would move there sooner than later. (Sentry@v7 is doing a lot of work on that front if you want validation, your products are really close when it comes to SDKs)

The bigger this class keeps getting, the harder it will be replace it or justify the breaking change, and the Sunk Cost Fallacy will be a thing.

My :2cents:

src/browser.ts
silesky: > Yeah, probably almost impossible state. Maybe if someone uses a snippet with invalid methods, or manually add 💩 to the buffer.

Someone would  have to manually update the snippet  copypasta  itself to add a fake method to the string array, and then call that fake method!

chrisradek:
@silesky I've been thinking about this more.

My main concerns about treating `AnalyticsBrowser` as the analytics class:
- It's not obvious that you have to call `load()` before the events start getting sent. Perhaps documentation is enough and at least it should be obvious that events aren't getting sent during dev.
- Calling `load()` multiple times causes unexpected results. What happens if I call `load()` a 2nd time with a new writeKey? Does that overwrite the instance that `AnalyticsBrowser` uses? (Currently it looks like yes). That can cause problems with loading destinations.

I suspect this new strategy will end up being what's used by consumers bundling the library in their application, so I just want to make sure we think about and limit opportunities for devs to accidentally do the wrong thing.

On a separate note I do agree with @yordis about not using classes with only static methods since that makes it harder for tools to do some tree-shaking. For example, `analytics.ts` and `standalone-analytics.ts` have a reference to `AnalyticsBrowser.load()`. With this change, the standalone bundler will end up pulling in all the static methods on AnalyticsBrowser that are added. If `load()` were its own method and we referenced that directly instead of `AnalyticsBrowser.load` I bet it'd reduce our bundle size compared to what it is with your changes here.

src/browser.ts
AbhiPrasad: Just chiming in with some context in regards to the Sentry side (https://github.com/getsentry/sentry-javascript)!

Static methods and attributes in classes are something we've stayed away from in our Browser SDK primarily because they just can't get minified. Something like `AnalyticsBrowser.trackSubmit()` minifies to `aa.trackSubmit()`, vs. for a function -> `function trackSubmit()` -> `function aa()`. This adds up over time (and suddenly makes function name length actually matter, never a good convo to have in code reviews). As an unrelated note, notice how this also applies for all objects in general, which makes destructuring and object keys all have a cost as well.

This is further exacerbated by the fact that the bundle increases even more when classes are transpiled down to es5 (though this is more an argument to just not use classes at all).

Then there's the general problem of bringing in unrelated logic whenever you import in the class to use static methods, but that is generally solved by keeping classes small + using dependency injection as much as possible (also just not using classes at all).

In general, we've had a lot of gains in moving private/static methods outside of a class (with shims for backwards compatibility until we can make the breaking change). See https://github.com/getsentry/sentry-javascript/pull/4281 as an example!

If you have any questions, please ping, happy to help out :)

src/browser.ts
arielsilvestri: It's possible that prior to your changes, maybe certain functions wouldn't exist for some edge case, but with all your stricter typings and setup, I think dropping this from future implementations is fine

silesky:
> It’s not obvious that you have to call load() before the events start getting sent. Perhaps documentation is enough and at least it should be obvious that events aren’t getting sent during dev.
Calling load() multiple times causes unexpected results. What happens if I call load() a 2nd time with a new writeKey? Does that overwrite the instance that AnalyticsBrowser uses? (Currently it looks like yes). That can cause problems with loading destinations.
I suspect this new strategy will end up being what’s used by consumers bundling the library in their application, so I just want to make sure we think about and limit opportunities for devs to accidentally do the wrong thing.


I think it’s common for web developers using a library to initialize something once at the beginning of their application.  They can even call it outside of the scope of a component. 
```tsx
AnalyticsBrowser.load()

const App = () => …
```
Most database / ORM clients (as default behavior) expect a user to connect once early for the duration of the application. For example, [mongoose](https://mongoosejs.com/) has a similar API, ditto Apollo Client, etc. This feature has no new behavior in that regard, and will have no affect on any existing users, other than they might be alerted if they're using unsupported behavior.

> On a separate note I do agree with @yordis about not using classes with only static methods since that makes it harder for tools to do some tree-shaking. For example, analytics.ts and standalone-analytics.ts have a reference to AnalyticsBrowser.load(). With this change, the standalone bundler will end up pulling in all the static methods on AnalyticsBrowser that are added. If load() were its own method and we referenced that directly instead of AnalyticsBrowser.load I bet it’d reduce our bundle size compared to what it is with your changes here.

I’m not so sure about this actually — AnalyticsBrowser.load still returns an instance (even if not referenced) and that code would still be part of the bundle.

Somewhat related — I did some naive comparison testing in our playground using webpack-bundle-analyzer before / after (master vs this branch) and there was essentially no difference in the bundle size. My hunch is that, while we might be able to save some bytes here or there, unlike, say, lodash or ramda, the number of functions / methods in this library is quite small but that each public API method has a large amounts of shared logic (e.g. priority queue) that can’t neccessarily be tree-shook away — at least, not without huge refactoring and API changes that’s out of the scope of this issue.

Minification is a separate issue that I'm more sympathetic to -- but for something like method names, it's still a savings on the level of bytes. Once out, we can iterate on the bundle size and structure, but it's much harder to change the public API.

arielsilvestri:
Would the analyzer be able to account for tree-shaking when externally consumed? or is it all just using locally built bundles when testing things out?

silesky:
> Would the analyzer be able to account for tree-shaking when externally consumed? or is it all just using locally built bundles when testing things out?

I don't 100% know. Unless we did something in our webpack prod build that was getting in the way, the analyzer would be able to tree shake local dependencies just as if they were external. The size will be off, but it works for comparison. If we wanted to be sophisticated, we'd have to test using `rollup` and `webpack ` under a range of different conditions. 

However, I just did this for fun -- there's not a reason why they would be different. If you're using classes in general, you're not thinking about tree shaking. My intuition says that adding static references to "track" etc are not going to increase bundle size beyond the method signature, since "track" etc are always being referenced in the same context in the instance returned by .load.

yordis:
> Somewhat related — I did some naive comparison testing in our playground using webpack-bundle-analyzer before / after (master vs this branch) and there was essentially no difference in the bundle size.

As I pointed out here: https://github.com/segmentio/analytics-next/pull/391 is just a simple change already reducing the bundle size. Is it minimal? yes, most definitely. But for such minimal work that is exactly the same just changing the habits already pays off. As an application developer, I would appreciate every byte you save.

I could continue with the rabbit hole, as @AbhiPrasad pointed out, Sentry already proved some of these concepts with huge benefits.

I can try to do what I did on https://github.com/getsentry/sentry-javascript/pull/4381 if you need more proof that it will eventually compound.

If I am not mistaken Sentry has seen up to 24% reduction and that is not even counting optimizations that you could continue doing, nor counting that it is not 100% of the work that could be done.

Anyway, it could a discussion for later, but as I pointed out "Sunk Cost" may make things harder to change, we are all humans after all.

❤️

chrisradek:
> I’m not so sure about this actually — AnalyticsBrowser.load still returns an instance (even if not referenced) and that code would still be part of the bundle.

For what it's worth, I was thinking more about the bundle we serve via our CDN. Taking what's on `master` and changing the static methods `load` and `standalone` to exported functions reduced the final bundle size ~20 bytes, not huge but pretty much free.

It looks like the bundle size for the CDN build with this PR increased ~550 bytes - about 2.2% increase which isn't huge but also for a feature not used by the CDN build. If there are improvements to buffering for the snippet included there then I'd concede the size increase for the CDN build is worth it.

silesky:
> > I’m not so sure about this actually — AnalyticsBrowser.load still returns an instance (even if not referenced) and that code would still be part of the bundle.
> 
> For what it's worth, I was thinking more about the bundle we serve via our CDN. Taking what's on `master` and changing the static methods `load` and `standalone` to exported functions reduced the final bundle size ~20 bytes, not huge but pretty much free.
> 
> It looks like the bundle size for the CDN build with this PR increased ~550 bytes - about 2.2% increase which isn't huge but also for a feature not used by the CDN build. If there are improvements to buffering for the snippet included there then I'd concede the size increase for the CDN build is worth it.

The other part of that 550 bytes of additional code (beyond the static method names) is that the code is the additional normalization for the snippet so we can use a single interface that represents a method call type (the PreInitBuffer class etc). This is just the extra mapping code you always need if you want to cleanly share logic between two separate interfaces. 

I'm not saying that we couldn't shave some bytes off, but yeah, not a ton of savings.

silesky:
> > Somewhat related — I did some naive comparison testing in our playground using webpack-bundle-analyzer before / after (master vs this branch) and there was essentially no difference in the bundle size.
> 
> As I pointed out here: #391 is just a simple change already reducing the bundle size. Is it minimal? yes, most definitely. But for such minimal work that is exactly the same just changing the habits already pays off. As an application developer, I would appreciate every byte you save.
> 
> I could continue with the rabbit hole, as @AbhiPrasad pointed out, Sentry already proved some of these concepts with huge benefits.
> 
> I can try to do what I did on [getsentry/sentry-javascript#4381](https://github.com/getsentry/sentry-javascript/pull/4381) if you need more proof that it will eventually compound.
> 
> If I am not mistaken Sentry has seen up to 24% reduction and that is not even counting optimizations that you could continue doing, nor counting that it is not 100% of the work that could be done.
> 
> Anyway, it could a discussion for later, but as I pointed out "Sunk Cost" may make things harder to change, we are all humans after all.
> 
> ❤️

As usual, thanks @yordis for the input. I don't think the static methods precludes us from refactoring to a more "functional" and "tree-shaking" friendly interface (per your previous PR).

I can image us maybe refactoring the library to have an entry point that is something like:
```
export { track } from '../track'
export { load } from '../load'

// for legacy users -- it looks like a class, but it's just a POJO
export AnalyticsBrowser = {
   load,
   track,
}

```

src/browser.ts
silesky: > Related to: #391
> 
> I wish the codebase start to move away from these classes, especially if every method is a static method or variable.
> 
> Such practices make the codebase less tree-shakeable and treat the class as a bag of things (a file module) when you can achieve the same doing `import * as MyName from '...'` if you care about the aesthetic aspect of it. Also, it is quite hard to follow, in the FP version you would have clear boundaries with data, state, and behavior.
> 
> I made the PR to showcase what it will look like and hopefully you would move there sooner than later. (Sentry@v7 is doing a lot of work on that front if you want validation, your products are really close when it comes to SDKs)
> 
> The bigger this class keeps getting, the harder it will be replace it or justify the breaking change, and the Sunk Cost Fallacy will be a thing.
> 
> My :2cents:

> > Somewhat related — I did some naive comparison testing in our playground using webpack-bundle-analyzer before / after (master vs this branch) and there was essentially no difference in the bundle size.
> 
> As I pointed out here: #391 is just a simple change already reducing the bundle size. Is it minimal? yes, most definitely. But for such minimal work that is exactly the same just changing the habits already pays off. As an application developer, I would appreciate every byte you save.
> 
> I could continue with the rabbit hole, as @AbhiPrasad pointed out, Sentry already proved some of these concepts with huge benefits.
> 
> I can try to do what I did on [getsentry/sentry-javascript#4381](https://github.com/getsentry/sentry-javascript/pull/4381) if you need more proof that it will eventually compound.
> 
> If I am not mistaken Sentry has seen up to 24% reduction and that is not even counting optimizations that you could continue doing, nor counting that it is not 100% of the work that could be done.
> 
> Anyway, it could a discussion for later, but as I pointed out "Sunk Cost" may make things harder to change, we are all humans after all.
> 
> ❤️

As usual, thanks @yordis for the input. I don't think the static methods precludes us from refactoring to a more "functional" and "tree-shaking" friendly interface (per your previous PR).

I can image us eventually refactoring the library to have an entry point that is something like:
```ts
export { track } from './track'
export { load } from './load'
// ... more methods


// for legacy users -- it looks like a class, but it's just a POJO
export const AnalyticsBrowser = {
   load, // or dynamic import?
   track,
// ... more methods 
}

```

 And thank you @AbhiPrasad, these is great feedback and I will definitely look @sentry's SDK moving forward.

AbhiPrasad: I’m available on [Sentry’s discord](https://discord.gg/Ww9hbqr) if you ever need anything.

Also apologizes if I came across like I was bikeshedding or causing friction on your PR - totally not the intention, just trying to share some learnings (and hopefully learn some stuff from you folks as well!)

silesky:
I just did a comparison and the static methods added 16 bytes to the size of the umd bundle. The rest was other code.

yordis:
Leaving some notes,

`static` isn't that concerning (at the moment), but the programming style could continue optimizing things with simple changes.
Instead of thinking about "Classes" first, it would be much better to lean on typespec, and simple JS Objects, and do it more functional.

For example, `PreInitMethodCallBuffer` could become an object, and every permutation and mutations and whatnot could be a function that takes a `PreInitMethodCallBuffer` argument.

Making things classes will continue increasing the bundle size. This could be avoided, especially in cases that they are internal to your lib, so you have full control over what is going on inside.

Anyway, probably we could (if you want to), continue the conversation somewhere else, focus on getting the features to work first, and then follow up.

Make it pretty 💜 

image

src/browser.ts
silesky: @AbhiPrasad no; you weren’t bikeshedding at all, this is great stuff to bring up!

src/browser.ts
silesky: I made it a bit more functional instead of using a foreach -- it's still O(n) -- let me know your thoughts.

src/analytics-preload.ts
```diff
@@ -0,0 +1,249 @@
...
+  once = this._createMethod('once')
+  off = this._createMethod('off')
+  on = this._createMethod('on')
+  addSourceMiddleware = this._createMethod('addSourceMiddleware')
+  addIntegrationMiddleware = this._createMethod('addIntegrationMiddleware')
+  setAnonymousId = this._createMethod('setAnonymousId')
+  addDestinationMiddleware = this._createMethod('addDestinationMiddleware')
+}
+
+const main = async () => {
```
silesky, line 240: Just an example

src/analytics-preload.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  constructor(myExecutor: Executor) {
+    const promise = myExecutor(this.preInitBuffer)
+    void this.run(promise)
+  }
+
+  // register onFulfilled and onRejected methods on class.
+  // while AnalyticsWithBuffer instance is only promise-like,
+  // .then should return a Promise<[Analytics...]> and not another AnalyticsWithBuffer
+  async then<T, T2 = never>(
```
silesky, line 173: WIP

src/analytics-preload.ts
```diff
@@ -0,0 +1,271 @@
...
+ *
+ */
+
+/**
+ * Inject a preInitBuffer into the getAnalytics function so load can call
+ */
+type Executor = (
+  preInitBuffer: PreInitMethodCallBuffer
+) => Promise<[Analytics, Context]>
+
```
silesky, line 146: WIP -- mulling over having this whole class have a method that returns an extended Promise (can't extend it directly, per Liskov Substitution Principle).
e.g.
```ts
const thenable = new AnalyticsWithBuffer((preInitBuffer) => AnalyticsBrowser.load(arg1, preInitBuffer))
  .createThenableInstance()
```

src/analytics-preload.ts
```diff
@@ -0,0 +1,248 @@
...
+    } catch (err) {
+      if (this.onRejected) {
+        this.onRejected(err)
+      } else {
+        throw err
+      }
+    }
+  }
+
+  private _createMethod<T extends PreInitMethodName>(methodName: T) {
```
chrisradek, line 210: Does this make every method an async method? I'm trying to think about if our synchronous methods are incapable of throwing errors today. If they start returning promises then the way users handle errors will have to change as well since a try/catch will only work if they await the function. That could be a breaking change.

chrisradek (commented):
I really like the direction this is moving in. I haven't gone through all the implementation yet (it's still in draft so I don't want to go too deep if the state is changing) and need some coffee before I dig into the `then` implementation :smile:

src/analytics-preload.ts
silesky: Every method is technically an async method because we really have no idea when the method is going to be called -- which is why we store the "reference" to resolve in the class.

This is an upgrade over the snippet API, where the type signature was basically a lie -- whether or not window.analytics.track returns a promise has to do with where it's called in our lifecycle -- which seems way more error prone than this.

src/browser.ts
```diff
@@ -88,31 +93,26 @@ function hasLegacyDestinations(settings: LegacySettings): boolean {
...
+  const callBufferedAnalyticsMethod = (methodCall: PreInitMethodCall) => {
+    const { method } = methodCall
+    // this guard is probably not needed.
+    if (typeof analytics[method] !== 'function') {
+      return console.warn(
+        `invariant error: method call "${method}" does not exist on analytics instance: ${analytics}`
+      )
+    }
+    if (method === 'addSourceMiddleware') {
+      void buffer.callMethod(analytics, methodCall)
```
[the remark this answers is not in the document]
silesky: "void" is something that eslint wants for expressions that return a promise that we will explicitly _not be handling_. Otherwise, you get `no-floating-promises`.

silesky:
> I really like the direction this is moving in. I haven't gone through all the implementation yet (it's still in draft so I don't want to go too deep if the state is changing) and need some coffee before I dig into the `then` implementation 😄

Thank you!

src/analytics-preload.ts
```diff
@@ -0,0 +1,248 @@
...
+    methodCall: PreInitMethodCall<T>
+  ) {
+    if (methodCall.called) {
+      return undefined
+    }
+
+    methodCall.called = true
+
+    const { method, args, resolve } = methodCall
+    const result = await (analytics[method] as Function)(...args)
```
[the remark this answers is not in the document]
silesky: > It's possible for a method to throw an error, so we might want to capture the promise's reject handler too and pass any thrown errors to that.

Good callout!

src/browser.ts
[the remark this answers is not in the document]
chrisradek: Hmm, that's interesting, but it looks like it also prevents us from handling any errors that might be thrown by the promise.

I just did a quick test [here](https://jsfiddle.net/okm2fh5y/2/) and when I look at the dev tools console after running I see the uncaught error (in promise) thrown. In node.js this would cause the process to exit. If we truly want to not worry about errors from some promises we may want to explicitly ignore them or log them.

src/analytics-preload.ts
chrisradek: Totally agree in the value of the APIs that currently return a promise returning a promise even when they're buffered.

It looks like of the methods that don't return a promise are:
- reset
  - returns void already
- debug
  - returns `Analytics` so could return `AnalyticsBuffered`
- on/off/once
  - returns `Analytics` so could return `AnalyticsBuffered`. This one feels the most weird to make promise-based since event emitters are so common in JS and it's expected you can just chain `.on()` calls.  
- setAnonymousId
  - returns `ID`. This is the trickiest one. _Technically_ ID can return undefined - we could document that it returns undefined if the instance hasn't loaded yet which is today's behavior on the CDN. I'm not sure what to do for this method - if we returned void from the buffered call like we do today with the snippet the API is still unchanged - maybe it is just a matter of documenting the behavior.

src/browser.ts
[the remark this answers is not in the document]
silesky: > Hmm, that's interesting, but it looks like it also prevents us from handling any errors that might be thrown by the promise.

Oops, good callout. I've tried to preserve the existing logic and I made a mistake looking at the old code; this should be awaited!

On that note, the existing logic allows for unhandled promise rejections:

https://github.com/segmentio/analytics-next/blob/aa423cc828fc9da0a4df6603dbda9ab3978db698/src/browser.ts#L109-L112

^ should we handle these?

> I just did a quick test [here](https://jsfiddle.net/okm2fh5y/2/) and when I look at the dev tools console after running I see the uncaught error (in promise) thrown. In node.js this would cause the process to exit.

Yep, 'void' is like a comment!
```ts
void Promise.reject(123) // VM364:1 Uncaught (in promise) 123 
```
That's a good point about nodejs --  I guess the fact that this is browser code means that we haven't had any complaints.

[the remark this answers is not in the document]
chrisradek: > ^ should we handle these?

Oh nice catch! I think so yes. For what it's worth, I think we do make it pretty hard for most of our APIs to actually throw an error to the end-user, though I wouldn't bet my node process on that!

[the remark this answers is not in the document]
silesky: I added error handling to the `callMethod` function, and have some tests around it.

src/browser.ts
```diff
@@ -235,56 +216,74 @@ async function registerPlugins(
...
-    flushPreBuffer(analytics)
-
-    const ctx = await registerPlugins(
-      legacySettings,
-      analytics,
-      opts,
-      options,
-      plugins
-    )
+async function loadAnalytics(
```
silesky, line 219: Moved this outside the class to make it "private private". When we enable ES2015 in tsconfig, we can use _real_ private fields.

src/analytics-pre-init.ts
```diff
@@ -0,0 +1,276 @@
...
+  calls: PreInitMethodCall[]
+): Promise<void> {
+  for (const c of calls) {
+    if (c.method === name) {
+      await callAnalyticsMethod(analytics, c).catch(console.error)
+    }
+  }
+}
+
+export const flushAddSourceMiddleware = flushAnalyticsCallsSeriallyByName.bind(
```
silesky, line 75: The previous logic had multiple methods of "addSourceMiddleware" being called serially, rather than concurrently -- so I created a function to handle this case. This potentially slows down full initialization (however that matters); However, there were no tests asserting that -- so I am still puzzling together the previous intention. It's pretty common for devs to be confused in javascript about the async behavior in loops vs higher-order-functions vs Promise.whatever, so I never rule that out. 

So my outstanding questions is:
-  can addSourceMiddleware be safely refactored to be called concurrently using promise.allSettled, like the other methods?

silesky: That's a good point about the order of middleware.

The "concurrent" function that uses `allSettled` would still preserve the order that the middleware is registered, since the Array.map/filter calls that creates the promises will operate on the middleware functions in order.

The only possibilities I could think of:
 - it's a mistake -- serial calls are not necessary (for example, promise.allSettled would work)
 - there's a race condition around registering middleware too quickly, but it's not caught in the tests.

src/analytics-pre-init.ts
```diff
@@ -0,0 +1,279 @@
...
+        this.instance = ajs
+        this.ctx = ctx
+      })
+      .catch(() => {
+        // intentionally do nothing...
+        // this result of this promise will be caught by the 'catch' block on this class.
+      })
+  }
+
+  private _createMethod<T extends PreInitMethodName>(methodName: T) {
```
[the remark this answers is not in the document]
silesky: Thanks, I've been thinking this over
> on/off/once 
this is a good call out -- I think this is in the spirit of the event emitter interface. Since we're registering a handler, and a user knowing _when_ the handler is registered is of limited use -- it's worth the trade off.

> setAnonymousId 
I think we just document this behavior, we don't actually have any way of returning anything but a promise<ID | undefined> 

> reset
as is -- Promise<void> is fine.

> debug 
as is -- Promise<void>

src/analytics-pre-init.ts
arielsilvestri: We can add some emit functionality to curtail this issue where we emit a "done" message once all the middlewares are finished registering, and then wait on that for the rest of the tasks.

src/analytics-pre-init.ts
```diff
@@ -0,0 +1,307 @@
...
+  this,
+  'addSourceMiddleware'
+)
+
+export const flushOn = flushAnalyticsCallsConcurrentlyByName.bind(this, 'on')
+
+export const flushSetAnonymousID = flushAnalyticsCallsConcurrentlyByName.bind(
+  this,
+  'setAnonymousId'
+)
```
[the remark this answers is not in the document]
silesky: Yep, this doesn't change the previous behavior in terms of how these are flushed.

The previous implementation flushed the `on` methods in basically the same way, the fact of its concurrency was just obscured by the implementation -- since it was in a forEach which doesn't "await" between iterations. You can refactor any forEach(async () =>) to a `promise.allSettled` without changing the iteration behavior, and you get the added benefit of the option of awaiting the entire operation when you're done (we're not doing this all the time, but it's an option). I'm just refactoring to be explicit here, since the previous implementation had a wide range of async behaviors / treatments.

setAnonymousID is a synchronous function if analytics is instantiated, but if a user calls it during pre-init, it returns a promise that will resolve when it gets flushed -- so it needs to be called in an async way.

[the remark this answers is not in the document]
pooyaj: So about `on`, is this the method we are calling: https://github.com/segmentio/analytics-next/blob/master/src/core/emitter/index.ts#L4-L7 ? This is what confuses me, as it seems to be just adding the callback to the array synchronously 🤔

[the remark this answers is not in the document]
pooyaj: Yeah! too many methods with different signatures 😓  Good opportunity to trim the runConcurrently utility, and probably remove the import of AllSettled to save on bundle size

src/__tests__/analytics-pre-init.integration.test.ts
```diff
@@ -0,0 +1,341 @@
...
+      expect(
+        [
+          analytics.on('track', jest.fn),
+          analytics.off('track', jest.fn),
+          analytics.once('track', jest.fn),
+        ].map((el) => el instanceof AnalyticsBuffered)
+      ).toEqual([true, true, true])
+    })
+
+    test('"emitted" events should be chainable', async () => {
```
chrisradek, line 297: Can we have a test that also ensures that `this` passed to the event listener callbacks is the `analytics` instance? It looks like it is from my testing, I just want to make sure we don't accidentally break that in the future.

chrisradek (commented):
Looking good! Nothing serious, just a few suggestions on test cases and some size limit savings.

src/__tests__/analytics-pre-init.integration.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    })
+
+    describe('.catch', () => {
+      it('should be capable of handling errors if using promise syntax', () => {
+        browserLoadSpy.mockImplementationOnce((): any => Promise.reject(errMsg))
+
+        const ajsBuffered = AnalyticsBrowser.load({ writeKey: 'abc' })
+        const newPromise = ajsBuffered.catch((reason) => {
+          expect(reason).toBe(errMsg)
+        })
```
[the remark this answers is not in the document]
silesky: Yep. For example, that case is also covered  here: https://github.com/segmentio/analytics-next/pull/436/files/f1042bc17a4f13b7335433390d1892e77b630aed#diff-a3756d7d2cc9639c2f00f5871656d78a5aa10fddd6ba6bdae04447a577140d4aR165 (and probably some other places)

I think I preferred to test promise syntax here in order to be extremely explicit that we are talking about async functions. You can see that there are places that test async/await.

_Technically_, if something is asserted to be a thenable (an object with a .then and optional .catch method), I don't think it's technically neccessary to test both syntax forms, as that's part of the spec.

src/analytics-pre-init.ts
```diff
@@ -0,0 +1,320 @@
...
+  public getCalls<T extends PreInitMethodName>(
+    methodName: T
+  ): PreInitMethodCall<T>[] {
+    return (this._value[methodName] ?? []) as PreInitMethodCall<T>[]
+  }
+
+  push(...calls: PreInitMethodCall[]): void {
+    calls.forEach((el) => {
+      if (this._value[el.method]) {
+        this._value[el.method]?.push(el)
```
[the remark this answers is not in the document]
silesky: There is a refactoring but I think I rolled it back because IIRC it was involved -- It comes down to the MethodMap type being a bit loose. =)

src/analytics-pre-init.ts
```diff
@@ -0,0 +1,320 @@
...
+  once = this._createChainableMethod('once')
+  off = this._createChainableMethod('off')
+  on = this._createChainableMethod('on')
+  addSourceMiddleware = this._createMethod('addSourceMiddleware')
+  addIntegrationMiddleware = this._createMethod('addIntegrationMiddleware')
+  setAnonymousId = this._createMethod('setAnonymousId')
+  addDestinationMiddleware = this._createMethod('addDestinationMiddleware')
+
+  private _createMethod<T extends PreInitMethodName>(methodName: T) {
+    return async (
```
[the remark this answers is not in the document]
silesky: Did not know that there was a cost savings! 

If so, we should enable https://eslint.org/docs/rules/require-await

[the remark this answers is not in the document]
silesky: image

chrisradek (approved):
Whoo! Looks good to me!

pooyaj (approved):
🚀🚀🚀

PR #439, opened 2022-04-28
add bundle analyzer to example 

silesky:
next start currently doesn't work because our version of typescript is too old for next, so I didn't add any commands until typescript is upgraded (I was planing to do this when I had some time). You can get some insight by analyzing the dev bundle, but it's not really _for_ that.


chrisradek (approved):
Sweet!

PR #441, opened 2022-04-30
Update typescript, node, eslint and other dependencies, fix lint errors

silesky:
We were almost two years behind on dev dependencies, like Typescript. 

When I upgraded typescript, as usual, I had to do a cascade of small fixes. 😊

Among other things, This PR:
- upgrades typescript to 4.3
- upgrades JSDOM to 19
- makes some small code modifications to fix type
- runs prettier as a lint rule, which means that unprettiered code will be caught by CI/CD
- formats all "unformatted code"
- removes electron which wasn't being used

Blocker -- buildkite version of node is too old -- we should upgrade to node 14.
image


pooyaj:
Hopefully this fixes it: https://github.com/segmentio/analytics-next/pull/442 feel free to merge it into your branch anytime

.prettierrc.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
```
chrisradek, line 1: Are these what we were already using or is this something completely new?

silesky: It's what we were using (but it includes json formatting abilities). Changes to this will be picked up in eslint.

src/core/context/index.ts
```diff
@@ -72,7 +72,7 @@ export class Context implements AbstractContext {
     throw new ContextCancelation({ reason: 'Context Cancel' })
   }
 
-  log(level: LogLevel, message: string, extras?: object): void {
+  log(level: LogLevel, message: string, extras?: any): void {
```
chrisradek, line 75: Do we require an object or can it actually be anything? If it needs to be an object maybe something like `Record<string, any>` would be more accurate and appease the linters.

src/analytics.ts
```diff
@@ -139,13 +139,8 @@ export class Analytics extends Emitter {
...
 
   async page(...args: PageParams): Promise<DispatchedEvent> {
-    const [
-      category,
-      page,
-      properties,
-      options,
-      callback,
-    ] = resolvePageArguments(...args)
+    const [category, page, properties, options, callback] =
```
chrisradek, line 142: What rule changed this behavior? Purely subjective tastes - I find the former easier to read compared to an `=` followed by newline but that might just be me.

package.json
```diff
@@ -31,7 +31,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "25.0 KB"
```
pooyaj, line 34: This is like inflation, always goes up, never goes down 😆

src/lib/parse-cdn.ts
```diff
@@ -1,6 +1,7 @@
 import { embeddedWriteKey } from './embedded-write-key'
 
-const analyticsScriptRegex = /(https:\/\/.*)\/analytics\.js\/v1\/(?:.*?)\/(?:platform|analytics.*)?/
+const analyticsScriptRegex =
```
pooyaj, line 3: expected change 🤔 ?

pooyaj (approved):

package.json
silesky: by 3 bytes!

src/core/context/index.ts
silesky: I agree that it's not ideal, but I assume that extras could also be a string? I see it's being used logged artifacts from error handling, which were very loosely-typed, so I didn't want to refactor any more code than I absolutely needed -- I'm kind of afraid to make any assumptions in the catch clauses.

chrisradek: As long as we don't require it to be in the shape of an object and will allow anything, then `any` seems fine. Right now we would have had TypeScript enforce it as at least an object, but ok with loosening if that better reflects the actual state of things.

src/analytics.ts
silesky: I agree, but this was something `new` in prettier, which I assume was basically choosing to not wrap list items like this unless it hit the print width.

silesky: I just checked and `prettier-eslint-config` (correctly) disables it, as it conflicts with prettier's opinions. I think we can enable something like that by adding a rule for  "array-element-newline".

src/lib/parse-cdn.ts
silesky: yep, print-width

src/core/context/index.ts
silesky: Thanks for bringing this up. I changed this back, I think  it's safe to tighten this back up to at least object and then do unsafe coercions in the catch clause (until someone wants to refactor later).

chrisradek (approved):

PR #442, opened 2022-04-30
upgrade bk node version

pooyaj:
Update BK Node


silesky (approved):
Thanks!! Just in time!

PR #443, opened 2022-05-03
Fixes race condition where querystring params could be ignored

chrisradek:
This PR moves updated `AnalyticsBrowser.load()` to:
- Move the querystring API call prior to considering analytics.js as `initialized`
- Waits for `querystring()` to complete before considering analytics.js as `initialized`

Previously there was a race condition that would cause the `anonymousId` passed in via a querystring to be overridden if there were any calls made to `analytics` before it finished loading.

An easy way to demonstrate this is to make calls to analytics before loading it and (optionally) simulate network latency since the querystring API is fetched dynamically when loading analytics.js. This would cause analytics.js to generate a random anonymousId for itself for the events to use. In some cases, if the querystring API completed before the events hit the `Segment.io` plugin, analytics.js would have its anonymousId reset to match the event's anonymousId and override what was set by the querystring API.

### DIFF
--- a/src/__tests__/query-string.integration.test.ts
+++ b/src/__tests__/query-string.integration.test.ts
@@ -35,6 +35,35 @@ describe('queryString', () => {
     )
   })
 
+  it('applies query string logic before analytics is finished initializing', async () => {
+    let analyticsInitializedBeforeQs: boolean | undefined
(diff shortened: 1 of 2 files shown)


pooyaj (approved):
good catch 🙌

Issue #393, opened 2022-03-11
Make protocol configurable


livthomas:
We use our backend as a proxy to Segment API in order to bypass ad-blockers. But we are unable to test it in local dev environment on our machines where HTTP is used instead of HTTPS.

It looks like HTTPS protocol is hardcoded [here](https://github.com/segmentio/analytics-next/blob/7c63868477ede9272ac9c072931110fe37f0f721/src/plugins/segmentio/index.ts#L58). Can you create a new settings property which specifies protocol and defaults to `https`?

I've already asked for this in https://github.com/segmentio/analytics.js-integrations/issues/559 and there hasn't been any response.

arielsilvestri:
Hey @livthomas, you should now be able to do the following in your load configuration to achieve this:

```javascript
analytics.load('writeKey', {
  integrations: {
    "Segment.io": {
     // Other options...
      protocol: "http"
    }
  }
);
```

PR #450, opened 2022-05-04
Adds support for configurable protocol

arielsilvestri:
Per #393, provides a configurable protocol option for calls to Segment api.

Tested both with a new unit test and using local example.

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -67,7 +67,6 @@ export interface InitOptions {
   plan?: Plan
   retryQueue?: boolean
   obfuscate?: boolean
-  protocol?: 'http' | 'https'
 }
 
 export class Analytics extends Emitter {
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -67,6 +67,7 @@ export interface InitOptions {
   plan?: Plan
   retryQueue?: boolean
   obfuscate?: boolean
+  protocol?: 'http' | 'https'
 }
 
 export class Analytics extends Emitter {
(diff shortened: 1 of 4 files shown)


src/analytics.ts
```diff
@@ -67,6 +67,7 @@ export interface InitOptions {
   plan?: Plan
   retryQueue?: boolean
   obfuscate?: boolean
+  protocol?: 'http' | 'https'
```
chrisradek, line 70: Since this setting is specific to the segment.io plugin, I think it makes more sense to add this to https://github.com/segmentio/analytics-next/blob/aa423cc828fc9da0a4df6603dbda9ab3978db698/src/plugins/segmentio/index.ts#L15-L33

That way there's also no confusion on why we're still loading bundles and talking to destinations via https instead of http :smile:

arielsilvestri: Hey @chrisradek just curious where they'd need to set that value if we move it to the deliveryStrategy property? I'm guessing somewhere in the plugin configuration for segment.io? I want to make sure I convey how their devs can use the property properly.

chrisradek: I was imagining it would live next to `apiHost`, so if they wanted to talk to their dev environment they'd be able to load analytics like:
```js
analytics.load('writeKey', {
  integrations: {
    "Segment.io": {
      apiHost: "localhost",
      protocol: "http"
    }
  }
);
```

arielsilvestri: Oh I understand now. Yes, I agree with that - sorry, the GitHub preview of your snippet centered on that `deliveryStrategy`, so i thought you meant to put it there.

chrisradek: Woops, I should have called out the interface! 😬

chrisradek (approved):
Looks good to me!

pooyaj (approved):

PR #452, opened 2022-05-06
Add react examples to ReadMe

catligad:
## Description
We're using the [React Example](https://github.com/segmentio/react-example) repo instead of the current example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,15 @@ The easiest and quickest way to get started with Analytics 2.0 is to [use it thr
 
 ## 💡 Using with Segment
 
+**There is a [React example repo](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-quick-start) which outlines using the Segment snippet.**
+
 1. Create a javascript source at [Segment](https://app.segment.com) - new sources will automatically be using Analytics 2.0! Segment will automatically generate a snippet that you can add to your website. For more information visit our [documentation](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/)).
 
 2. Start tracking!
 
 ## 💻 Using as an NPM package
+
+**There is a [React example repo](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-package) which outlines using the Segment npm package.**
 
 1. Install the package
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,12 +20,16 @@ The easiest and quickest way to get started with Analytics 2.0 is to [use it thr
 
 ## 💡 Using with Segment
 
+**There is a [React example repo](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-quick-start) which outlines using the Segment snippet.**
```
catligad, line 23: The links out to these examples are well documented; developers should be able to read it along with going through the repo and see how to implement the examples.

README.md
```diff
@@ -20,12 +20,16 @@ The easiest and quickest way to get started with Analytics 2.0 is to [use it thr
...
+
 1. Create a javascript source at [Segment](https://app.segment.com) - new sources will automatically be using Analytics 2.0! Segment will automatically generate a snippet that you can add to your website. For more information visit our [documentation](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/)).
 
 2. Start tracking!
 
 ## 💻 Using as an NPM package
 
+**There is a [React example repo](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-package) which outlines using the Segment npm package.**
```
chrisradek, line 31: I wonder if it makes more sense to create a `using React` subsection like we have for Vue and place this comment there.

I think we should probably still have a snippet that shows calling `AnalyticsBrowser.load()` under item 2 to get an instance of analytics since that's the bare minimum you'd need today to get started.

README.md
```diff
@@ -39,51 +43,8 @@ yarn add @segment/analytics-next
 pnpm add @segment/analytics-next
 ```
 
-2. Import the package into your project and you're good to go (with working types)! Example react app:
+2. Import the package into your project and you're good to go (with working types)!
```
chrisradek, line 46: I think the intent of the sample here was to show what the API looks like to load analytics when using the package directly, we just happened to use react as the example.

If we link to the react-example section, maybe this snippet can be reduced to something like:

```ts
import { Analytics, AnalyticsBrowser, Context } from '@segment/analytics-next'

async function loadAnalytics(): Promise<Analytics> {
  const [ analytics, context ] = await AnalyticsBrowser.load({ writeKey })
  return analytics
}
```

chrisradek (commented):
@catligad Thank you so much for working on this react example! Happy to link to it from this README! Just had a comment around the using segment npm package change.

chrisradek (approved):
Looks great, thank you so much @catligad for putting this together!

PR #459, opened 2022-05-16
[feature] Updates AnalyticsNode API

chrisradek:
This change is part of the effort to improve using analytics in node.js. The goal is that this library should be able to replace the `analytics-node` library.

This PR focuses on 2 aspects:
1. Defines a node.js-specific analytics interface that represents the APIs available in node.js
2. Updates the `Analytics` class so that it can function without maintaining user/group state.

## Node.js interface

The `AnalyticsNodeJs` interface exposes the analytics API that works in node.js. If we continued exporting `Analytics` from `AnalyticsNode.load()` directly then we'd have 2 issues:

1. Many APIs are either only supported in the browser (e.g. `trackLink`) or were included for backwards compatibility with analytics.js classic and have been deprecated (e.g. `use`).
2. Analytics assumes that 1 user/group is being used at a time.

This change addresses both these points by only exposing the APIs that work in node.js and enforcing that either a `userId` or `anonymousId` is provided in each operation that generates an event.

## Analytics sans users

The `Analytics` class assumed that there would always be a single user (anonymous and/or known) at a time and would automatically keep track of the user details so our customers wouldn't have to. This causes an issue when using `Analytics` to a server environment because it's common to be handling multiple users concurrently.

The simplest way I found to handle this was to support disabling the `User` and `Group` objects so that they didn't store any state. This didn't require changing any method signatures. I thought of some alternatives such as:
- creating a `UserLike`/`GroupLike` interface to pass to Analytics instead of classes so we can pass in a no-op implementation in node.js
- Subclassing `User`/`Group` with `NullUser` and `NullGroup`
The 1st ended up leading to broader API changes that would leak into the browser usage unless we did our own type casting.
The 2nd option is safe but caused some extra work to be done in the constructor.

## Notes
I've set this PR to target the feature-node-uplift branch. Since we are changing the API for the node.js `Analytics`, this is technically a breaking change and we should do a major version bump when we decide to ship this. I don't want to block releasing from the main branch while this work is ongoing.

### DIFF
--- a/src/node.ts
+++ b/src/node.ts
@@ -8,6 +8,7 @@ import { PriorityQueue } from './lib/priority-queue'
 import { Options } from './core/events/interfaces'
 import { Callback, DispatchedEvent } from './core/arguments-resolver'
 import { Emitter } from './core/emitter'
+import { UserOptions } from './core/user'
 
 export type Identity =
   | { userId: string; anonymousId?: string }
@@ -33,17 +34,6 @@ export interface AnalyticsNodeJs extends Emitter {
   /**
    * Associates an identified user with a collective.
    * @param groupId - The group id to associate with the provided user.
-   * @param options - A dictionary of options including the user id.
-   * @param callback
-   */
-  group(
-    groupId: string,
-    options: NodeOptions,
-    callback?: Callback
-  ): Promise<DispatchedEvent>
-  /**
-   * Associates an identified user with a collective.
-   * @param groupId - The group id to associate with the provided user.
    * @param traits - A dictionary of traits for the group.
    * @param options - A dictionary of options including the user id.
    * @param callback
@@ -56,15 +46,16 @@ export interface AnalyticsNodeJs extends Emitter {
   ): Promise<DispatchedEvent>
 
   /**
-   * Record traits about a user.
    * Includes a unique userId and/or anonymousId and any optional traits you know about them.
+   * @param userId
    * @param traits
    * @param options
    * @param callback
    */
   identify(
-    traits: object,
-    options: NodeOptions,
+    userId: string,
+    traits?: object,
+    options?: Options,
     callback?: Callback
   ): Promise<DispatchedEvent>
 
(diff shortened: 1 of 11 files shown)


chrisradek:
Note: I believe some tests are sometimes timing out because validation errors are getting retried. I believe validation errors should fail without retries since the data being validated doesn't change between attempts. I'll update this PR with those changes.

src/node.ts
```diff
@@ -5,32 +5,180 @@ import { analyticsNode } from './plugins/analytics-node'
...
+    disable: true,
+  }
+
+  const queue = new EventQueue(new PriorityQueue(3, []))
+  const options = { user: userOptions, group: userOptions }
+  const analytics = new Analytics(settings, options, queue)
+
+  const nodeSettings = {
+    writeKey: settings.writeKey,
+    name: 'analytics-node-next',
```
pooyaj, line 175: might want to keep the name the same to avoid app changes (I think we have some custom renderers in the app debugger that are tied to library names). we can inject the `next` keyword in the version instead to keep the library name the same.

pooyaj (approved):
Looks great 🙌 🙌

src/node.ts
```diff
@@ -5,32 +5,180 @@ import { analyticsNode } from './plugins/analytics-node'
...
+
+  /**
+   * Records page views on your website, along with optional extra information
+   * about the page viewed by the user.
+   * @param properties
+   * @param options
+   * @param callback
+   */
+  page(
+    properties: object,
```
silesky, line 70: The docstring says this is a dictionary. If so, wouldn’t  the correct type be Record<string | number, unknown> or something like that? object includes Array, null, etc IIRC.

src/node.ts
chrisradek: You mean keep it the same as what `AnalyticsBrowser` uses or what analytics-node uses? This is the name that's being used today.

src/node.ts
chrisradek: That sounds reasonable! I copied the type from what `Analytics` is using but might as well tighten the types up here!

silesky: ah K, figured. We might want to take this opportunity to create an official "Properties" or "PropertiesDictionary" type, since we use it all over the app

src/node.ts
danieljackins: Similarly do we maybe want to include node in the messageID? analytics-node uses `node-`, maybe something like `node-next-`? 😄

PR #464, opened 2022-05-19
async methods should have await

silesky:
As @chrisradek  pointed out, async methods have a few more bytes than non-async methods.
This adds the `require-await` rule to our codebase so we're consistent in our approach. 🎸

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,9 +21,7 @@
   ],
   "overrides": [
     {
-      "files": [
-        "*.{ts,tsx}"
-      ],
+      "files": ["*.{ts,tsx}"],
       "parser": "@typescript-eslint/parser",
       "parserOptions": {
         "project": "./tsconfig.json"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
-      "limit": "25.9 KB"
+      "limit": "25.7 KB"
```
silesky, line 35: Sometimes the size does go down.

src/analytics.ts
```diff
@@ -377,7 +378,7 @@ export class Analytics extends Emitter {
     legacyDestinations.forEach((destination) => {
       destination.addMiddleware(...middlewares)
     })
-    return this
+    return Promise.resolve(this)
```
silesky, line 381: I added Promise.resolve to avoid a breaking change.

src/plugins/ajs-destination/index.ts
```diff
@@ -305,11 +305,11 @@ export class LegacyDestination implements Plugin {
   }
 }
 
-export async function ajsDestinations(
+export function ajsDestinations(
```
silesky, line 308: unless we have a policy that all these plugins need to be async, this looked safe.

.eslintrc
```diff
@@ -51,10 +49,20 @@
...
-        "@typescript-eslint/require-await": "off",
+        "require-await": "off",
+        "@typescript-eslint/require-await": "error",
         "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/ban-types": "off" // TODO: turn on
-      }
+      },
+      "overrides": [
+        {
+          "files": ["*.test.*"],
```
silesky, line 59: There were a _lot_ of test cases that violated this, and I don't necessarily think it's worth enforcing this for test cases. 😄

package.json
pooyaj: 😱😱😱

pooyaj (approved):

PR #465, opened 2022-05-19
[bugfix] integration overrides now work with action destinations

chrisradek:
This PR updates how we resolve `integrations` passed to `analytics.load()` so that they now apply towards action destinations. Prior to this change they only applied to classic destinations.

The gist is now we merge the remote plugin settings from the CDN with the integration settings/overrides the user provides, whereas previously we were using the settings exclusively.

### DIFF
--- a/src/plugins/remote-loader/__tests__/index.test.ts
+++ b/src/plugins/remote-loader/__tests__/index.test.ts
@@ -1,5 +1,7 @@
 import * as loader from '../../../lib/load-script'
 import { remoteLoader } from '..'
+import { AnalyticsBrowser } from '../../../browser'
+import { InitOptions } from '../../../analytics'
 
 const pluginFactory = jest.fn()
 
@@ -30,6 +32,7 @@ describe('Remote Loader', () => {
           },
         ],
       },
+      {},
       {}
     )
 
@@ -49,6 +52,7 @@ describe('Remote Loader', () => {
           },
         ],
       },
+      {},
       {},
       true
     )
@@ -73,6 +77,7 @@ describe('Remote Loader', () => {
           },
         ],
       },
+      {},
       {}
     )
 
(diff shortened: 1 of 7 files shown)


src/plugins/remote-loader/__tests__/index.test.ts
```diff
@@ -339,11 +351,105 @@ describe('Remote Loader', () => {
...
+
+    const initOptions: InitOptions = {
+      integrations: {
+        remotePlugin: {
+          name: 'Chris Radek',
+        },
+      },
+    }
+
+    await AnalyticsBrowser.load(
```
chrisradek, line 438: I wanted to test this at the AnalyticsBrowser.load() level as well to make sure the integration options merging is working correctly.

src/plugins/remote-loader/index.ts
```diff
@@ -14,7 +14,7 @@ export interface RemotePlugin {
   /** The UMD/global name the plugin uses. Plugins are expected to exist here with the `PluginFactory` method signature */
   libraryName: string
   /** The settings related to this plugin. */
-  settings: JSONValue
+  settings: JSONObject
```
chrisradek, line 17: I checked how we build the settings that get served by the CDN and it's always an object.

src/plugins/segmentio/index.ts
```diff
@@ -12,7 +12,7 @@ import standard from './fetch-dispatcher'
 import { normalize } from './normalize'
 import { scheduleFlush } from './schedule-flush'
 
-export interface SegmentioSettings {
+export type SegmentioSettings = {
```
silesky, line 15: Why the change?

chrisradek: TypeScript was complaining that `SegmentioSettings` wasn't a subset of the `JSONObject` shape (but it is since it's just an interface containing properties) - changing this from an interface to a type allowed TypeScript to recognize that.

pooyaj (approved):
:shipit:

PR #469, opened 2022-05-21
update docs for new api

silesky:
waiting for https://github.com/segmentio/analytics-next/pull/470


example/tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "moduleResolution": "node",
```
silesky, line 12: This was auto-updated by next

pooyaj (approved):
💥 I think we should ask the enablement to update their react example app as well

PR #470, opened 2022-05-23
Pre-init API tweak for Typescript users

silesky:
A nice API tweak for Typescript users that occurred to me over the weekend that slightly reduces the public API surface area. We don't need to expose the AnalyticsBuffer class; it's can just be an implementation detail.

 (assuming a user who want to wrap our library)…

```ts
import { AnalyticsBrowser } from '@segment/analytics-next' // no need to know about  AnalyticsBuffered 

const loadAnalytics = (writeKey: string): AnalyticsBrowser => {
   return AnalyticsBrowser.load({ writeKey })
}

```

### DIFF
--- a/src/__tests__/analytics-pre-init.integration.test.ts
+++ b/src/__tests__/analytics-pre-init.integration.test.ts
@@ -2,7 +2,6 @@ import { AnalyticsBrowser } from '..'
 import unfetch from 'unfetch'
 import { mocked } from 'ts-jest/utils'
 import { Analytics } from '../analytics'
-import { AnalyticsBuffered } from '../core/buffer'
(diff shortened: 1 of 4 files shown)


chrisradek (approved):
So the benefit is that if you don't rely on type inference, you can reference the `AnalyticsBrowser` class instead of `AnalyticsBuffered` since you'd already have to be importing `AnalyticsBrowser` already?

silesky:
> So the benefit is that if you don't rely on type inference, you can reference the `AnalyticsBrowser` class instead of `AnalyticsBuffered` since you'd already have to be importing `AnalyticsBrowser` already?

Yep.

PR #471, opened 2022-05-25
update playground with idiomatic next instantiation

silesky:
This PR makes the example page reflect best practices (i.e. single load instantiation,  SSR-ready).

The canonical example for how our app should be used should be colocated with our code, so it's important that they are mantained, and that we can point to. 

(side note: yep, the playground is not a "one-size-fits-all",  but we can have a full directory for independent examples once we move to a more typical monorepo setup).

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,10 @@ const App = () => {
   )
 ```
 
-For another example,
-There is a [React example repo](https://github.com/segmentio/react-example/) which outlines using the [Segment snippet](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-quick-start) and using the [Segment npm package](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-package).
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -120,8 +120,10 @@ const App = () => {
   )
 ```
 
-For another example,
-There is a [React example repo](https://github.com/segmentio/react-example/) which outlines using the [Segment snippet](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-quick-start) and using the [Segment npm package](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-package).
+More React Examples:
+
+- Our [playground](example/pages/_app.tsx) (written in NextJS) -- this can be run with `make dev`.
```
pooyaj, line 125: 🙌

pooyaj (approved):

PR #472, opened 2022-05-27
Chore/libweb 1127 monorepo refactoring pt. 1

silesky:
In order to reduce risk and make reviewing easier, I'm refactoring in stages. The goal is to have _only_ the runtime-agnostic code live in core. 😸 

This creates a "node" and "browser" directory and moves the appropriate tests into each. This also moves analytics into "core". Eventually, we will be able to remove runtime specific logic from core.

New Structure (note, "packages" is not introduced yet). See [SDD](https://paper.dropbox.com/doc/SDD-Break-up-analytics-next-into-discrete-packages--BiRfAKKbwIYRfHR2lXRqXjMdAg-kJZSy7qZEFOaBgYCtUDIm) for Summary!

image


.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn test
```
silesky, line 4: Sneaking this in =) 

(I've found this is the least controversial of all the hooks as it runs infrequently and fairly quickly, and I've found it extremely helpful).

zikaari (approved):
Looking nice! Good start 👍

src/__tests__/standalone-analytics.test.ts
```diff
@@ -1,247 +0,0 @@
-import jsdom, { JSDOM } from 'jsdom'
```
chrisradek, line 1: I don't see these tests tests got moved to - are these still covered?

chrisradek (approved):
Sweet, :ship::it:

PR #474, opened 2022-06-01
upgrade yarn in prep for workspaces

silesky:
Our yarn is way out-of-date, and doesn't support things like _modern workspaces_, plugins, etc.


silesky:
> I would enable Zero-Install and also `nodeLinker: pnp` it has been an amazing experience

I've been looking at examples of monorepos that are specifically [open source libraries](https://github.com/facebook/jest) and they tend to use old fashioned node_modules linking, so I think it makes more sense to use that. I've used zero-install in the past and had some mixed experiences with it. It seems IME to make a bit more sense for traditional web apps where there's more dependency churn, and, as an open source lib, ballooning the total size of the library is a definite concern.

chrisradek (approved):
Looks straightforward - doesn't seem like there are any real changes to how someone would get started contributing.

Got to learn more about yarn options too!

PR #476, opened 2022-06-03
- add yarn plugins - fix example - fix husky - fix eslint - fix buildkite

silesky:
Thing this does:
- Moves "examples" to a package
- Sets up a base eslint and tsconfig that are extended
- Lots of other stuff =)

Things this doesn't do:
- Adds an additional package alias (e.g. segment/analytics-browser)
- Refactors buildkite in a way that's more friendly to monorepos


.buildkite/pipeline.yml
```diff
@@ -2,88 +2,87 @@ env:
...
+      - yarn workspace @segment/analytics-next exec make test-qa
     retry:
       automatic:
         - exit_status: '*'
           limit: 2
     plugins:
       - ssh://git@github.com/segmentio/cache-buildkite-plugin#v2.0.0:
           key: "v1.1-cache-dev-{{ checksum 'yarn.lock' }}"
-          paths: ['node_modules/']
+          paths: ['.yarn/cache/']
```
silesky, line 35: https://github.com/yarnpkg/berry/discussions/2621#discussioncomment-505872

README.md
```diff
@@ -178,8 +178,8 @@ First, clone the repo and then startup our local dev environment:
 
 ```sh
 $ git clone git@github.com:segmentio/analytics-next.git
-$ cd analytics-next
-$ make dev
+$ cd analytics-next && yarn
+$ yarn dev
```
chrisradek, line 182: I think the unit testing section needs to be updated to `yarn test` from `make test-unit` as well.

packages/browser/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+  "extends": "../../tsconfig.json",
```
chrisradek, line 2: Is this giving us any value extending the root tsconfig.json? It looks like we redefine everything in `@tsconfig/node14` except `skipLibCheck` and `forceConsistentCasingInFileNames`, and ultimately this build is for browsers right? (Or am I misunderstanding?)

chrisradek (commented):
Overall this looks good and seems to work locally! One thing I think we need is a README.md in the package/browser directory - I'm not sure what would be published as the README to NPM without one. Maybe for now we basically copy the existing README until we rewrite the root one?

packages/browser/tsconfig.json
silesky: I had that thought -- I think having a root "tsconfig" to extend from is super common and potentially quite useful, so I have this as a kind of placeholder.

silesky: Added a more sensible tsconfig.

README.md
silesky: Good fallout. README and  other build behavior should actually have tests, I can write some

silesky:
> Overall this looks good and seems to work locally! One thing I think we need is a README.md in the package/browser directory - I'm not sure what would be published as the README to NPM without one. Maybe for now we basically copy the existing README until we rewrite the root one?


Good callout. README and  other npm behavior should actually have tests, I can write some.

zikaari:
Wouldn’t it make sense to move `snippet_example` up into `examples`, as a sibling of `with-next-js`?

silesky:
> Wouldn’t it make sense to move `snippet_example` up into `examples`, as a sibling of `with-next-js`?

snippet_example is used as a test harness, and as such, it's not a very clean example.

chrisradek (approved):
Looks good to me! Let the merge conflicts with existing PRs begin! 😁

PR #477, opened 2022-06-06
remove tests from tsconfig and extend typechecking

silesky:
Tests were being included in our npm package.

1. no more tests included in pkg
image

2. Extends tsconfig to include e2e tests, so type checking actually works
image


packages/browser/qa/__tests__/destinations.test.ts
```diff
@@ -3,15 +3,22 @@ import { reportMetrics } from '../lib/benchmark'
 import { browser } from '../lib/browser'
 import { run } from '../lib/runner'
 import { server } from '../lib/server'
-
 jest.setTimeout(100000)
 
+if (!process.env.QA_SAMPLES) {
+  throw new Error('no process.env.QA_SAMPLES')
+}
```
silesky, line 10: Had to fix all the typescript errors now that the normal (base editor) typescript compiler was actually including them.

packages/browser/qa/__tests__/destinations.test.ts
```diff
@@ -61,7 +68,7 @@ describe('Destination Tests', () => {
           host: url.host + url.pathname,
           queryParams: Array.from(url.searchParams.entries())
             .map((entry) => entry[0])
-            .filter((p) => !p.match(/^\d+$/))
```
chrisradek, line 64: Why this change? It looks like it was filtering out any query string parameters where the key was a number which...I don't actually know why we do that (I don't think a.js does anything with query string parameters that are just numbers though...), but I don't think we want the opposite of that. It also doesn't match what we do for analytics-next requests so those won't be matching now.

chrisradek (commented):
Mostly looks great, I think there was just 1 extra `!` removed! :smile:

packages/browser/qa/__tests__/destinations.test.ts
silesky: Oops, this was an accident from when I was fixing the types (and testing them). Hmm, weird that it didn't trigger a test failure =)

PR #478, opened 2022-06-08
[chore] updates jest to 28.x

chrisradek:
This PR updates jest to the latest major version (28.x at this time).

The main reason I wanted to update jest is because they've updated their fake timers to also mock system time which made testing some timestamp related fixes easier.

There are a number of breaking changes between jest 25 and 28, so this PR only deals with those breaking changes to get our tests in a passing state.

## Changes that needed to be handled

- The jest `testEnvironment` now defaults to `node`, so had to explicitly set it back to `jsdom`.
- jest removed `setImmediate` (fair since it's not a browser standard) but jsdom used setImmediate internally for it's preflight XHR implementation. Tests pass without providing setImmediate but causes ReferenceErrors that get logged in our tests that call device mode integrations.
- `mocked` API moved from `ts-jest/utils` to `jest`
- jest now enforces that you can't mix-and-match passing in a function that returns a promise and uses `done` for a single test
- window.close() causes issues when used more than once in a file due to a change in how jest does its teardown, we didn't actually need to call it for our tests - refactored state cleanup instead.
- jasmine API no longer exposed by default. We were using a mix of jest and jasmine spies - now we only use jest spies.

Edit: Few more issues encountered with e2e tests.
- IE smoke tests encountered parse errors. Turns out these were all skipped already, and given IE is retiring in a week I just removed the test.
- lighthouse test was timing out - it looks like with this update, `jest.setTimeout()` doesn't affect the test it's running in - it needs to be called outside of the given test. So I passed the timeout as an argument to the individual tests instead.
- forceExit - the e2e tests hang indefinitely with this update. Turns out there are leaking handlers even in our current state but jest is terminating the process after 1 second post test completion. I briefly investigated and found 1 file descriptor open but didn't pinpoint where - setting --forceExit just for e2e tests to keep the same behavior we have currently until we can investigate.


chrisradek:
The build failure looks like it's possibly not finding the cached node_modules - I think this was changed recently so maybe once @silesky's monorepo changes go in I can rebase and see if that resolves it.

src/core/queue/__tests__/event-queue.test.ts
```diff
@@ -98,7 +98,7 @@ test('does not enqueue multiple flushes at once', async () => {
   eq.dispatch(fruitBasket)
   eq.dispatch(anothaOne)
 
-  expect(setTimeout).toHaveBeenCalledTimes(1)
+  expect(jest.getTimerCount()).toBe(1)
```
silesky, line 101: nice

src/browser/__tests__/standalone-analytics.test.ts
```diff
@@ -204,22 +212,23 @@ describe('standalone bundle', () => {
...
-        // should run before any plugin is registered
-        'setAnonymousId',
-        // should run before any events are sent downstream
-        'register',
-        // should run after all plugins have been registered
-        'track',
-      ])
-      done()
-    }, 0)
+    await sleep(0)
```
silesky, line 215: 👍

src/plugins/segmentio/__tests__/normalize.test.ts
```diff
@@ -316,7 +316,7 @@ describe('before loading', () => {
       expect(object.context.referrer.id).toEqual('medium')
       assert(object.context.referrer.type === 'millennial-media')
       expect(cookie.get('s:context.referrer')).toBeUndefined()
-      expect(setCookieSpy.calls.count()).toBe(0)
+      expect(setCookieSpy.mock.calls.length).toBe(0)
```
silesky, line 319: expect(setCookieSpy).not.toBeCalled() is also an option

src/plugins/segmentio/__tests__/batched-dispatcher.test.ts
```diff
@@ -49,10 +49,13 @@ describe('Batching', () => {
   beforeEach(() => {
     jest.resetAllMocks()
     jest.restoreAllMocks()
+    jest.useFakeTimers()
   })
 
   afterEach(() => {
-    window.close()
+    // clear any pending sendBatch calls
+    jest.runAllTimers()
```
silesky, line 57: Does jest.clearAllTimers() work here instead?

silesky:
> The build failure looks like it's possibly not finding the cached node_modules - I think this was changed recently so maybe once @silesky's monorepo changes go in I can rebase and see if that resolves it.

Should work.


And thank you for doing this!

src/plugins/segmentio/__tests__/normalize.test.ts
chrisradek: Yup, you've found my 'find and replace' edit :smile:

src/plugins/segmentio/__tests__/batched-dispatcher.test.ts
chrisradek: I tried that, but for these tests we need the timers to actually run to completion so that the batched dispatcher flushes all its events.

If I just clear the timeouts instead what ends up happening is later on when we dispatch a `beforeunload` event it triggers the batched dispatchers we created in earlier tests to flush their events and this state leaks into other tests.

`window.close()` was triggering the `beforeunload` before so it wasn't a problem but since we can't use that anymore (unless each test is in its own file or making updates to jest's teardown) running all timers ensures the buffered events are flushed before starting the next test.

silesky: For a module we're there's some impure state, I think it's best to have the  import of the module as part of the beforeEach (and then reset in the afterEach), so each test case is isolated with a "fresh" import.

chrisradek: I tried some simple refactoring but this looks like it'll be more involved. I think ultimately out of scope for just upgrading jest :smile:

silesky (approved):
LGTM

PR #479, opened 2022-06-08
[bugfix] ignore invalid URLs with TLD lookup

chrisradek:
Turns out we had an `it.only` in the auto-track tests so most of those tests weren't being ran. Many of the tests were failing. This PR re-enables the auto-track tests and makes the `tld` check more resilient by parsing URLs internally.

Discovered as part of #478


pooyaj (approved):
💥

PR #482, opened 2022-06-09
enhance typechecking and remove tests from build

silesky:
This fixes tests being included in our npm package.

### Before:
```
npm notice === Tarball Details ===
npm notice name:          @segment/analytics-next
npm notice package size:  1.1 MB
npm notice unpacked size: 4.5 MB
npm notice total files:   940
```
### After:
```
npm notice === Tarball Details ===
npm notice name:          @segment/analytics-next
npm notice package size:  873.7 kB
npm notice unpacked size: 2.8 MB
npm notice total files:   568
```

This _also_ creates a base tsconfig that include e2e tests, so type checking actually works on those files.
image


chrisradek (approved):
Did you happen to check the list of files included before and after to make sure only test-related files were removed? I'm sure you did but since we aren't running a test against the built package I have to ask :smile:

silesky:
> Did you happen to check the list of files included before and after to make sure only test-related files were removed? I'm sure you did but since we aren't running a test against the built package I have to ask 😄

yeah, I did manually check. But yeah, those tests are a todo ;-)

PR #483, opened 2022-06-09
[bugfix] set event timestamp at event creation

chrisradek:
This PR updates when the `timestamp` property of an event gets set. Currently for `AnalyticsBrowser`, it is set right before sending the event to Segment. This means it's almost exactly the same as the `sentAt` timestamp, and is updated every time we retry sending the event.

Instead it should reflect the time that the event occurred (i.e. when the event was created) and should be unchanged across retries. This is the behavior analytics.js (classic) exhibits.

This also puts us in accordance with the [spec for timestamps](https://segment.com/docs/connections/spec/common/#timestamps). Through some testing I found that the `timestamp` we set on the client-side becomes the `originalTimestamp` in the event after it hits the tracking api. The `timestamp` field then gets replaced by the `originalTimestamp` with clock-skew taken into account.

## Testing

I tested this in my own app and also added some tests to ensure `timestamp` is being added. We actually have many more tests that show `timestamp` is added to the event sent to the tracking API so I didn't have to write too many new ones :smile:

### DIFF
--- a/packages/browser/src/core/events/__tests__/index.test.ts
+++ b/packages/browser/src/core/events/__tests__/index.test.ts
@@ -101,6 +101,11 @@ describe('Event Factory', () => {
       expect(track.messageId).toContain('ajs-next')
     })
 
+    test('adds a timestamp', () => {
+      const track = factory.track('Order Completed', shoes)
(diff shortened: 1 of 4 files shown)


silesky (approved):
💯

PR #484, opened 2022-06-10
make tests less brittle

silesky:
- make lighthouse less brittle
- fix batched dispatcher brittle test

image

### DIFF
--- a/packages/browser/e2e-tests/performance/ajs-perf-browser.test.ts
+++ b/packages/browser/e2e-tests/performance/ajs-perf-browser.test.ts
@@ -28,8 +28,8 @@ describe('Performance', () => {
     const results: any = await gatherLighthouseMetrics(analyticsStub.browserPage)
     const audits = results.audits // Lighthouse audits
 
-    // lighthouse perf score of 100
-    expect(results['categories'].performance.score).toEqual(1)
+    // lighthouse perf score
+    expect(results['categories'].performance.score).toBeGreaterThanOrEqual(0.99)
     // no render blocking resource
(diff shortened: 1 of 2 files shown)


pooyaj (approved):

chrisradek:
The callback one is really interesting. I just ran the following test 100 times and 3/100 failed:
```ts
it('test sleep', (done) => {
      const startTime = performance.now()

      setTimeout(() => {
        const endTime = performance.now()
        try {
          expect(endTime - startTime).toBeGreaterThanOrEqual(100)
          done()
        } catch (err) {
          done(err)
        }
      }, 100)
    })
```

`setTimeout` is supposed to be invoked at or after the specified ms has elapsed, but not before - but testing is showing otherwise. I used performance.now() to remove the system clock from the equation - bizarre.

pooyaj:
@chrisradek this could be due to use of fakeTimers in Jest 🤔 ? I'm not sure tho ...

chrisradek:
@pooyaj I also tried a plain JS script in node, no frameworks:
```js
function test() {
  return new Promise((resolve, reject) => {
    const startTime = performance.now();

    setTimeout(() => {
      const endTime = performance.now();
      const delta = endTime - startTime;
      console.assert(delta >= 100, `Expected >= 100, got ${delta}`);
      resolve();
    }, 100);
  });
}

async function runner() {
  for (let i = 0; i < 100; i++) {
    await test();
  }
}

runner();
```

And still see the same results. Super interesting...

silesky:
Javascript is running so fast that 100ms only takes 99

PR #485, opened 2022-06-10
Upgrade js-cookie

pooyaj:
Re-doing an old PR, as the codebase has changed substantially since then.


danieljackins (approved):

packages/browser/package.json
```diff
@@ -53,7 +53,7 @@
     "@segment/facade": "^3.4.9",
     "@segment/tsub": "^0.1.12",
     "dset": "^3.1.2",
-    "js-cookie": "^2.2.1",
+    "js-cookie": "3.0.1",
```
oleg-semyonov, line 56: Hey @pooyaj, thanks for pushing this update through! 🙏🏻 
However, I would suggest not to pin the dep version (not sure if it was done accidentally or on purpose). That would prevent consumer application from using latest `js-cookie` versions, and I see you're using semver ranges for every other dependency in this project.

sethjeffery: Please can this be changed to no longer pin the version, for the reason stated above?

PR #487, opened 2022-06-10
chore/fix-vulnerable-dependencies

silesky:
side benefit: we can remove all these old saucelabs artifacts


pooyaj (approved):
Even removing packages increases the bundle size. Life is unfair 😭

PR #492, opened 2022-06-15
Dont throw on mutex failure

danieljackins:
Make sure we're just logging the error instead of throwing, especially since if the error is thrown in the setTimeout call it won't be catchable


chrisradek (approved):
Best PR I've reviewed this week! :smile:

Issue #494, opened 2022-06-17
Link to "existing plugins" in README is broken


dobesv:
To reproduce:

1. Go to https://github.com/segmentio/analytics-next
2. Scroll to the bottom
3. Click "existing plugins"

Maybe it's supposed to link to https://github.com/segmentio/analytics-next/tree/master/packages/browser/src/plugins ?

silesky:
Thanks @dobesv. Fixed in https://github.com/segmentio/analytics-next/pull/495/files

PR #495, opened 2022-06-19
fix markdown links, add markdown link checker

silesky:
#494

### DIFF
--- a/packages/browser/README.md
+++ b/packages/browser/README.md
@@ -120,7 +120,7 @@ const App = () => {
 
 More React Examples:
 
-- Our [playground](example/pages/_app.tsx) (written in NextJS) -- this can be run with `yarn dev`.
+- Our [playground](/examples/with-next-js/) (written in NextJS) -- this can be run with `yarn dev`.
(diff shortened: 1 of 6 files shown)


packages/browser/README.md
```diff
@@ -223,7 +223,7 @@ export const lowercase: Plugin = {
 }
 ```
 
-For further examples check out our [existing plugins](https://github.com/segmentio/analytics-next/tree/master/src/plugins).
+For further examples check out our [existing plugins](/packages/browser/src/plugins).
```
chrisradek, line 226: Is it possible to use the full url with the linter? I suspect these links will be broken on NPM if we change them to relative paths.

silesky: That's a good point -- `npm` _should_ resolve it automatically, but I just checked, and our repository field in the package.json looks wrong. My last commit should fix it. 

See: https://stackoverflow.com/questions/62575382/relative-link-from-readme-md-to-another-file-in-package-rendered-in-npmjs

silesky: I also added the repository field to the yarn constraints file, so its enforced.

pooyaj (approved):
❇️

PR #496, opened 2022-06-20
Introduce AnalyticsCore without any of the deprecated logic (non-breaking)

gmathieu:
This PR introduces `analytics-core.ts` which extracts all logic that is not deprecated from `analytics.ts`. Ideally, we'd rename `Analytics` to something like `AnalyticsLegacy` but my goal was to keep 100% backwards compatibility. I'm not married to the name `AnalyticsCore`, I even thought about naming it `AnalyticsNext`. Open to suggestions...

Splitting the core logic from `analytics.ts` enables the following:
- 3rd parties can leverage the core analytics features without the legacy overhead
- removes the dependency on `global` which some bundlers are removing by default (ex. Angular)
- reduces package size significantly (see below)

Note: `AnalyticsNode` could benefit from using `AnalyticsCore` directly since most of the `analytics.ts` functionality is browser related or deprecated but this would introduce a breaking change.

### Bundling Analytics

Screen Shot 2022-06-20 at 10 39 52 AM

### Bundling AnalyticsCore
Screen Shot 2022-06-20 at 10 40 02 AM


silesky:
@gmathieu As you may have noticed by the merge conflicts, there's been some prior art around some of this refactoring. Naming things is hard -- the  working definition of "core" that we're using is "code that is not specific to a runtime". I actually started to created an interface called 'AnalyticsClassic' and I know we are also going to be removing a bunch of these "deprecated methods" somewhat soon, as most of them are no-ops.  
If you want to see the thing I was hacking on:
https://gist.github.com/silesky/9fb1c33afa32c1a5e2c215204f93bdf2

gmathieu:
oh wow, sorry my branch is so out of date. I pulled the latest upstream this morning but now realize I didn't apply it to my local branch 🤦. 

Thanks for sharing the vision and your gist. I would recommend  that `AnalyticsClassic` extends `AnalyticsCore` and not the other way around. This allows `AnalyticsCore` to be the base for both browser and node.

The direction of slipping browser and node makes a lot of sense. I would take it a step further and define even more boundaries. Now that you have the beginnings of a monorepo, it might make sense to package the following:
- `packages/browser` (@segment/analytics-browser): as you suggested
- `packages/core` (@segment/analytics-core): the core folder with no side effects (ex. the use of `global`) and legacy functionality removed from `core/analytics/index.ts`. You'd likely include some of the lib folder shared across node and browser
- `packages/plugins` (@segment/analytics-plugins): the plugins folder (could also be part of core if too much of a hassle)
- `packages/node` (@segment/analytics-node): the node version which initializes core + the node plugin
- `packages/standalone` (not published): the standalone version hosted on segment.io. Since this is compiled by Webpack, it's a totally different environment with globals that shouldn't be defined in other packages (ex. `ASSET_PATH`)
- `packages/next` (@segment/analytics-next): if you choose to maintain the deprecated classic functionality as it is today 

If you're curious to see how we leverage analytics-next today, here's [our implementation](https://github.com/variancehq/variance-js/blob/7c44b646b570c676130029b291f75e017777e1f7/src/VarianceJs.ts). As you can see we have zero dependency on the classic functionality. Our implementation is similar to the node version: initialize AnalyticsCore + configure a few plugins.

Let me know if you'd like me to help.

Issue #325, opened 2021-11-12
Unsafe global usage


mwood23:
Always targeting `global` in the code breaks the build for some bundlers like Vite unless you stub it in the define function. However, doing so can break other packages. I would suggest safely getting global like [Sentry](https://github.com/getsentry/sentry-javascript/blob/20d5fa73e823f0d05dacefed9c8b769f2ca57c8e/packages/utils/src/global.ts#L36).

```
/**
 * Safely get global scope object
 *
 * @returns Global scope object
 */
export function getGlobalObject<T>(): T & SentryGlobal {
  return (isNodeEnv()
    ? global
    : typeof window !== 'undefined' // eslint-disable-line no-restricted-globals
    ? window // eslint-disable-line no-restricted-globals
    : typeof self !== 'undefined'
    ? self
    : fallbackGlobalObject) as T & SentryGlobal;
}
```

mwood23:
Here's a patch package for anyone that's blocked by this:

```patch
diff --git a/node_modules/@segment/analytics-next/dist/pkg/analytics.js b/node_modules/@segment/analytics-next/dist/pkg/analytics.js
index 80f0914..790c2bd 100644
--- a/node_modules/@segment/analytics-next/dist/pkg/analytics.js
+++ b/node_modules/@segment/analytics-next/dist/pkg/analytics.js
@@ -13,7 +13,7 @@ import { PersistedPriorityQueue } from './lib/priority-queue/persisted';
 var deprecationWarning = 'This is being deprecated and will be not be available in future releases of Analytics JS';
 // reference any pre-existing "analytics" object so a user can restore the reference
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-var globalAny = global;
+var globalAny = {};
 var _analytics = globalAny.analytics;
 var Analytics = /** @class */ (function (_super) {
     __extends(Analytics, _super);
diff --git a/node_modules/@segment/analytics-next/dist/pkg/lib/load-script.js b/node_modules/@segment/analytics-next/dist/pkg/lib/load-script.js
index 43e22d7..de9e70a 100644
--- a/node_modules/@segment/analytics-next/dist/pkg/lib/load-script.js
+++ b/node_modules/@segment/analytics-next/dist/pkg/lib/load-script.js
@@ -18,7 +18,7 @@ export function loadScript(src, attributes) {
     }
     return new Promise(function (resolve, reject) {
         var _a;
-        var script = global.window.document.createElement('script');
[...]

PR #502, opened 2022-06-23
chore/fix vite add example

silesky:
This PR:
-  fixes a regression that broke vite.
- allows this SDK to support `vite` without any extra boilerplate such as adding a `window.global` object.
- removes unneccessary? references to `global`


packages/browser/src/lib/load-script.ts
```diff
@@ -27,7 +27,7 @@ export function loadScript(
   }
 
   return new Promise((resolve, reject) => {
-    const script = global.window.document.createElement('script')
+    const script = window.document.createElement('script')
```
silesky, line 30: @pooyaj FYI -- sanity check

packages/browser/src/plugins/ajs-destination/loader.ts
```diff
@@ -17,7 +17,7 @@ function obfuscatePathName(pathName: string, obfuscate = false): string | void {
 function recordLoadMetrics(fullPath: string, ctx: Context, name: string): void {
   try {
     const [metric] =
-      global.window?.performance?.getEntriesByName(fullPath, 'resource') ?? []
```
silesky, line 20: @pooyaj FYI -- sanity check

README.md
```diff
@@ -123,18 +121,9 @@ More React Examples:
 - Our [playground](/examples/with-next-js/) (written in NextJS) -- this can be run with `yarn dev`.
 - Complex [React example repo](https://github.com/segmentio/react-example/) which outlines using the [Segment snippet](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-quick-start) and using the [Segment npm package](https://github.com/segmentio/react-example/tree/main/src/examples/analytics-package).
 
-### using `Vite` with `Vue 3`
-
-1. add to your `index.html`
-
-```html
-<script>
-  window.global = window
```
silesky, line 132: boilerplate no longer needed

packages/browser/src/core/analytics/index.ts
```diff
@@ -34,9 +34,9 @@ const deprecationWarning =
   'This is being deprecated and will be not be available in future releases of Analytics JS'
 
 // reference any pre-existing "analytics" object so a user can restore the reference
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const globalAny: any = global
-const _analytics = globalAny.analytics
+const _analytics: Analytics | undefined = (
+  (typeof window !== 'undefined' ? window : global) as any
```
chrisradek, line 38: This looks safe to me since you're doing the typeof check on window.

I'm assuming `global` was used instead of `window` so that `AnalyticsNode` wouldn't break. This didn't break our UMD build since we use webpack and webpack was polyfilling `global` to use globalThis/this/window, and we expected other consumers to polyfill it.

I think everywhere else you removed `global` is never executed by `AnalyticsNode` so that's why we don't need the check in those places.

chrisradek (approved):
Looks great, thanks for including a vite example!

You can probably close https://github.com/segmentio/analytics-next/issues/325 with this.

pooyaj (approved):
💣

packages/browser/src/core/analytics/index.ts
silesky: Added a getGlobal function

https://github.com/segmentio/analytics-next/pull/502/files#diff-5e3d9ac53f16f5c22406dbdc9b21686a6b9694960dfac623c3340c14e19a6841

packages/browser/src/lib/get-global.ts
```diff
@@ -0,0 +1,16 @@
+// This an imperfect polyfill for globalThis
+export const getGlobal = () => {
```
chrisradek, line 2: With this, I wonder - can we turn off webpack polyfilling `global` then? That might save us some bytes on the UMD build.
https://webpack.js.org/configuration/node#node

silesky: Our webpack build is only targeting browser -- there's no extra polyfill.

chrisradek: > Our webpack build is only targeting browser -- there's no extra polyfill.

I'm not sure I follow. If you look at the current standalone build you'll find that webpack has polyfilled the global object for us. You've created a new `getGlobal()` function that accomplishes the same goal - but since it references `global` itself I would expect webpack to still polyfill `global` when creating the UMD bundle.

silesky: Maybe you can show me where the actual polyfill is happening? I'm still not seeing it.

yordis: Would this get compiled even when I use the `npm` version of the package?

PR #504, opened 2022-06-27
add node scaffolding

silesky:
This adds the node scaffolding (in preparation of breaking up node).

Things this PR adds:
- ts build script that creates both `esm` and `cjs`.
- tests
- CI

Things this does not do:
- Adds a "publish" workflow  (it might make sense to move to something purposefully built for monorepos, like `lerna publish`).

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -1,12 +1,7 @@
 {
   "name": "@segment/analytics-node",
   "version": "1.0.0",
-  "repository": {
-    "type": "git",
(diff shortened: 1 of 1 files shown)


packages/node/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@segment/analytics-node",
+  "version": "1.0.0",
+  "repository": {
```
chrisradek, line 4: Can we add `private: true` until we're ready to publish?

chrisradek (approved):

PR #505, opened 2022-06-28
Updates analytics error handling to support checking delivery failures on Context

chrisradek:
This PR attempts to standardize how plugin errors impact analytics API calls and provides a straightforward way to check if event delivery failed.

## Background

Errors in `enrichment`, `destination`, and `after` plugins don't cause analytics calls to throw. `before` plugins do, but what they throw depends on if retries are enabled or not. If there are no retries, their errors bubble up to the analytics call. If there are retries, the event's `Context` itself gets thrown.
To further complicate things, source middleware (implemented as a `before` plugin) does not bubble errors if `next` is never called, but does otherwise.

Only `before` plugins and source middleware errors can cause event delivery to fail entirely, but the inconsistent behavior makes it difficult to write code that can react to these failures.

## Changes

The following has been changed:
- all plugins now act consistently in that errors are no longer bubbled up to the analytics calling code (e.g. `analytics.track()`)
- `before` plugins now support cancelation: this is a bug fix, existing test missed that this was broken
- `Context.failedDelivery()` is added, so now the following works:
  ```ts
  const context = await analytics.track('event');
  if (context.failedDelivery()) {
    // log failure, retry, etc
  }
  ```

This also has the benefit that callbacks can now be invoked even on failures, so you can do:
```ts
analytics.track('event', (context) => {
  if (context.failedDelivery()) {
    // log failure, retry, etc
  }
})
```
Previously callbacks weren't being invoked if the analytics call threw.

### Chore

Also updated the vscode launch.json so the actions work in `packages/browser`.


packages/browser/src/core/analytics/__tests__/integration.test.ts
```diff
@@ -0,0 +1,199 @@
...
+import { Analytics } from '../index'
+import {
+  AfterPlugin,
+  BeforePlugin,
+  DestinationPlugin,
+  EnrichmentPlugin,
+} from './test-plugins'
+
+describe('Analytics', () => {
+  describe('plugin error behavior', () => {
```
pooyaj, line 14: amazing 🙌

pooyaj (approved):

chrisradek:
Just realized I missed one scenario - `before` plugin fails the 1st time then succeeds on retry. Will need to update the context in this case so that `context.getFailedDelivery()` does not return anything.

PR #506, opened 2022-06-29
[bugfix] fixes race condition where some analytics calls were ignored while analytics was loading via snippet

chrisradek:
This PR fixes a race condition that occurs when loading analytics via the snippet. Any calls made on the global analytics object between `flushPreBuffer` and `flushFinalBuffer` during load would be ignored.
https://github.com/segmentio/analytics-next/blob/62dcce271dac64c89849cbb060c23764ea006834/packages/browser/src/browser/index.ts#L243-L269

This is due to analytics only reading the `window.analytics` array during `flushPreBuffer`.

## How does this update fix the race condition?

2 changes:
1. Now we get the buffered list of calls from window.analytics before every flush task during load.
2. `getSnippetWindowBuffer` is removes elements from window.analytics so that we don't get duplicates by calling it multiple times.

Edit: Never try to solve a race condition late at night...

### DIFF
--- a/packages/browser/src/core/buffer/snippet.ts
+++ b/packages/browser/src/core/buffer/snippet.ts
@@ -4,17 +4,21 @@ import type {
   PreInitMethodParams,
 } from '.'
 
+export function transformSnippetCall([
+  methodName,
+  ...args
+]: SnippetWindowBufferedMethodCall): PreInitMethodCall {
+  return {
+    method: methodName,
+    resolve: () => {},
+    reject: console.error,
+    args,
+    called: false,
+  }
+}
+
 const normalizeSnippetBuffer = (buffer: SnippetBuffer): PreInitMethodCall[] => {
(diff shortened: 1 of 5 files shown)


packages/browser/src/core/buffer/snippet.ts
```diff
@@ -32,8 +36,7 @@ type SnippetBuffer = SnippetWindowBufferedMethodCall[]
  */
 export const getSnippetWindowBuffer = (): PreInitMethodCall[] => {
```
zikaari, line 37: Now that the behaviour has been changed, it's more like `popSnippetWindowBuffer`, similar to `git stash pop` and `Array.prototype.pop`. Usually `get*` functions are non-mutating.

Just a thought, totally not necessary.

zikaari (approved):
pretty nice!

pooyaj (approved):

PR #510, opened 2022-06-30
add changesets and documentation

silesky:
(todo: the changeset publish CI will not work until we add an `NPM_TOKEN` (with write access) to github secrets.)

changeset docs: https://github.com/changesets/changesets#documentation

TLDR: When developing a feature/bug/chore, run `yarn changeset` in your branch and commit the generated artifacts. When it’s time to release, merge the automatically created Version Package PR. This will kick off the release process, which results in the CI triggering `yarn release` to push tags to GitHub / publish packages to NPM.

**Why publishing pipelines are extra complicated in a monorepo**…
        - How do you ensure the package you're releasing is also released in sync with all of its dependencies? If analytics-browser  and analytics-node depend on analytics-core, everytime analytics-core is updated, we need to ensure that analytics-browser and analytics-node are also updated.
        - You can no longer easily look at the commit history and see what should package or packages that commit affects. Even knowing what file it touched is not enough — if I update a file in core, I need to release core as a potentially major release and also, all packages that  depend on core might need to released as minor releases. 
        - Generating a changelog is also annoying, it means figuring out which recent commits affect a specific packages and collecting them (manually) -/ there could be a lot of noise.
        - Tools like semantic release and changeset force the user to create some kind of metadata in their PRs so the system can easily map each merge to its relevant package or packages, and can calculate the appropriate bump.  

Changeset upsides (over lerna with conventional commits)
- way easier to fix if a non-standard commit gets onto master; does not heavily require annoying precommit hooks or magic tags like “feature” — you can run this at the beginning  and not worry about commit messages the rest of the time
- user can have multiple commits and they can still all be squashed by GitHub without a user having to conform to the semantic release format.
- Does not rely on a user having to squash all of their commits very carefully / break them up into special selected tags
- Can be created at any time (during a PR would be ideal)
- PR automatically opens via GitHub 

Changesets release bot:

Version packages example:
image


package.json
```diff
@@ -14,10 +14,16 @@
     "lint": "echo 'Linting all workspaces...' && yarn constraints && yarn workspaces foreach -p run lint",
     "build:dev": "yarn workspace @segment/analytics-next build:dev",
     "dev": "yarn workspace @segment/analytics-next exec make dev",
-    "postinstall": "husky install"
+    "postinstall": "husky install",
+    "version": "changeset version && yarn version-run-all",
+    "release": "yarn prepublish-run-all && changeset publish",
+    "prepublish-run-all": "yarn workspaces foreach -pt --no-private run prepublish",
```
silesky, line 20: `changeset` does not (at first glance) to support lifecycle scripts natively. I created an issue.

chrisradek (commented):
I like this method of generating changelogs!

A question and a suggestion.

Q: What does the final changelog look like? It's at the repo root not per package right?

Suggestion: Can you update the PR template to also include instructions to generate a changeset?

.changeset/fifty-shoes-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@segment/analytics-next': patch
+---
+
+Add changeset publish pipeline!
```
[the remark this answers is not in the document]
silesky: Correct; it's just for demonstration purposes (to trigger the bot). [Not everything needs a changeset.](https://github.com/changesets/changesets/blob/main/docs/intro-to-using-changesets.md#not-every-change-requires-a-changeset)! I can revert this later.

silesky:
> Q: What does the final changelog look like? 
The final changelog is here: https://github.com/segmentio/analytics-next/pull/511/files (BTW, you can do `changeset version` to generate the changelog at any time on your local). Yes
> It's at the repo root not per package right?
It's per package (e.g. browser/changelog.md, node/changelog.md). 
> Suggestion: Can you update the PR template to also include instructions to generate a changeset?
Sure, I can update the PR template to suggest that user's run `yarn changeset` if you think that's useful.

chrisradek (approved):
Love that we'll have a changelog now versus combing through release notes!

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,9 +1,12 @@
...
+- Please add:
+  - a description of your PR, including the what and why
+  - a changeset (if applicable)
 
 Also make sure to describe how you tested this change, include any gifs or screenshots you find necessary.
-
 --->
+
+
+[] I've included a changeset (psst. run `yarn changeset`. Read about changesets [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)).
```
pooyaj, line 12: 🙌

package.json
```diff
@@ -33,5 +40,9 @@
     "ts-jest": "^28.0.4",
     "ts-node": "^10.8.0",
     "typescript": "^4.6.4"
+  },
+  "resolutions": {
+    "@segment/analytics-next": "workspace:*",
```
pooyaj, line 45: just for my curiosity ... what does this do?

pooyaj (approved):
amazing 💥

package.json
silesky: I stole this pattern in another repo, but I don't remember which one.

It should allow us to have fixed version numbers in our packages for the sake of publishing, but when developing locally or running tests, this resolutions key will always guarantee that both our dependencies and any transient dependencies will use the "current" workspace, rather than (accidentally?) using the version that is published (or maybe not-yet published) to npm.  For example, if we have a not-yet-published `@segment/utils` that is updated, and is a transient dependency of `analytics-core` that is installed in `analytics-node`. Will an install pick up on the local changes when we run our tests? I could be wrong, but I don't _think_ so.

silesky: You can see an example here: https://github.com/facebook/jest/blob/main/package.json#L172

PR #516, opened 2022-07-01
use buildkite for publishing

silesky:


package.json
```diff
@@ -17,8 +17,9 @@
     "postinstall": "husky install",
     "changeset": "changeset",
     "update-versions-and-changelogs": "changeset version && yarn version-run-all",
-    "release": "yarn prepublish-run-all && changeset publish && git push --follow-tags",
-    "prepublish-run-all": "yarn workspaces foreach -pt --no-private run prepublish",
+    "release": "yarn prepare-run-all && changeset publish && git push --follow-tags && yarn postpublish-run-all",
```
silesky, line 20: I've found that `changeset publish` doesn't run lifecycle scripts such as 'yarn prepare'. I created an issue around this; for now, I'm just running all lifecycle script for every release, as they should be idempotent.

PR #522, opened 2022-07-02
Add / improve release CI

silesky:
This PR: 
- fixes release conditional so publish only happens on when Version Packages is merged onto master branch.
- tidies up CI and release script to add variable validation, safeguards, naming, etc

This PR does not:
- address github "release" tagging. Many of the most popular monorepos do not use GitHub releases at all FWIW, and just rely on the changelog. The new format will be something like  @segment/analytics-next@1.39.1); they should be created automatically on `changeset publish`, but this is actually impossible to test in a sandbox without actually publishing something, and the `changeset tag` behavior has at least one outstanding issue that makes its [behavior quite different](https://github.com/changesets/changesets/issues/833) from `changeset publish`. I will be adding both auto tag creation and looking at migrating our old tags, but pushing a tag manually on release continues to work as before.


.buildkite/pipeline.yml
```diff
@@ -108,17 +93,45 @@ steps:
...
-  #     queue: v1
-  #   commands:
-  #     - npm config set "//registry.npmjs.org/:_authToken" $${NPM_TOKEN}
-  #     - HUSKY=0 yarn install --immutable
-  #     - yarn release
-  #   plugins:
-  #     - ssh://git@github.com/segmentio/cache-buildkite-plugin#v2.0.0:
-  #         key: "v1.1-cache-dev-{{ checksum 'yarn.lock' }}"
-  #         paths: [".yarn/cache/"]
+  - label: "Release / Publish Packages :rocket:"
```
silesky, line 114: Updated release step ->

pooyaj (approved):

PR #527, opened 2022-07-07
add seperate types to node build

silesky:
1. add separate types folder (not specific to any module format).
2. publish source maps (similar to browser) -- common in popular packages like sentry, commercetools, etc.
3. publish declaration maps (enable go-to-definition) -- fairly common (e.g. react-hook-form)


chrisradek (approved):
LGTM!

PR #528, opened 2022-07-07
add push tags, update workflow

silesky:


chrisradek (approved):
Looks good!

With regards to tags, have you thought about how to name them once we start publishing more than 1 package? (Not something needed for this PR)

silesky:
@chrisradek the tag format from here on out  is <package_name>@<version number>. So the next release will be something like `@segment/analytics-next@1.39.2`

PR #533, opened 2022-07-11
add initial analytics core package

silesky:
This is mostly boilerplate -- the only abstraction imported was `emitter`, to keep this small. This PR is non-blocking. The new package will be automatically published as `@segment/analytics-core@1.0.0` on the next release (regardless of version).

Now that we have `core`, there is some refactoring of scripts to allow us to build individual packages in a way that also builds those required workspace dependencies without having to explicitly name them. 

e.g.  build analytics-next, starting by running `build` recursively on any workspace dependencies.
```
yarn workspaces foreach -vtR --from '@segment/analytics-next' run build  
```

this reminds me of a TODO:
- dependency-aware script / build manager  for our CI -- something like Make, but maybe better (e.g. `turbo run build --filter=./packages/*.`)


packages/browser/Makefile
```diff
@@ -65,10 +64,6 @@ ci:
 	bash ./scripts/ci.sh
 .PHONY: ci
 
-size: ## Verify the final size of Analytics-Next
```
chrisradek, line 68: Do we still do the size check somewhere?

silesky: Yes, that make target wasn’t being used

chrisradek: Ah, I’ve been using that, but can switch to yarn :smile:

PR #535, opened 2022-07-13
Fix node sdk

adisbanda:


adisbanda:
Apologies for opening this. It was a mistake!

PR #541, opened 2022-07-15
Add typings instructions from snippet users and export Snippet type

silesky:
In the immediate future, we may want to also publish a package that just contains AnalyticsSnippet  (and other, related types) — or update the segment-analytics package (and attempt to become maintainers of it) on DefinitelyTyped. This is fine for now.


pooyaj (approved):
amazing  🔥🔥🔥

PR #545, opened 2022-07-15
update dev experience so running 'dev' watches all dependencies

silesky:
- create a watch command that can watch all packages (and rebuild changes)
- create a dev command for each example that will watch all dependencies (and rebuild changes automatically, no matter which workspace in the dependency tree triggered the change)

This is more PoC... we have jest / tests... most people wouldn't use these "examples" for development -- However, if we're going to support a "dev" command, we might as well make it work as expected!


zikaari (approved):
nice!

PR #546, opened 2022-07-16
Create github release from tags after tags are pushed

silesky:
As the final automation step, automatically create a github release once tags are pushed (in the release step).

Github actions includes the gh CLI (pre-installed) by default.

This doesn't include logic to add "notes" ATM -- we can add some logic to pass the "notes property" and link to the changelog later. For now, we can manually edit the release after-the-fact.

### DIFF
--- a/scripts/create-release-from-tags.sh
+++ b/scripts/create-release-from-tags.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# Create github release(s) for every tag of the current commit. Use the github cli.
+
+tags=$(git tag --points-at HEAD)
+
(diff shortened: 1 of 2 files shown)


scripts/create-release-from-tags.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Create github release(s) for every tag of the current commit. Use the github cli.
+
+tags=$(git tag --points-at HEAD)
+
+for tag in "${tags[@]}"; do
+  gh release create "$tag"
```
silesky, line 8: supports --notes

PR #547, opened 2022-07-18
refactor tsconfig -- make consistent

silesky:
- Browser:
   -  generate a `types` folder in `browser`, similar to the rest of the packages
     - reference tsconfig.build.json in webpack, which is the correct tsconfig to reference.
     - no longer include the `src` folder (reduces package size significantly, and it's much more common not to include src. does not affect any debugging within the monorepo, and if really needed, can always "add folder to workspace" for external sources.)
     
 - make tsconfig and tsconfig.build.json consistent across packages -- tsconfig.build.json will be a build configuration; therefore, it has an `outDir` and `typesDir` and be responsible for specifics around generating the `dist` folder. `tsconfig.json` is a base tsconfig and should contains information around the overall environment (e.g. target, libs), with enough info run lint / jest / etc correctly without building a dist folder or generating declarations.

### DIFF
--- a/.changeset/tidy-chairs-whisper.md
+++ b/.changeset/tidy-chairs-whisper.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-core': patch
+---
+
+Fix "failed to parse source map" in analytics-core ([#420](https://github.com/segmentio/analytics-next/issues/420))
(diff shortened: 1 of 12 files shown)


zikaari (approved):
Nice!

PR #552, opened 2022-07-20
Suspend event deliveries until middlewares are ready

zikaari:
This patch fixes an important issue where a middleware will miss out
on augmenting any events while it's being prepped up. Those events will
reach all the destinations as-is without going through the said
middleware.

With this patch, `addSourceMiddleware` is now classified as a critical
task, and thus will seize all event deliveries until all middlewares are
in an operational state. The suspension happens internally, and the
public API is unaffected.

### DIFF
--- a/packages/browser/src/core/task/task-group.ts
+++ b/packages/browser/src/core/task/task-group.ts
@@ -1,0 +1,31 @@
+import { isThenable } from '../../lib/is-thenable'
+
+export type TaskGroup = {
+  done: () => Promise<void>
+  run: <Operation extends (...args: any[]) => any>(
+    op: Operation
+  ) => ReturnType<Operation>
+}
+
+export const createTaskGroup = (): TaskGroup => {
+  let taskCompletionPromise: Promise<void>
+  let resolvePromise: () => void
+  let count = 0
+
+  return {
+    done: () => taskCompletionPromise,
+    run: (op) => {
+      const returnValue = op()
+
+      if (isThenable(returnValue)) {
+        if (++count === 1) {
+          taskCompletionPromise = new Promise((res) => (resolvePromise = res))
+        }
+
+        returnValue.finally(() => --count === 0 && resolvePromise())
+      }
+
+      return returnValue
+    },
+  }
+}
--- a/packages/browser/src/lib/is-thenable.ts
+++ b/packages/browser/src/lib/is-thenable.ts
@@ -2,7 +2,7 @@
  *  Check if  thenable
  *  (instanceof Promise doesn't respect realms)
  */
-export const isThenable = (value: unknown): value is Promise<unknown> =>
+export const isThenable = (value: unknown): boolean =>
   typeof value === 'object' &&
   value !== null &&
   'then' in value &&
--- a/packages/browser/src/lib/is-thenable.ts
+++ b/packages/browser/src/lib/is-thenable.ts
@@ -2,7 +2,7 @@
  *  Check if  thenable
  *  (instanceof Promise doesn't respect realms)
  */
-export const isThenable = (value: unknown): boolean =>
+export const isThenable = (value: unknown): value is Promise<unknown> =>
   typeof value === 'object' &&
   value !== null &&
   'then' in value &&
--- a/.changeset/ten-lions-look.md
+++ b/.changeset/ten-lions-look.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': patch
+---
+
+Suspend event deliveries until middlewares are ready
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -41,7 +41,7 @@
   "size-limit": [
     {
       "path": "dist/umd/index.js",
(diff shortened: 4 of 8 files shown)


silesky:
Changeset needed (either patch or minor)

danieljackins (commented):
implementation looks good to me!

packages/browser/src/core/task/task-group.ts
```diff
@@ -0,0 +1,22 @@
+export class TaskGroup {
+  private promise!: Promise<void>
```
silesky, line 2: Can we have tests that cover this code path?

packages/browser/src/core/task/task-group.ts
```diff
@@ -0,0 +1,22 @@
...
+  private res!: () => void
+  count = 0
+
+  done = () => this.promise
+  run = <Operation extends (...args: any[]) => any>(
```
silesky, line 7: Another way to type this is “<ReturnT, Operation extends (…args: unknown[]) => T>(op: Operation): ReturnT

packages/browser/src/core/task/task-group.ts
```diff
@@ -0,0 +1,29 @@
...
+  let taskCompletionPromise: Promise<void>
+  let resolvePromise: () => void
+  let count = 0
+
+  return {
+    done: () => taskCompletionPromise,
+    run: (op) => {
+      const returnValue = op()
+
+      if (returnValue instanceof Promise) {
```
silesky, line 18: @chrisradek alerted me to the possible gotchas of “instanceof Promise” (does not work in all environments)  — there should be an “isPromise” or “isThenable” helper in utils that we used instead.

danieljackins (approved):

packages/browser/src/core/task/task-group.ts
zikaari: Fixed up

packages/browser/src/core/task/task-group.ts
```diff
@@ -0,0 +1,31 @@
...
+    done: () => taskCompletionPromise,
+    run: (op) => {
+      const returnValue = op()
+
+      if (isThenable(returnValue)) {
+        if (++count === 1) {
+          taskCompletionPromise = new Promise((res) => (resolvePromise = res))
+        }
+
+        returnValue.finally(() => --count === 0 && resolvePromise())
```
chrisradek, line 25: Can you confirm that the polyfill we use in the standalone version (CDN version) includes Promise.finally? I think this is the 1st time we'd be using Promise.finally outside npm-specific code.

chrisradek (approved):
Overall looks good to me! Just want to make sure using `Promise.finally` will be safe! Alternatively could use `await` with try/finally since typescript will generate that code - but I do think what you have looks cleaner.

packages/browser/src/core/task/task-group.ts
zikaari: I didn't find any polyfilling at all anywhere. Tried to follow our build pipeline till the end, nothing, and no reference to 'core-js` either.

chrisradek: We only polyfill for the CDN version but we load a file here depending on if we determine we should polyfill - https://github.com/segmentio/analytics-next/blob/aa3c901acd77597aa3c9bcb992369a39ef860919/packages/browser/src/browser/standalone.ts#L57-L63

zikaari: The polyfill works and makes `finally` work as expected. Tested in IE11 which has no Promise support at all, but once the polyfill script is executed in the console, Promise and `finally` work nominally.

chrisradek: Sweet, thanks!

packages/browser/src/lib/is-thenable.ts
```diff
@@ -2,7 +2,7 @@
  *  Check if  thenable
  *  (instanceof Promise doesn't respect realms)
  */
-export const isThenable = (value: unknown): boolean =>
+export const isThenable = (value: unknown): value is Promise<unknown> =>
```
silesky, line 5: Unfortunately, we don’t actually know if this is a Promise from this —- we know it’s a thenable. PromiseLike<unknown> might be closer, but I vaguely remember some issue  that stopped me from making this a type guard.

zikaari: What do you recommend? Revert this change altogether?

silesky: I might just create another function called isPromise that does the same thing you want, but also checks for finally and catch

zikaari: I'll revert it.

zikaari:
```suggestion
export const isThenable = (value: unknown): boolean =>
```

zikaari: Although can someone explain what does this mean "instance of Promise doesn't support realms"?

cc @chrisradek

PR #559, opened 2022-07-28
Add analytics-react package

zikaari:
@silesky Please review the package.json with extra caution to make sure things are as they should be

### DIFF
--- a/examples/standalone-playground/package.json
+++ b/examples/standalone-playground/package.json
@@ -8,7 +8,7 @@
     "dev": "yarn concurrently 'yarn run -T watch --filter=standalone-playground' 'sleep 10 && npx http-server .'"
   },
   "dependencies": {
-    "@segment/analytics-next": "workspace:^"
+    "@segment/analytics-next": "1.41.1"
   },
   "devDependencies": {
     "concurrently": "^7.2.1"
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,0 +1,45 @@
+{
+  "name": "@segment/analytics-react",
+  "version": "0.0.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/segmentio/analytics-next",
+    "directory": "packages/react"
+  },
+  "license": "MIT",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/pkg/index.js",
+  "types": "./dist/types/index.d.ts",
+  "files": [
+    "dist/"
+  ],
+  "sideEffects": false,
+  "scripts": {
+    "test": "yarn jest",
+    "lint": "yarn concurrently 'yarn:eslint .' 'yarn:tsc --noEmit'",
+    "build": "rm -rf dist && yarn concurrently 'yarn:build:*'",
+    "build:esm": "yarn tsc -p tsconfig.build.json",
+    "build:cjs": "yarn tsc -p tsconfig.build.json --outDir ./dist/cjs --module commonjs",
+    "watch": "yarn concurrently 'yarn:build:cjs --watch' 'yarn:build:esm --watch'",
+    "watch:test": "yarn test --watch",
+    "tsc": "yarn run -T tsc",
+    "eslint": "yarn run -T eslint",
+    "concurrently": "yarn run -T concurrently",
+    "jest": "yarn run -T jest"
+  },
+  "packageManager": "yarn@3.2.1",
+  "dependencies": {
+    "tslib": "^2.4.0"
+  },
+  "peerDependencies": {
+    "@segment/analytics-next": "^1.40.0",
+    "react": ">=16.8.0"
+  },
+  "devDependencies": {
+    "@segment/analytics-next": "^1.40.0",
+    "@testing-library/react": "^13.3.0",
+    "@types/react": "16.8.0",
+    "react": "18.0.0",
+    "react-dom": "18.0.0"
+  }
+}
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -36,7 +36,6 @@
     "react": ">=16.8.0"
   },
   "devDependencies": {
-    "@segment/analytics-next": "^1.40.0",
     "@testing-library/react": "^13.3.0",
     "@types/react": "16.8.0",
     "react": "18.0.0",
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -1,0 +1,47 @@
+import React, { createContext, useContext } from 'react'
+import {
+  AnalyticsBrowser,
+  AnalyticsBrowserSettings,
+  InitOptions,
+} from '@segment/analytics-next'
+
+export const createClient = (
+  settings: AnalyticsBrowserSettings,
+  options?: InitOptions
+) => AnalyticsBrowser.load(settings, options)
+
+const AnalyticsContext = createContext<AnalyticsBrowser | null>(null)
+
+export type AnalyticsProviderProps = {
+  client: AnalyticsBrowser
+  children?: React.ReactNode
+}
+
+export const AnalyticsProvider = ({
+  client,
+  children,
+}: AnalyticsProviderProps) => {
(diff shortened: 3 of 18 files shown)


silesky:
Pushed a commit that fixes constraints.

packages/react/src/index.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { createContext, useContext } from 'react'
+import {
+  AnalyticsBrowser,
+  AnalyticsBrowserSettings,
+  InitOptions,
+} from '@segment/analytics-next'
+
+export const createClient = (
+  settings: AnalyticsBrowserSettings,
```
silesky, line 9: Why the decision a createClient function wrapper versus just exporting AnalyticsBrowser (I’ve seen this re-export done in some other react SDKs such as Anplitude’s, and it seemed logical).

zikaari: I was going for maximum compliance with our React Native library as seen here - https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/#usage

> ...my inclination would have that (and options as well) as props on “Provider”, instead of making a user always handle client creation themself

"Externalizing" the segment client and then passing it in has benefits. First is that it allows the "Root" component to access it (the one that is wrapping `AnalyticsProvider`), second it allows for setting the client via a factory function, like I'm doing in the Inspector. Similar to `createStore` from Redux.

silesky: > First is that it allows the "Root" component to access it (the one that is wrapping AnalyticsProvider),

In this rare case they could create their own intermediate component or instantiate the client as an escape valve — per the examples.

> second it allows for setting the client via a factory function, like I'm doing in the Inspector. Similar to createStore from Redux.

Not sure I see the benefit — I don’t  think comparing redux is a good point of comparison — in redux, you aren’t passing the store down — you’re passing “dispatch” — the store is essentially treated like a configuration object. I would compare this to other react wrappers and similar SDKs.

zikaari: The API compliance with our React Native library and factory functions are still a stronghold I believe.

```ts
const getSegmentClient = () => {
  const client = createClient({ ... })
  // setup middlewares
  // setup plugins
  // pre-identify the user

  return client
}
```

cc @chrisradek @pooyaj - You're welcome to shed your opinions on this subject

chrisradek: My 2 cents: I think there's value in being consistent with our patterns with regards to the react-native library and in being able to pass a client directly to the provider.

As far as whether it should be required - my intuition is that it makes sense to support passing in just the `writeKey` instead of a client - anecdotally this is what I ended up doing in my own provider. That said, I think this PR is meant to get the code checked in for the published RC library. I think it's worth reaching out to the mobile team to see if they've gotten any feedback on supporting just writeKey and create an issue asking for feedback.

examples/standalone-playground/package.json
```diff
@@ -8,7 +8,7 @@
     "dev": "yarn concurrently 'yarn run -T watch --filter=standalone-playground' 'sleep 10 && npx http-server .'"
   },
   "dependencies": {
-    "@segment/analytics-next": "workspace:^"
+    "@segment/analytics-next": "1.41.1"
```
chrisradek, line 11: How come these were changed from `workspace`? Should they be pinned to the major version? `^1.41.1`?

packages/react/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@segment/analytics-react",
+  "version": "0.0.0",
```
chrisradek, line 3: I think this should match what's in npm, so `1.0.0-rc`.

packages/react/src/index.tsx
silesky: Not sure I see the benefit of copying the react native API, especially if it makes things more complicated? Users aren't going to be running our code in a react-native environment, and  there are significant differences:
- React native doesn't use our code at all, `getSegmentClient` is not a wrapper around AnalyticsBrowser.load -- it's a total replacement. If we wanted to ape their API, we'd refactor `analytics-next` to use that.
- React native analytics doesn't use peer dependencies at all.
- React native will never be run on the server, so instantiating in the global scope is always going to make sense.

examples/standalone-playground/package.json
silesky: Yes, they should be pinned! The reason is to a fix yarn constraints issue where all the packages should depend on the same version, but in my research it's much more common to use a pinned version (see storybook monorepo, etc) -- don't worry, development is unaffected -- it still hardlinks to the workspace version!

packages/react/package.json
silesky: I might hold off on that.  When this gets published (1.0) as part of changeset, it will auto bump to 1.0 -- I don't 100% know what will happen if we add "rc" =).  I haven't looked into the prerelease workflow. https://github.com/changesets/changesets/blob/main/docs/prereleases.md

packages/react/src/index.tsx
silesky: > as far as whether it should be required - my intuition is that it makes sense to support passing in just the writeKey instead of a client -
> That said, I think this PR is meant to get the code checked in for the published RC library.

If so, we will want to hold off on the docs change and the changeset.

packages/react/package.json
chrisradek: @silesky did some digging into how changeset will version prereleases - I think for now we should mark this as `"private": true` to prevent changeset from accidentally publishing it, and then we can manually release it as we need until we mark it GA.

packages/react/src/index.tsx
zikaari: > Not sure I see the benefit of copying the react native API, especially if it makes things more complicated? Users aren't going to be running our code in a react-native environment, and there are significant differences

It doesn't make things more complicated. It is less code for us to maintain, and better for the end-users as they have an opportunity to augment the client if need be. When there are less disparities between web and mobile libraries, it is better for our customers as their on-boarding experience will be much smoother and quicker, when they are going from web to react-native or vice versa. It'll be second nature.

> React native doesn't use our code at all, `getSegmentClient` is not a wrapper around `AnalyticsBrowser.load` -- it's a total replacement. If we wanted to ape their API, we'd refactor analytics-next to use that.

`getSegmentClient` was a demo of how `createClient` can be augmented further before exposing it to the app through `AnalyticsProvider`. The public API of React Native lib is `createClient` and `AnalyticsProvider`, and so is now in case of React library.

> React native analytics doesn't use peer dependencies at all

We can very well make analytics-react be like this as well. It's just a matter of choice. Plus installation procedure compliance is not the goal here, API compliance is.

> React native will never be run on the server, so instantiating in the global scope is always going to make sense.

Please elaborate

> Ultimately, I think we should take inspiration from other libraries that are more similar to ours

But that'd mean straying away from the consistency of our own "sibling" libraries. I'm not opposing your suggestions, I'm just striving for maximum parity. I would be happy to involve someone who manages react-native library, and see their outlook on making `AnalyticsProvider` "direct" by just passing in `writeKey`.

packages/react/package.json
zikaari: Do we need to wait for something else before this goes into GA?

packages/react/src/index.tsx
pooyaj: Not strongly opinionated but my two cents on consistency between libs: don't mind that much about consistency over correctness here. They are two different TS based libraries with different architecture, use-cases, etc. So trying to make the interface consistent doesn't add value if it makes using `analytics-react` harder/confusing. We shouldn't be bound to how React Native lib was built originally. 

AFAIK, the reason the React Native library supports passing in a client is that the library supports running multiple instances of the analytics. So you can have multiple providers to track across multiple write-keys. We don't support that at the moment in AJS, so I don't see someone creating multiple clients and use multiple providers. I'm definitely +1 on having the option to pass in the client ( for reasons Neek explained + maybe someone wants to use React library but load AJS through snippet 🤔 ) but whether to make it the only way te use the react library is something to think about.

PR #561, opened 2022-08-01
Add new buffered methods, improve types

silesky:
This PR:
- Adds 'screen', 'register', 'deregister', 'user' method and 'VERSION' property on AnalyticsBrowser.   🎸 
- Allows buffering of 'screen', 'register', 'deregister' methods for snippet users. 🎹 
- Improves typing -- e.g better names, add overloads 🎷 
> Outstanding todo: update the _actual_ UI snippet (the one that shows up on app.segment.com?) so that the copypasta from includes  'screen', 'register', and 'deregister' ⛑️ 

😢 Because of some differences in return types between AnalyticsBrowser and Analytics (AnalyticsBrowser always being async, while Analytics is inconsistent in that regard) the dream of a unified AnalyticsCore interface between Analytics and AnalyticsBrowser is on hold. 

It’s IMO advisable to not try to obsess over strict sharing of a public interfaces, as that common abstraction does not seem possible, and it can in addition become  hard to maintain either for the legacy behavior of one system or due to subtle differences between platforms. Ultimately, our API is just a bags of methods and we want to evolve and be nimble, and not hamstring ourselves.

We can continue to share models and types where appropriate.

### DIFF
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -10,11 +10,18 @@ import {
   UserParams,
 } from '../arguments-resolver'
 import type { FormArgs, LinkArgs } from '../auto-track'
-import { Callback, invokeCallback } from '../callback'
+import { invokeCallback } from '../callback'
 import { isOffline } from '../connection'
 import { Context } from '../context'
 import { Emitter } from '@segment/analytics-core'
-import { EventFactory, Integrations, Plan, SegmentEvent } from '../events'
+import {
+  Callback,
+  EventFactory,
+  Integrations,
+  Plan,
+  EventProperties,
+  SegmentEvent,
+} from '../events'
 import { Plugin } from '../plugin'
 import { EventQueue } from '../queue/event-queue'
 import { CookieOptions, Group, ID, User, UserOptions } from '../user'
@@ -30,6 +37,7 @@ import { version } from '../../generated/version'
 import { PriorityQueue } from '../../lib/priority-queue'
 import { getGlobal } from '../../lib/get-global'
 import { inspectorHost } from '../inspector'
+import { AnalyticsClassic, AnalyticsSnippetCore } from './interfaces'
 
 const deprecationWarning =
   'This is being deprecated and will be not be available in future releases of Analytics JS'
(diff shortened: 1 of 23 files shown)


packages/browser/src/core/events/interfaces.ts
```diff
@@ -6,6 +7,10 @@ export type JSONValue = JSONPrimitive | JSONObject | JSONArray
 export type JSONObject = { [member: string]: JSONValue }
 export type JSONArray = Array<JSONValue>
 
+// TODO: the change is that context is no longer "undefined". "Callback" was defined twice in this codebase previously, with that being the only difference.
+// Are there methods where the context object for Callback is undefined? If so, which ones?
```
pooyaj, line 11: 🤔 can't think of any. Seems to be only used in the `dispatch` and ctx is always defined.

pooyaj (approved):

packages/browser/src/core/analytics/index.ts
```diff
@@ -503,58 +516,20 @@ export class Analytics extends Emitter {
...
-
-  // This function is only used to add GA and Appcue, but these are already being added to Integrations by AJSN
-  addIntegration() {
-    console.warn(deprecationWarning)
-    return
-  }
-
-  add() {
+  /* analytics-classic stubs */
+  stub(this: never) {
```
chrisradek, line 520: Does this need to be a public method on the class? Seems like this can be a standalone function that we assign to the deprecated methods.

packages/browser/src/core/events/interfaces.ts
chrisradek: Agreed, I can't think of when `ctx` would be undefined.

packages/browser/src/core/events/interfaces.ts
```diff
@@ -106,13 +116,9 @@ export interface SegmentEvent {
   category?: string
   name?: string
 
-  properties?: object & {
-    [k: string]: JSONValue
-  }
+  properties?: EventProperties
```
chrisradek, line 119: This might be too strict - we have some users passing in an instance of a class as event properties and TypeScript will complain that the class doesn't have the index signature.

chrisradek (commented):
Love seeing the addition of more methods on the buffered Analytics class! Had some pretty minor comments.

packages/browser/src/core/events/interfaces.ts
silesky: I don’t think the typescript types should support people passing “any class”, since it’s not clear how distinguish between a Function and an anonymous class. This is not a breaking change for anyone but potentially typescript users who are doing things I don’t think we should support on the first place.

I think we should write the typescript  API to encourage people to do the right way, i.e do something like {…new MyClass()} (if they must) or have a toJSON() or manually construct an object literal.

chrisradek (approved):

PR #567, opened 2022-08-04
[Bugfix] Do not allow the user method to change its return types over its lifecycle

silesky:
I discovered a bug in the recently added "user" method for analytics buffered. It's the only method that's both synchronous and non-chained, so the bug didn't reveal itself previously.

Do not allow the "user" method to change its return types over its lifecycle. We should always return a promise for wrapped methods in AnalyticsBrowser, regardless if the underlying Analytics method is sync or async.


chrisradek (approved):
:shipit:

PR #573, opened 2022-08-09
Improved error message logging in console

arielsilvestri:
Adds a console error with the error status text from the API explaining to users what went wrong with the `settings` call.

Screen Shot 2022-08-03 at 12 57 19 PM
.

### DIFF
--- a/packages/browser/src/browser/__tests__/analytics-pre-init.integration.test.ts
+++ b/packages/browser/src/browser/__tests__/analytics-pre-init.integration.test.ts
@@ -5,7 +5,6 @@ import { Context } from '../../core/context'
 import * as Factory from '../../test-helpers/factories'
 import { sleep } from '../../test-helpers/sleep'
 import { setGlobalCDNUrl } from '../../lib/parse-cdn'
-import { User } from '../../core/user'
 
 jest.mock('unfetch')
 
@@ -60,38 +59,19 @@ describe('Pre-initialization', () => {
       expect(trackSpy).toBeCalledTimes(1)
     })
 
-    test('"return types should not change over the lifecycle for async methods', async () => {
+    test('"return types should not change over the lifecycle for ordinary methods', async () => {
       const ajsBrowser = AnalyticsBrowser.load({ writeKey })
 
       const trackCtxPromise1 = ajsBrowser.track('foo', { name: 'john' })
       expect(trackCtxPromise1).toBeInstanceOf(Promise)
-      await ajsBrowser
+      const ctx1 = await trackCtxPromise1
+      expect(ctx1).toBeInstanceOf(Context)
 
       // loaded
       const trackCtxPromise2 = ajsBrowser.track('foo', { name: 'john' })
       expect(trackCtxPromise2).toBeInstanceOf(Promise)
-
-      expect(await trackCtxPromise1).toBeInstanceOf(Context)
-      expect(await trackCtxPromise2).toBeInstanceOf(Context)
-    })
-
-    test('return types should not change over lifecycle for sync methods', async () => {
-      const ajsBrowser = AnalyticsBrowser.load({ writeKey })
-      const user = ajsBrowser.user()
-      expect(user).toBeInstanceOf(Promise)
-      await ajsBrowser
-
-      // loaded
-      const user2 = ajsBrowser.user()
-      expect(user2).toBeInstanceOf(Promise)
-
-      expect(await user).toBeInstanceOf(User)
-      expect(await user2).toBeInstanceOf(User)
-    })
-
-    test('version should return version', async () => {
-      const ajsBrowser = AnalyticsBrowser.load({ writeKey })
-      expect(typeof ajsBrowser.VERSION).toBe('string')
+      const ctx2 = await trackCtxPromise2
+      expect(ctx2).toBeInstanceOf(Context)
     })
 
     test('If a user sends multiple events, all of those event gets flushed', async () => {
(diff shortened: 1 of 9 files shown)


packages/browser/src/browser/index.ts
```diff
@@ -83,9 +83,16 @@ export function loadLegacySettings(
   const baseUrl = cdnURL ?? getCDN()
 
   return fetch(`${baseUrl}/v1/projects/${writeKey}/settings`)
-    .then((res) => res.json())
+    .then((res) => {
```
silesky, line 86: Looks good! Can we get some client tests?

packages/browser/src/browser/index.ts
```diff
@@ -83,9 +83,16 @@ export function loadLegacySettings(
...
+      if (!res.ok) {
+        return res.text().then((statusText) => {
```
silesky, line 88: Hey, maybe I'm missing something, but why do we need text() here?


```
void fetch('https://httpstat.us/300')
  .then((res) => {
    if (!res.ok) {
      throw res
    }
  })
  .catch((e) => {
    console.error(e.statusText)
    throw err
 })
```

silesky:
Hey, you may need to update all the tests that use mock fetch to use more robust HTTP mock, since it's checking for `OK`
e.g. https://github.com/segmentio/analytics-next/pull/573#discussion_r941597513
image

This is why using nock is in general so much better than mocking fetch =S

packages/browser/src/browser/index.ts
arielsilvestri: Just throwing didn't give us the status text iirc, but I'll look again very quickly.

silesky: Works in my REPL
image

arielsilvestri: Confirmed, needed to unwrap the `text` in order to get the statusText.

silesky: hmmm... `statusText` is part of the [response specification](https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText)

packages/browser/src/browser/__tests__/analytics-pre-init.integration.test.ts
```diff
@@ -60,38 +59,19 @@ describe('Pre-initialization', () => {
...
-      expect(user2).toBeInstanceOf(Promise)
-
-      expect(await user).toBeInstanceOf(User)
-      expect(await user2).toBeInstanceOf(User)
-    })
-
-    test('version should return version', async () => {
-      const ajsBrowser = AnalyticsBrowser.load({ writeKey })
-      expect(typeof ajsBrowser.VERSION).toBe('string')
+      const ctx2 = await trackCtxPromise2
```
silesky, line 73: hmmm, should this change be in here?

arielsilvestri: Seems like it.

packages/browser/src/browser/index.ts
```diff
@@ -83,9 +83,16 @@ export function loadLegacySettings(
   const baseUrl = cdnURL ?? getCDN()
 
   return fetch(`${baseUrl}/v1/projects/${writeKey}/settings`)
-    .then((res) => res.json())
+    .then((res) => {
+      if (!res.ok) {
+        return res.text().then((errorResponseMessage) => {
+          throw { ...res, errorResponseMessage }
```
chrisradek, line 89: We have metrics that keep track of errors when the library is loading, but it currently only works if the thing thrown is an error.

What do you think of throwing an error instead with the errorResponseMessage as the error.message?

packages/browser/src/browser/index.ts
```diff
@@ -83,9 +83,16 @@ export function loadLegacySettings(
   const baseUrl = cdnURL ?? getCDN()
 
   return fetch(`${baseUrl}/v1/projects/${writeKey}/settings`)
-    .then((res) => res.json())
+    .then((res) => {
+      if (!res.ok) {
+        return res.text().then((errorResponseMessage) => {
```
chrisradek, line 88: What does the `text()` end up looking like? In your screenshot it looks like the `statusText` has a useful error message - I wonder if it makes sense to use that instead.

silesky: I'll chip in because this is something I discussed offline -- that screenshot is obsolete. That text() returns the more detailed message -- it's in the response body. Now the object will be something like:
```
errorResponseMessage: "Cannot GET -- invalid writekey ..."
status: 404
statusText: "Not Found"
```

chrisradek: Oh awesome, keep `.text()` then :smile:

arielsilvestri: So that `statusText` is misleading - it wasn't actually the `statusText` that you'd normally get assigned, but rather the one I assigned to `statusText`. It was empty without unwrapping `res.text`.

packages/browser/src/browser/index.ts
silesky: @cradek Really, the library only catches errors if the thing thing thrown passes an "instanceof Error" check? Seems dicyyy

arielsilvestri: So just a quick update with: `throw {...res, message: errorResponseMessage }`? Works for me

chrisradek: > @cradek Really, the library only reports errors if the thing thing thrown passes an "instanceof Error" check? Seems dicyyyyyy

Yeah - no we should definitely update that. We're also severely limited to how much text we can report (just learned about this last week), so we can't simply JSON.stringify whatever gets thrown and report it. 😢

arielsilvestri: I also like how it looks better as well, great call!

chrisradek (approved):

PR #577, opened 2022-08-10
support secure cookie option for identity cookies

chrisradek:
This PR fixes an issue where the `secure` cookie property wasn't being applied to cookies created by the `User` and `Group` classes. This meant it was always being ignore (effectively false)


pooyaj (approved):
:shipit:

PR #582, opened 2022-08-11
improve error message when sending metrics fails

chrisradek:
This PR updates the error message logged when sending client-side performance metrics to segment fails.
See https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/troubleshooting/#why-do-i-see-a-network-request-to-m for background info on client-side performance metrics.

These metrics do not impact customer event delivery, but can cause confusion when showing up on platforms like Sentry when an error occurs. This provides a way to filter out these messages by the new error message preceding the error.

### DIFF
--- a/.changeset/moody-rats-double.md
+++ b/.changeset/moody-rats-double.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': patch
+---
+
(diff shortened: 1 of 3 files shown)


silesky (approved):
Nice!

PR #587, opened 2022-08-24
Create rc of analytics node

silesky:
Sorry for this huge PR

Changelog:
- copy / refactor logic from core and use inside of nodejs
- adds plugin-validation package
- adds internal config package (for jest stuff and lint-staged)
- adds core-integration-test package
- adds better test DX experience

Fast follow:
- Add node tests / example that use express 
- refactor other repos to use internal config and incremental building
- Update browser to use the analytics-core interfaces
  - Refactor analytics to use / extend Core classes and interfaces -- register segment inspector via event emitter.
   - Node tests that rely on stats can be refactored to use event emitter instead.
- Refactor CI to acknowledge additional packages (will use turbo this time).


packages/core/src/priority-queue/persisted.ts
```diff
@@ -0,0 +1,136 @@
...
+  loc.setItem(
+    key,
+    JSON.stringify({
+      ...stored,
+      ...memory,
+    })
+  )
+}
+
+function remove(loc: Storage, key: string): void {
```
silesky, line 55: Not necessary to be in core (the tests weren’t as brittle as I thought), but I refactored this to no longer create any import side effects.

chrisradek:
Started going through this PR, wanted to know about this point:
> copy / refactor logic from core and use inside of nodejs

Can you share what was refactored vs what was just copied? That'll just help with knowing what to focus extra attention on.

packages/core/src/priority-queue/persisted.ts
```diff
@@ -0,0 +1,136 @@
+import { PriorityQueue } from '.'
```
silesky, line 1: File refactored to have no side effects

packages/core/src/plugins/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  priority: 'critical' | 'non-critical' // whether AJS should expect this plugin to be loaded before starting event delivery
+}
+
+export type PluginType =
+  | 'before'
+  | 'after'
+  | 'destination'
+  | 'enrichment'
+  | 'utility'
+
```
silesky, line 16: This file is basically ally unchanged, other than generics

packages/core/src/events/interfaces.ts
```diff
@@ -0,0 +1,210 @@
...
+
+export type Traits = Record<string, any>
+
+export type EventProperties = Record<string, any>
+
+export type Integrations = {
+  All?: boolean
+  [integration: string]: boolean | JSONObject | undefined
+}
+
```
silesky, line 27: Other than the word "core", materially unchanged

packages/core/src/context/index.ts
```diff
@@ -0,0 +1,125 @@
+import { CoreSegmentEvent } from '../events/interfaces'
+
+import { v4 as uuid } from '@lukeed/uuid'
+import { dset } from 'dset'
+import { CoreLogger, LogLevel, LogMessage } from '../logger'
+import Stats, { Metric } from '../stats'
+
+export function isContext(v?: object): v is CoreContext {
```
silesky, line 8: This function is new to allow duck typing as well as instanceOf -- I thought it would be good to add the edge case, but it's not critical

packages/core/src/connection/__tests__/index.test.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * @jest-environment jsdom
+ */
```
silesky, line 3: Declaring tests using this magic comment allows us to easily identify where our dom dependencies are

packages/core/src/arguments-resolver/page.ts
```diff
@@ -0,0 +1,57 @@
+import { isFunction, isPlainObject, isString } from '../validation'
+import { JSONObject, EventProperties, CoreOptions } from '../events'
+import { Callback } from '../events/interfaces'
+
```
silesky, line 4: Basically unchanged, except for moving the arguments resolver to its own "page" file -- it's the only resolver we need to bring in (because of the overloads)

internal/config/src/jest/index.js
```diff
@@ -0,0 +1,18 @@
+const getPackages = require('get-monorepo-packages')
+
```
silesky, line 2: This module allows  ts-jest to pick up changes in other packages without having to run build on every external change.

packages/core/src/context/index.ts
```diff
@@ -0,0 +1,125 @@
...
+  reason?: string
+
+  constructor(options: CancelationOptions) {
+    this.retry = options.retry ?? true
+    this.type = options.type ?? 'plugin Error'
+    this.reason = options.reason ?? ''
+  }
+}
+
+export class CoreContext<Event extends CoreSegmentEvent = CoreSegmentEvent> {
```
silesky, line 41: refactored from the browser Context class to make this new "base" class (meant to be used by node and browser) -- adds a generic, improves decoupling by taking Stats and Logger via constructor injection, removes extraneous access modifiers  like the word "public" to make a bit prettier.

packages/core/src/emitter/index.ts
```diff
@@ -1,12 +1,12 @@
-export class Emitter {
-  private callbacks: Record<string, Function[]> = {}
+export class Emitter<EventName extends string = string> {
```
silesky, line 1: Change to allow one to specific events

packages/core/src/events/index.ts
```diff
@@ -0,0 +1,251 @@
...
+    })
+  }
+
+  private baseEvent(): Partial<CoreSegmentEvent> {
+    const base: Partial<CoreSegmentEvent> = {
+      integrations: {},
+      options: {},
+    }
+
+    if (!this.user) return base
```
silesky, line 163: the only material change to this file IIRC

packages/core/src/analytics/dispatch.ts
```diff
@@ -0,0 +1,51 @@
...
+  timeout?: number
+  debug?: boolean
+  callback?: Callback
+  retryQueue?: boolean
+}
+
+export async function dispatch(
+  event: CoreSegmentEvent,
+  queue: EventQueue,
+  emitter: Emitter,
```
silesky, line 18: This argument is new -- will allow us to decouple this function from listeners like segment inspector

silesky:
> Started going through this PR, wanted to know about this point:
> 
> > copy / refactor logic from core and use inside of nodejs
> 
> Can you share what was refactored vs what was just copied? That'll just help with knowing what to focus extra attention on.

I skimmed and added comments to what I think are the relevant files!

packages/core/src/queue/event-queue.ts
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  private async deliver(ctx: CoreContext): Promise<CoreContext> {
+    await this.criticalTasks.done()
+
+    const start = Date.now()
+    try {
+      ctx = await this.flushOne(ctx)
+      const done = Date.now() - start
+      this.emit('delivery_success', done)
```
silesky, line 182: new emitter -- I could see us moving the stats stuff outside of this function altogether and just regally on emitter events

packages/core/src/queue/event-queue.ts
```diff
@@ -0,0 +1,319 @@
...
+        const attempts = destinations.map((destination) =>
+          attempt(ctx, destination)
+        )
+        Promise.all(attempts).then(resolve).catch(reject)
+      }, 0)
+    })
+
+    this.emit('message_delivered', ctx)
+    ctx.stats?.increment('message_delivered')
+    // should emit to: TODO: inspectorHost.delivered?.(ctx as any, ['segment.io']) -- can we move all this inspector stuff to a plugin that listens to these events?
```
silesky, line 312: don't want inspectorHost in core -- it can use the event emitter to listen to events

packages/node/src/app/analytics-node.ts
```diff
@@ -0,0 +1,345 @@
...
+    options: InitOptions,
+    queue: EventQueue
+  ) {
+    super()
+    this.settings = settings
+    this.eventFactory = new EventFactory()
+    this.integrations = options?.integrations ?? {}
+    this.options = options ?? {}
+    this.queue = queue
+    bindAll(this)
```
silesky, line 69: new

packages/node/src/app/analytics-node.ts
```diff
@@ -0,0 +1,345 @@
...
+   * Registers one or more plugins to augment Analytics functionality.
+   * @param plugins
+   */
+  async register(...plugins: CorePlugin<any, any>[]): Promise<ContextNode> {
+    const ctx = ContextNode.system()
+
+    const registrations = plugins.map((xt) =>
+      this.queue.register(ctx, xt, this)
+    )
+    await Promise.all(registrations)
```
silesky, line 320: we might want to swallow these errors, but left as is

packages/core-integration-tests/package.json
```diff
@@ -0,0 +1,20 @@
+{
```
silesky, line 1: new package!

package.json
```diff
@@ -4,31 +4,31 @@
...
     "postinstall": "husky install",
     "changeset": "changeset",
     "update-versions-and-changelogs": "changeset version && yarn version-run-all && bash scripts/update-lockfile.sh",
-    "release": "yarn build:packages --force && changeset publish && yarn postpublish-run-all && git push origin --tags --no-verify",
+    "release": "yarn build --force && changeset publish && yarn postpublish-run-all && git push origin --tags --no-verify",
     "postpublish-run-all": "yarn workspaces foreach -vpt --no-private run postpublish",
     "version-run-all": "yarn workspaces foreach -vpt --no-private run version",
     "core": "yarn workspace @segment/analytics-core",
-    "core+deps": "turbo run --filter='analytics-core'",
+    "core+deps": "turbo run --filter=@segment/analytics-core...",
```
silesky, line 27: the "..." syntax is how turborepo expresses "include all dependencies".

packages/core/jest.config.js
```diff
@@ -6,6 +6,11 @@ module.exports = {
   testEnvironment: 'node',
   testMatch: ["**/?(*.)+(test).[jt]s?(x)"],
   clearMocks: true,
+  setupFilesAfterEnv: ['./jest.setup.js'],
+  projects: [
+    "<rootDir>",
+    "<rootDir>/../core-integration-tests"
```
silesky, line 12: running "yarn test" and so forth in core automatically runs integration tests (yes, jest has its own support for monorepos!)

packages/core/src/callback/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  return (
+    sleep(delayTimeout)
+      // pTimeout ensures that the callback can't cause the context to hang
+      .then(() => pTimeout(cb(), timeout ?? 1000))
+      .catch((err) => {
+        ctx?.log('warn', 'Callback Error', { error: err })
+        // ctx?.stats?.increment('callback_error')
```
chrisradek, line 48: Did you mean to comment this out?

packages/node/src/app/analytics-node.ts
```diff
@@ -0,0 +1,345 @@
...
+
+export type SegmentEventOptionsNode = CoreOptions & Identity
+
+export type Identity =
+  | { userId: string; anonymousId?: string }
+  | { userId?: string; anonymousId: string }
+
+type SegmentEventTypeNode = 'track' | 'page' | 'identify' | 'alias' | 'screen'
+
+export interface SegmentEventNode extends CoreSegmentEvent {
```
chrisradek, line 33: Minor: `NodeSegmentEvent` feels like a more natural extension of `CoreSegmentEvent` to me since then we're always just modifying the prefix.

packages/core/src/analytics/dispatch.ts
```diff
@@ -0,0 +1,51 @@
...
+  options?: DispatchOptions
+): Promise<CoreContext> {
+  const ctx = new CoreContext(event)
+  emitter.emit('message_dispatch_pending', ctx) // TODO: inspectorHost.triggered?.(ctx as any)
```
chrisradek, line 22: /cc @zikaari since this would end up impacting inspector work (though this PR doesn't change anything for AnalyticsBrowser yet)

packages/core/src/analytics/dispatch.ts
chrisradek: Does it make sense for the emitter to ever _not_ be analytics? Wonder if naming the argument something like `analytics` would make it more clear what should be passed/what's emitting events.

packages/core/src/analytics/index.ts
```diff
@@ -0,0 +1,12 @@
+import { CoreContext } from '../context'
+
+export interface CoreAnalytics {
+  track(...args: unknown[]): Promise<CoreContext>
```
chrisradek, line 4: Is `group` intentionally missing?

packages/core/src/queue/event-queue.ts
chrisradek: /cc @zikaari

packages/core/src/queue/event-queue.ts
```diff
@@ -0,0 +1,319 @@
...
+    try {
+      ctx = await this.flushOne(ctx)
+      const done = Date.now() - start
+      this.emit('delivery_success', done)
+      ctx.stats?.gauge('delivered', done)
+      ctx.log('debug', 'Delivered', ctx.event)
+      return ctx
+    } catch (err) {
+      ctx.log('error', 'Failed to deliver', err as object)
+      this.emit('delivery_failure') //     TODO: ctx.stats.increment('delivery_failed')
```
chrisradek, line 188: Is that TODO meant to be there?

I've only seen `delivery_failed` occur multiple times if a `before` plugin (which would include sourceMiddleware as well) failed but didn't disable retries.

packages/core/src/queue/event-queue.ts
chrisradek: Sounds like a `utility` plugin - /cc @zikaari

packages/core/src/analytics/dispatch.ts
```diff
@@ -0,0 +1,51 @@
...
+  const startTime = Date.now()
+  let dispatched: CoreContext
+  if (queue.isEmpty()) {
+    dispatched = await queue.dispatchSingle(ctx)
+  } else {
+    dispatched = await queue.dispatch(ctx)
+  }
+  const elapsedTime = Date.now() - startTime
+  const timeoutInMs = options?.timeout
+
```
chrisradek, line 37: Note: This might be a good spot to remove the context from the priorityQueue's `seen` map. That can be done in a follow-up PR (I can create an issue for that unless you already fixed that memory leak) - basically want to make sure we don't let that map just grow especially in node.js.

packages/node/src/index.ts
```diff
@@ -1,5 +1,6 @@
-import { foo } from './example'
-// TODO
-
-console.log('hello world')
-foo(123)
+export { load } from './app/load'
```
chrisradek, line 1: Can't comment in the example files but those can probably be removed now :smile:

chrisradek (commented):
Overall I think this looks good - thanks for commenting on what was changed vs copied.

constraints.pro
```diff
@@ -26,7 +26,12 @@ gen_enforced_dependency(WorkspaceCwd, DependencyIdent, DependencyRange2, Depende
   \+ member(DependencyIdent, [
     % Allow examples to use different versions of react and
     'react', 'react-dom',
-    '@types/react'
+    '@types/react',
+    % Allow the usage of workspace^ -- there is a better way to do this =)
```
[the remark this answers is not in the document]
silesky: There is very limited documentation on yarn constraints and their API sure, but I'm someone who knew prolog could figure out how to exclude based on  some kind of path regex, but I couldn't figure it out or find any similar "recipes".

package.json
```diff
@@ -4,31 +4,31 @@
...
     "browser": "yarn workspace @segment/analytics-next",
-    "browser+deps": "turbo run --filter='analytics-next'",
+    "browser+deps": "turbo run --filter=@segment/analytics-next...",
     "node": "yarn workspace @segment/analytics-node",
-    "node+deps": "turbo run --filter='analytics-node'",
+    "node+deps": "turbo run --filter=@segment/analytics-node... --filter=!analytics-next",
```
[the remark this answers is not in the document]
silesky: Yes, the validation plug-in has consumes it as a dev dependency for unit testing.

packages/core/src/queue/event-queue.ts
silesky: Oh yes -- my original intention (possibly still) would be to remove stats if favor of a single emitter.

packages/core/src/analytics/dispatch.ts
silesky: I prefer the variable name reflect only the minimal required interface, per "principal of least privelege / interface segration principal", rather than the larger concrete class -- this is  at minimum helpful for unit tests =)

packages/core/src/callback/index.ts
silesky: Oh yes, I can add the stats back (even if not use ATM) -- for context, my original intention was be to remove stats in favor of just using our emitter to collect stats, but it ended up being a little more work than I wanted to do atm.

packages/core/src/analytics/dispatch.ts
chrisradek: To be clear - I was expecting the type to still be `Emitter`, just the name of the property to be `analytics`. Can we have a doc string then so someone looking at the function understands we expect it to be analytics (even if unit tests will pass anything)

silesky: I prefer the name to reflect the type, since if you pass an emitter that's not analytics, it's a bit confusing.

packages/core/src/analytics/dispatch.ts
silesky: I didn't fix the memory leak, but that's a good callout -- we can prune old events.

packages/core/src/analytics/index.ts
silesky: Good catch!

packages/node/src/app/analytics-node.ts
silesky: I've been doing Noun_Suffix (e.g. SegmentEventTypeNode, etc) because AnalyticsNode uses "Noun_Suffix" /shrug

chrisradek: Yeah, you can take or leave my suggestion :smile: I just wondered why it wasn't SegmentEventCore then :D

silesky: I think it was a good one! Updated.

chrisradek (approved):
Thanks for going through the feedback and making updates! Looks good to me!

PR #590, opened 2022-09-02
Fix memory leak

silesky:
> The PriorityQueue class keeps track of any events (context) it sees so it can calculate whether an event can still be retried and how long the retry backoff should be.

> We don’t clear an event’s entry from this map after it has been delivered, so overtime this map can get large.

> Retries typically occur when there’s a network issue. For browsers, the memory leak is hard to detect because the in-memory map is cleared whenever the user navigates to a new page, and the persisted (localStorage) map has storage limits.

> In node.js, an application can run for a long time, so the memory leak is more likely to build up over time.

### DIFF
--- a/packages/core/src/priority-queue/index.ts
+++ b/packages/core/src/priority-queue/index.ts
@@ -1,23 +1,46 @@
 import { Emitter } from '../emitter'
-import { backoff } from './backoff'
+import { backoff, calculateMaxTotalRetryTime } from './backoff'
 
 /**
  * @internal
  */
 export const ON_REMOVE_FROM_FUTURE = 'onRemoveFromFuture'
 
-export interface WithID {
+export interface QueueItem {
   id: string
 }
 
-export class PriorityQueue<T extends WithID = WithID> extends Emitter {
+export type Seen = {
+  [id: string]: {
+    /**
+     * Number of retry attempts.
+     */
+    attempts: number
+    /**
+     *  Epoch time representing when this the current key/value is eligible for deletion.
+     */
+    expiration: number
+  }
+}
+
+const _pruneExpiredFromSeenMap = (seen: Seen) => {
+  // moving private methods outside outside of classes allows minification of those method names
+  const now = Date.now()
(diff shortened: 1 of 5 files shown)


packages/core/src/priority-queue/index.ts
```diff
@@ -86,6 +123,7 @@ export class PriorityQueue<T extends WithID = WithID> extends Emitter {
   }
 
   pop(): T | undefined {
+    this._cleanupSeen()
```
silesky, line 126: <sanity check>

chrisradek:
The `calculateMaxTotalRetryTime` calculation makes sense - it's the sum of the maximum backoff amounts given an allowed number of attempts.

I think this leads to a lower expiration than expected though.

- The expiration doesn't take into account time spent actually processing an event (going through plugins for example)
- The persisted queue will persist any unsent events to localstorage when the user navigates away from the page. When the user navigates to a page in the same domain, those events sent. This is the riskier scenario as it becomes much easier to hit the expiration in this case.

I haven't though this all the way through but I wonder if we can move the attempts data to the `Context`. We already have attempts stored there for the EventQueue - the priority queue is (I think) meant to track attempts per destination (each destination has its own priority queue).

silesky:
> * The expiration doesn't take into account time spent actually processing an event (going through plugins for example)

This is a good point -- we can add a cushion!

> * The persisted queue will persist any unsent events to localstorage when the user navigates away from the page. When the user navigates to a page in the same domain, those events sent. This is the riskier scenario as it becomes much easier to hit the expiration in this case.

Thanks for flagging this! This actually doesn't seem that bad, as it only applies to browsers, and events in the situation will never get lost -- they might only be retried over the maximium, and the chance of this happening a second time is really low. However, to allow resumes, we could store the queue init time in each event (like an ID keep track of which "session" an event  belongs to) and if we ever see events pop up with an old init time,  we update each event with the new time (expiration - oldInitTime + newInitTime) when it gets pushed to the queue .

PR #593, opened 2022-09-09
Add public API

silesky:
1. add criticalTasks to register/deregister, so registration only happens after plugins are loaded2
2. tweak EventEmitter interface (see README.md)
3. return void


packages/core/src/emitter/index.ts
```diff
@@ -1,29 +1,61 @@
...
+ * Event Emitter that takes the expected contract as a generic
+ * @example
+ * ```ts
+ *  type Contract = {
+ *    delivery_success: [DeliverySuccessResponse, Metrics],
+ *    delivery_failure: [DeliveryError]
+ * }
+ *  new Emitter<Contract>()
+ *  .on('delivery_success', (res, metrics) => ...)
+ *  .on('delivery_failure', (err) => ...)
```
chrisradek, line 15: Love that you added an example here - it gets hard to parse the types sometimes :smile:

packages/core/src/events/interfaces.ts
```diff
@@ -26,7 +26,7 @@ export type Integrations = {
 }
 
 // renamed
-export type CoreOptions = {
+export interface CoreOptions {
```
chrisradek, line 29: Just curious - was this change needed or just being consistent?

packages/node/README.md
```diff
@@ -5,22 +5,25 @@ https://segment.com/docs/connections/sources/catalog/libraries/server/node/
...
 
-
+## Basic Usage
 ```ts
-// TODO: finalize API
-import { load } from '@segment/analytics-node'
+// analytics.ts
+import { AnalyticsNode } from '@segment/analytics-node'
 
+export const analytics = new AnalyticsNode({ writeKey: '<MY_WRITE_KEY>' })
```
chrisradek, line 13: 🎉

packages/node/README.md
```diff
@@ -5,22 +5,25 @@ https://segment.com/docs/connections/sources/catalog/libraries/server/node/
...
-router.post('/user',  async (req, res) => {
-  ajsPromise.then((ajs) => ajs.track(user.id, "some event"))
-  res.json(user)
-})
+analytics.identify('Test User', { loggedIn: true }, { userId: "123456" })
+analytics.track('hello world', {}, { userId: "123456" })
 
+```
 
+# Event Emitter (Advanced Usage)
```
chrisradek, line 24: More of a todo, but we should make sure to document how to handle errors. (maybe that won't be apparent to us until the plugin is added)

chrisradek (approved):
I think this looks great!

packages/core/src/events/interfaces.ts
silesky: You mean the rename? I thought it made sense to prefix the things that were going to become base classes / interface (e.g. in node, we extend these options with Identity). However, being totally consistent here is hard.

packages/core/src/emitter/index.ts
silesky: Thanks!

packages/node/README.md
silesky: I created a new `dispatchWithErrorHandling` function that is worth a look -- it emits an 'error' whenever there's a delivery failure due to for example, a validation failure, or an outright rejection (such as a network error)

packages/core/src/analytics/dispatch.ts
```diff
@@ -57,3 +57,21 @@ export async function dispatch(
...
+  try {
+    const ctx = await dispatch(event, queue, emitter, options)
+    if (ctx.failedDelivery()) {
+      throw ctx
+    } else {
+      emitter.emit(event.type, ctx)
+      return ctx
+    }
+  } catch (ctx) {
+    emitter.emit('error', ctx)
```
silesky, line 74: do we want to wrap all our errors in custom objects to give more metadata about their context -- some errors may not have actual context object.

packages/core/src/events/index.ts
```diff
@@ -28,7 +28,7 @@ export class EventFactory {
       ...this.baseEvent(),
       event,
       type: 'track',
-      properties,
+      properties: properties ?? {},
```
pooyaj, line 31: Would this be a breaking change in the eventual `track` playload structure (i.e., after this change track payload will always have a `properties` object regardless )?

packages/node/src/app/analytics-node.ts
```diff
@@ -39,294 +57,263 @@ export interface AnalyticsSettings {
   writeKey: string
   timeout?: number
   plugins?: CorePlugin[]
+  integrations?: Integrations
+  retryQueue?: boolean
```
pooyaj, line 61: 🤔 I think eventually we can get rid of these, since server-side destinations do not have a notion of `integrations`. Also we are not implementing any retry mechanism? are we?

pooyaj (commented):
Couple of comments otherwise ✨

pooyaj (approved):
Nice 👌

packages/core/src/events/interfaces.ts
chrisradek: Oh, I meant why change from `type` to `interface`, but was just curious.

packages/core/src/events/index.ts
```diff
@@ -246,6 +247,7 @@ export class EventFactory {
       messageId,
     }
 
+    validateEvent(evt)
```
chrisradek, line 250: Does doing this validation here rather than as a plugin change how errors are surfaced to the user? (I thought there was some validation already happening outside of the plugin in AnalyticsBrowser so the plugin wouldn't actually catch much anyway)

packages/core/src/events/index.ts
```diff
@@ -28,7 +29,7 @@ export class EventFactory {
       ...this.baseEvent(),
       event,
       type: 'track',
-      properties,
+      properties: properties ?? {}, // TODO: why is this not a shallow copy like everywhere else?
```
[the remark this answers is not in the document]
silesky: The question was more like “why are we not making a shallow clones of properties here, but we do make shallow clones of other objects like traits and options in other methods. Even if potentially pointless, I was wondering if someone just forgot to apply the spread operator here  :-)

packages/node/src/app/analytics-node.ts
```diff
@@ -7,327 +7,298 @@ import {
...
+type IdentityOptions =
   | { userId: string; anonymousId?: string }
   | { userId?: string; anonymousId: string }
 
+/** Events from CoreOptions */
+export interface NodeSegmentEventOptions {
+  context?: NodeContext
+  timestamp?: CoreOptions['timestamp']
+  anonymousId?: CoreOptions['anonymousId']
+  userId?: CoreOptions['userId']
```
chrisradek, line 38: It looks like these are in the top-level object passed to our `track et al` calls. Can we omit these in node so they don't also show up under options?

chrisradek (approved):
Had a comment around userId/anonymousId appearing in top-level fields and options fields, otherwise looks great!

PR #594, opened 2022-09-13
Passing Storage to Plugins

pooyaj:
# RFC - Storage Layer for Plugins 💾  
Problem: If a plugin needs to use browser storage ( cookies, LS, etc. ), it has to implement the storage handling functionality, and Cookie handling in particular requires decent amount of code. Furthermore, plugins aren't aware of AJS storage configuration, so they may use storage regardless of how users configured the AJS. 

In order to enable plugins ( our particular use-case is action destinations ) access the storage layer of AJS, we are creating a universalStorage object during initialization and pass it down to plugins as a dependency when registering plugins. 

Few things of note: 
- [ ] We are using `PluginConfig` object to pass the dependencies to the plugin. Better options? 
- [ ] There is decent amount of overlap between `User` class and `UniversalStorage`, so there is some opportunity for refactoring. 

### DIFF
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -304,8 +304,7 @@ export class Analytics
 
     const registrations = plugins.map((xt) =>
       this.queue.register(ctx, xt, this, {
-        priority: 'critical',
-        options: { storage: this._storage },
+        dependencies: { storage: this._storage },
       })
     )
     await Promise.all(registrations)
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -24,7 +24,14 @@ import {
 } from '../events'
 import { Plugin } from '../plugin'
 import { EventQueue } from '../queue/event-queue'
-import { CookieOptions, Group, ID, User, UserOptions } from '../user'
+import {
+  CookieOptions,
+  Group,
+  ID,
+  UniversalStorage,
+  User,
+  UserOptions,
+} from '../user'
 import autoBind from '../../lib/bind-all'
 import { PersistedPriorityQueue } from '../../lib/priority-queue/persisted'
 import type { LegacyDestination } from '../../plugins/ajs-destination'
@@ -89,6 +96,7 @@ export class Analytics
   protected settings: AnalyticsSettings
   private _user: User
   private _group: Group
+  private _storage: UniversalStorage
   private eventFactory: EventFactory
   private _debug = false
 
@@ -127,6 +135,8 @@ export class Analytics
           : options?.group,
         cookieOptions
       ).load()
+
+    this._storage = new UniversalStorage(disablePersistance, cookieOptions)
     this.eventFactory = new EventFactory(this._user)
     this.integrations = options?.integrations ?? {}
     this.options = options ?? {}
@@ -293,7 +303,10 @@ export class Analytics
     const ctx = Context.system()
 
     const registrations = plugins.map((xt) =>
-      this.queue.register(ctx, xt, this)
+      this.queue.register(ctx, xt, this, {
+        priority: 'critical',
+        options: { storage: this._storage },
+      })
     )
     await Promise.all(registrations)
 
--- a/packages/browser/src/core/queue/event-queue.ts
+++ b/packages/browser/src/core/queue/event-queue.ts
@@ -6,7 +6,7 @@ import { isOnline } from '../connection'
 import { Context, ContextCancelation } from '../context'
 import { Emitter } from '@segment/analytics-core'
 import { Integrations } from '../events'
-import { Plugin } from '../plugin'
+import { Plugin, PluginConfig } from '../plugin'
 import { createTaskGroup, TaskGroup } from '../task/task-group'
 import { attempt, ensure } from './delivery'
 import { inspectorHost } from '../inspector'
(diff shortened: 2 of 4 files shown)


packages/browser/src/core/user/index.ts
```diff
@@ -398,3 +398,53 @@ export class Group extends User {
     return undefined
   }
 }
+
+export class UniversalStorage {
+  private cookies: Store
+  private localStorage: Store
+  private mem: Store
+
+  options: UserOptions = {}
```
silesky, line 407: what does this do here?

packages/browser/src/core/user/index.ts
```diff
@@ -398,3 +398,53 @@ export class Group extends User {
...
+      !isDisabled && Cookie.available()
+        ? new Cookie(cookieOptions)
+        : new NullStorage()
+
+    this.mem = isDisabled ? new NullStorage() : new Store()
+
+    autoBind(this)
+  }
+
+  chainGet<T>(key: string): T | null {
```
silesky, line 425: The word "chain" and "try" seem like implementation details, any reason to not continue to use the conventions of the web storage API (if not outright implementing it?).

packages/browser/src/core/user/index.ts
```diff
@@ -398,3 +398,53 @@ export class Group extends User {
     return undefined
   }
 }
```
silesky, line 400: Is the idea that this utility class can also be shared with the User class -- either injected or through inheritance?

packages/browser/src/core/user/index.ts
```diff
@@ -398,3 +398,53 @@ export class Group extends User {
...
+
+export class UniversalStorage {
+  private cookies: Store
+  private localStorage: Store
```
silesky, line 404: Looking at this file, I feel like it would make sense to create an actual interface and implement it where it makes sense, makes creating testing stubs so much easier and ensuring unity between all these similar storage classes.

packages/browser/src/core/analytics/index.ts
```diff
@@ -127,6 +135,8 @@ export class Analytics
           : options?.group,
         cookieOptions
       ).load()
+
+    this._storage = new UniversalStorage(disablePersistance, cookieOptions)
```
silesky, line 139: I feel like it might be better adherence to the  Interface Seg. Principal to inject this via constructor for max flexibility  -- especially if we allow you to pass any storage (I could see this being a cojfig option)

packages/browser/src/core/user/index.ts
pooyaj: Yes! either of those two. So either User inherits from `UniversalStorage` or we just inject the `_storage` instance to User class. Going to refactor those after we settle of overall approach.

packages/browser/src/core/user/index.ts
pooyaj: Hah, good catch, has no use ( remains of copy 🍝 )

packages/browser/src/core/queue/event-queue.ts
```diff
@@ -43,9 +43,10 @@ export class EventQueue extends Emitter {
   async register(
     ctx: Context,
     plugin: Plugin,
-    instance: Analytics
+    instance: Analytics,
+    pluginConfig?: PluginConfig
   ): Promise<void> {
-    await Promise.resolve(plugin.load(ctx, instance))
+    await Promise.resolve(plugin.load(ctx, instance, pluginConfig))
```
zikaari, line 49: The plugin will use this as `pluginConfig.dependencies.storage`. It kinda looks bit abrasive to me, why not just pass it in as a flattened argument. It is highly unlikely that there'll be anything more we'll need to make available to plugins.

zikaari (commented):
We should probably make a "scoped" instance of storage accessible to each plugin instead of the "shared" storage. Although highly unlikely, but there might be a case where 2 or more plugins are using the same key and writing conflicting values, or worse using the `clear` operation.

Here's what I mean:

```ts
	const createScopedStorage = (storage: UniversalStorage, scope: string) => ({
		get: (key: string) => storage.get(`${scope}::${key}`),
		set: (key: string, value: any) => storage.set(`${scope}::${key}`)
	})

```

packages/browser/src/core/user/index.ts
```diff
@@ -398,3 +398,53 @@ export class Group extends User {
...
+  private mem: Store
+
+  options: UserOptions = {}
+
+  constructor(isDisabled: boolean, cookieOptions?: CookieOptions) {
```
chrisradek, line 409: Can we allow selectively choosing what is enabled and disabled? Maybe that means replacing `isDisabled` with something like 
```
interface StorageOptions {
  cookies: boolean,
  localStorage: boolean,
  inMemory: boolean
}
```
or 
```
interface Storage {
  cookie: Store
  localStorage: Store
  inMemory: Store
}
```

Just thinking of all the different permutations we support/want to support (just cookies, just localstorage, all of the above, nothing)

chrisradek: Thinking of this more - instead of option 2, could even pass in an array of stores - we don't really need to know if they are cookies or localstorage or whatever.

silesky: @chrisradek I think they generally only need one type of storage AFAIK (they don’t use all of them?) — so we just accept a single storage interface as an override (they could even directly pass in window.localStorage)

silesky: FYI I think isDisabled as a configuration option is odd — I don’t think classes should be responsible for creating stub versions of themselves — I’d rather have that logic be moved up a level, seems a bit more idiomatic.

PR #597, opened 2022-09-15
Support destination filters for device mode action destinations

danieljackins:
This PR allows destination filters to work with action destinations, by adding destination middleware support to them.

This change was tested by adding destination filters to an action destination and making sure events and properties were dropped properly.


packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -315,7 +315,7 @@ export function ajsDestinations(
   }
 
   const routingRules = settings.middlewareSettings?.routingRules ?? []
-  const routingMiddleware = tsubMiddleware(routingRules)
+  // const routingMiddleware = tsubMiddleware(routingRules)
```
chrisradek, line 318: Dead code to remove!

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -1,14 +1,25 @@
...
 import { loadScript } from '../../lib/load-script'
 import { getCDN } from '../../lib/parse-cdn'
+import {
+  applyDestinationMiddleware,
+  DestinationMiddlewareFunction,
+} from '../middleware'
+import { Context, ContextCancelation } from '../../core/context'
+import { Analytics } from '../../core/analytics'
+
+const klona = (evt: SegmentEvent): SegmentEvent =>
```
chrisradek, line 15: I think `klona` also exists in the ajsDestinations plugin. Might be able to shave a few bytes if we consolidate them.

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -17,6 +28,120 @@ export interface RemotePlugin {
...
+    )
+
+    if (modifiedEvent === null) {
+      ctx.cancel(
+        new ContextCancelation({
+          retry: false,
+          reason: 'dropped by destination middleware',
+        })
+      )
+      return ctx
```
chrisradek, line 67: This return shouldn't be needed because `ctx.cancel` always throws.

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -46,11 +171,15 @@ export async function remoteLoader(
...
   mergedIntegrations: Record<string, JSONObject>,
-  obfuscate?: boolean
+  obfuscate?: boolean,
+  routingMiddleware?: DestinationMiddlewareFunction
 ): Promise<Plugin[]> {
   const allPlugins: Plugin[] = []
   const cdn = getCDN()
 
+  const routingRules = settings.middlewareSettings?.routingRules ?? []
+  // const routingMiddleware = tsubMiddleware(routingRules)
```
chrisradek, line 181: Dead code to remove!

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -100,7 +229,22 @@ export async function remoteLoader(
...
+            (rule) => rule.destinationName === remotePlugin.creationName
+          )
+
+          plugins.forEach((plugin) => {
+            const wrapper = new ActionDestination(
+              remotePlugin.creationName,
+              plugin
+            )
+
+            if (routing.length && routingMiddleware) {
```
chrisradek, line 242: I think we should also check if the plugin is `destination` type. That way we don't do any filtering on cloud-mode actions that have a web component (like amplitude/braze cloud)

chrisradek (commented):
Overall I think this looks great, excited to see this middleware support added to action destinations!

Things I wanted to check:
1. We may need to move `alternativeNames` to the plugin interface then include it in our checks here: https://github.com/segmentio/analytics-next/blob/9c935795ea962c17a869d6d8143617b98e7e29a2/packages/browser/src/core/queue/event-queue.ts#L236-L239
This way if someone is currently disabling/enabling based on the current plugin name (like `FullStory (actions) trackEvent`) it still works.
2. Can we apply the above logic for logging/metrics as well?
3. I commented that we should only apply destination filters on destination plugins - wonder if that should apply for all destination middleware as well. What are your thoughts there given action destinations can have non-destination actions.

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -17,6 +28,120 @@ export interface RemotePlugin {
...
+          retry: false,
+          reason: 'dropped by destination middleware',
+        })
+      )
+      return ctx
+    }
+
+    return new Context(modifiedEvent)
+  }
+
```
silesky, line 72: Could we save some bytes by being more DRY?
Something like:
```ts
createMethod(methodName: 'track' | 'page' | 'identify' | 'alias' | 'group') {
    return async (ctx: Context): Promise<Context> => {
      if (!this.action[methodName]) return ctx

      const transformedContext = await this.transform(ctx)
      await this.action[methodName](transformedContext)

      return ctx
    }
  }
  
  alias = createMethod('track')
  group = createMethod('group')
   ... etc
```

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -17,6 +28,120 @@ export interface RemotePlugin {
...
+    const transformedContext = await this.transform(ctx)
+    await this.action.track(transformedContext)
+
+    return ctx
+  }
+
+  /* --- PASSTHROUGH METHODS --- */
+  isLoaded(): boolean {
+    return this.action.isLoaded()
```
silesky, line 129: Can't we just do? 
```ts
const  isLoaded = this.action.isLoaded
```

danieljackins:
> Overall I think this looks great, excited to see this middleware support added to action destinations!
> 
> Things I wanted to check:
> 
> 1. We may need to move `alternativeNames` to the plugin interface then include it in our checks here: https://github.com/segmentio/analytics-next/blob/9c935795ea962c17a869d6d8143617b98e7e29a2/packages/browser/src/core/queue/event-queue.ts#L236-L239
>    
>    This way if someone is currently disabling/enabling based on the current plugin name (like `FullStory (actions) trackEvent`) it still works.
> 2. Can we apply the above logic for logging/metrics as well?
> 3. I commented that we should only apply destination filters on destination plugins - wonder if that should apply for all destination middleware as well. What are your thoughts there given action destinations can have non-destination actions.

1. Done!
2. Not exactly sure what you're getting at here
3. I think it would be safer to not have that restriction just in case they're trying to use them on 'enrichment' actions or something else

packages/browser/src/plugins/remote-loader/index.ts
danieljackins: thanks!

chrisradek:
>2. Not exactly sure what you're getting at here

The context keeps logs/stats about which plugins were ran. Currently you can see which actions are ran and some stats about them. With these changes I think we'll see the wrapper plugin's name instead, so `Fullstory` instead of `Fullstory (actions) trackEvent` - also makes it look like we're running the same plugin multiple times if a destination has more than 1 action.

from our dev environment:
image

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -100,7 +190,26 @@ export async function remoteLoader(
...
+          plugins.forEach((plugin) => {
+            const wrapper = new ActionDestination(
+              remotePlugin.creationName,
+              plugin
+            )
+
+            if (
+              routing.length &&
+              routingMiddleware &&
+              plugin.type === 'destination'
```
chrisradek, line 206: Might be worth adding a comment that we only want to apply destination filters to `destination` actions so that we don't cause issues for hybrid action destinations (where we simply augment the event and send it to the cloud rather than directly to the destination)

I see future selves forgetting why we limited this :smile:

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -17,6 +25,85 @@ export interface RemotePlugin {
...
+  private _createMethod(
+    methodName: 'track' | 'page' | 'identify' | 'alias' | 'group' | 'screen'
+  ) {
+    return async (ctx: Context): Promise<Context> => {
+      if (!this.action[methodName]) return ctx
+
+      const transformedContext = await this.transform(ctx)
+      await this.action[methodName]!(transformedContext)
+
+      return ctx
```
silesky, line 78: Quick question— What’s the reason that we return the original context instead of the transformed context?

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -17,6 +25,85 @@ export interface RemotePlugin {
...
+  ready(): Promise<unknown> {
+    return this.action.ready ? this.action.ready() : Promise.resolve()
+  }
+
+  load(ctx: Context, analytics: Analytics): Promise<unknown> {
+    return this.action.load(ctx, analytics)
+  }
+
+  unload(ctx: Context, analytics: Analytics): Promise<unknown> | unknown {
+    return this.action.unload && this.action.unload(ctx, analytics)
```
silesky, line 103: just in case you aren’t a Typescript nerd, you can shorten this to “return this.action.unload?.(ctx, analytics)” IIRC

packages/browser/src/core/queue/event-queue.ts
```diff
@@ -232,9 +232,16 @@ export class EventQueue extends Emitter {
         return true
       }
 
+      const alternativeNameMatch =
+        p.alternativeNames &&
+        p.alternativeNames.filter((name) =>
```
silesky, line 237: Optional changing nitpick! p.alternativeNames?.filter would be a bit nicer IMO

chrisradek: I typically agree - though the size savings are not insignificant due to the transpilations that occur (assuming variables/field names get minified). See [this](https://www.typescriptlang.org/play?target=1#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChnIBGA-AFzIAUAlMgLwB8yAbusACa4C+uuC6IAZzDI45DFlo4i5anUbZOybrjgA6Equor1yAGS6R66kA) as an example. (Just a tradeoff - if it makes the code more readable then go with the optional chaining - I was just surprised how much larger the transpiled code was)

silesky: interesting...

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -17,6 +25,85 @@ export interface RemotePlugin {
...
+  }
+
+  addMiddleware(...fn: DestinationMiddlewareFunction[]): void {
+    this.middleware.push(...fn)
+  }
+
+  private async transform(ctx: Context): Promise<Context> {
+    const modifiedEvent = await applyDestinationMiddleware(
+      this.name,
+      klona(ctx.event),
```
silesky, line 53: Why is this serialize/deserialize neccessary? sometimes its a hacky way of deserializing a class, but the event is just an object, right?

danieljackins: We do need to mutate the event I believe, but I moved the cloning logic into applyDestinationMiddleware and added a test that will fail if the original context gets modified (ie. if klona gets removed) 👍

packages/browser/src/browser/index.ts
```diff
@@ -105,6 +105,13 @@ function hasLegacyDestinations(settings: LegacySettings): boolean {
   )
 }
 
+function hasTsubMiddleware(settings: LegacySettings): boolean {
+  return (
+    getProcessEnv().NODE_ENV !== 'test' &&
+    (settings.middlewareSettings?.routingRules?.length ?? 0) > 0
```
chrisradek, line 111: Minor: This change will does the same thing - and should shave bytes (I think `??` is transpiled to be a lot more code than expected)
```suggestion
    settings.middlewareSettings?.routingRules?.length
```

packages/browser/src/core/queue/event-queue.ts
```diff
@@ -232,9 +232,14 @@ export class EventQueue extends Emitter {
...
       }
 
+      const alternativeNameMatch = p.alternativeNames?.filter((name) =>
+        Object.keys(denyList).includes(name)
+      )
+
       // Explicit integration option takes precedence, `All: false` does not apply to Segment.io
       return (
         denyList[p.name] ??
+        alternativeNameMatch ??
```
chrisradek, line 242: Isn't `alternativeNameMismatch` an array of strings here? I think that'd be evaluated as a truthy value so we'd never hit the next conditional, or am I missing something?

chrisradek (commented):
Overall looks great! Just had 1 question that might be an issue.

packages/browser/src/browser/index.ts
chrisradek: Ah - type signature wants a boolean so you'd need to wrap the whole thing in a `Boolean` or `!!`, sorry - not sure it's worth the shaved bytes.

packages/browser/src/core/queue/event-queue.ts
danieljackins: Nice catch - fixed the logic and added a new test

chrisradek (approved):
Looks good to me! :shipit:

PR #599, opened 2022-09-21
fixes event editor for nextjs sandbox and adds options editor

### LINKED ISSUE #598: Can't type into event input field
I have the tester playground pulled up so I can fire some test events. However, when I try to type anything into the event input field I get errors like:

Screen Shot 2022-09-19 at 2 34 36 PM

If I highlight the text box, I see:

Screen Shot 2022-09-19 at 2 35 15 PM

Problem presents on both chrome and safari.

chrisradek:
Fixes #598

This PR updates the event editor in our next.js dev example to better handle user input. Previously entering characters would throw an error since the JSON was being parsed on each input event.

This PR also adds a second editor to allow passing in the options to a track/identify call. I found this useful when testing integration overrides.

No changeset needed since there are no user-facing changes in published packages.

### DIFF
--- a/examples/with-next-js/pages/index.tsx
+++ b/examples/with-next-js/pages/index.tsx
@@ -45,7 +45,7 @@ export default function Home(): React.ReactElement {
     setWriteKey,
   } = useAnalytics()
 
-  const mockTraits = (): any => {
+  const mockTraits = (): string => {
     const fakerFns = [
       ...Object.entries(faker.name),
       ...Object.entries(faker.commerce),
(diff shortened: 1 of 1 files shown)


pooyaj (approved):

dcow (commented):
👍🏻

examples/with-next-js/pages/index.tsx
```diff
@@ -84,7 +87,8 @@ export default function Home(): React.ReactElement {
         console.error(err)
         setCtx(undefined)
         setAnalytics(undefined)
-        setEvent({})
+        setEvent('{}')
```
silesky, line 90: Eek

silesky (approved):
Ooh… two editors now! 👍 thanks.

also, thanks @dcow for reporting.

PR #600, opened 2022-09-22
limit node event options

silesky:


pooyaj:
Hmmm, I think as far as I remember, AJS tries to be "smart" about what we put in the options object, and selectively moves whatever is in the options either to the root of the event payload, or put them inside the `context` ( for example things like `userId` or `integrations` will move to the root, random properties will go in `context`, and if we have a `context` inside `options` then its content will be spread inside the event payload's context object.

pooyaj:
for the node library, I think if we can get rid of all the AJS's magic in event normalization it would be cleaner. With AJS, there was a concern with backwards compatibility, so we had to preserve the old logic. But with Node lib it is not a concern ( and the existing NodeJS has no magic anyways )

chrisradek (approved):
💥

PR #601, opened 2022-09-23
elaborate on emitter api

silesky:
- tighten emitter contract
- elaborate on emitter API by creating a standard emitter error in core and sharing it

### DIFF
--- a/packages/core/src/analytics/dispatch-emit.ts
+++ b/packages/core/src/analytics/dispatch-emit.ts
@@ -1,5 +1,4 @@
 import { dispatch } from './dispatch'
-import { CoreContext } from '../context'
 
 /* Dispatch function, but swallow promise rejections and use event emitter instead */
 export const dispatchAndEmit = async (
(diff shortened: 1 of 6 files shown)


packages/core/src/analytics/dispatch-emit.ts
```diff
@@ -8,13 +7,20 @@ export const dispatchAndEmit = async (
...
+        message: 'failed to deliver event',
+        ctx: ctx,
+      })
     } else {
       emitter.emit(event.type, ctx)
       return ctx
     }
   } catch (err) {
-    emitter.emit('error', err as CoreContext)
-    return err
```
pooyaj, line 18: `dispatchAndEmit` is not used outside Node ... right? Just wondering if removing the `return err` would break anything. If not, all good 👍

pooyaj (approved):

packages/core/src/analytics/dispatch-emit.ts
silesky: Yes, this is a node only method (for now) -- but I decided on putting it in core because I thought it might make sense to have browser adopt it.

PR #602, opened 2022-09-23
add dispatch unit tests and minor refactoring

silesky:
There were no unit tests covering dispatch, so I needed to add some to sleep at night, (and also because it seems a fairly importanish part of core). In doing so, uncovered weird behavior -- would like clarity on it.

### DIFF
--- a/packages/core/src/analytics/__tests__/dispatch.test.ts
+++ b/packages/core/src/analytics/__tests__/dispatch.test.ts
@@ -1,3 +1,13 @@
+/**
+ * Properly type mocked functions to make it easy to do assertions
+ * for example, myModule.mock.calls[0] will have the typed parameters instead of any.
+ *
+ * TODO: share with rest of project
+ */
+type JestMockedFn<Fn> = Fn extends (...args: infer Args) => infer ReturnT
+  ? jest.Mock<ReturnT, Args>
+  : never
+
 const isOnline = jest.fn().mockReturnValue(true)
 const isOffline = jest.fn().mockReturnValue(false)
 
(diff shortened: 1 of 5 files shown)


packages/core/src/analytics/__tests__/dispatch.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    })
+    expect(dispatchSpy).toBeCalledWith(screenCtxMatcher)
+    expect(invokeCallback).toBeCalledTimes(1)
+    const [ctx, _cb] = invokeCallback.mock.calls[0]
+    expect(ctx).toEqual(screenCtxMatcher)
+    expect(_cb).toBe(cb)
+  })
+
+  // TODO: Inconsistent behavior? This seems like a bug.
+  it('should have inconsistent timeout behavior where the delay is different based on whether timeout is explicitly set to 1000 or not', async () => {
```
silesky, line 114: cc @pooyaj

packages/core/src/analytics/dispatch.ts
```diff
@@ -40,15 +46,13 @@ export async function dispatch(
...
-  const timeoutInMs = options?.timeout
 
   if (options?.callback) {
     dispatched = await invokeCallback(
       dispatched,
       options.callback,
-      Math.max((timeoutInMs ?? 300) - elapsedTime, 0),
-      timeoutInMs
+      getDelay(startTime, options.timeout), //  TODO: I have no idea why this delay is based on the timeout -- so if you set a timeout of 5seconds, it will (in addition) actually delay ~5seconds before invoking.
+      options.timeout
```
pooyaj, line 55: I think that makes sense. So normally 300ms timeout, but let user configure the value to whatever they want?

packages/core/src/analytics/__tests__/dispatch.test.ts
pooyaj: I think this is intentional, 300ms default, but it is configurable.

pooyaj (approved):

Issue #607, opened 2022-10-03
'pageunload' does not work on iOS Safari


jonathan-fielding:
Recently we have tried to switch to using batching for our events using Segment analytics.js however we have found that we have lost approx 15% of our events being fired, upon investigation we found this was isolated to the last events being fired on iOS Safari. 

We did some investigation and found that the `pageunload` event is not supported by Safari on iOS. There are a few issues raised on the internet that support this.

* https://stackoverflow.com/questions/4127621/is-there-any-way-to-use-window-onbeforeunload-on-mobile-safari-for-ios-devices
* https://github.com/GoogleChromeLabs/page-lifecycle/issues/11#issuecomment-770031911

The official documentation from Apple https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html shows that they deprecated `pageunload` as well which supports the hypothesis that this won't be fixed.

Interestingly the `pageunload` event does work in Desktop Safari

The suggestion is that instead of using `pageunload` we switch the library to use `pagehide` which has good browser support with Chrome, Edge, Firefox, Opera and Safari having good long term support https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event

I debugged this with @agouil who is going to raise a PR for this change

silesky:
Thank you for bringing our attention to this! It would be great to replace `beforeunload`, but when it comes to simply replacing it with `pagehide` I left a comment here https://github.com/segmentio/analytics-next/pull/608#issuecomment-1267404204

PR #608, opened 2022-10-03
Listen to 'pagehide' event instead of 'unload' when exiting a page

agouil:
This PR updates the event the library is listening to when the browser page is closed from `beforeunload` to `pagehide` which is supported by all browsers include Mobile Safari on iOS.

More information about the above change can be found in this issue: https://github.com/segmentio/analytics-next/issues/607

### DIFF
--- a/.changeset/honest-pants-ring.md
+++ b/.changeset/honest-pants-ring.md
@@ -1,0 +1,6 @@
+---
+'@segment/analytics-next': minor
+'@segment/analytics-core': minor
+---
+
+Listen to 'pagehide' event instead of 'beforeunload' event when closing the browser page and flushing the persisted queue
--- a/packages/browser/src/lib/priority-queue/__tests__/persisted.test.ts
+++ b/packages/browser/src/lib/priority-queue/__tests__/persisted.test.ts
@@ -43,7 +43,7 @@ describe('Persisted Priority Queue', () => {
     jest
       .spyOn(window, 'addEventListener')
       .mockImplementation((evt, handler) => {
-        if (evt === 'beforeunload') {
+        if (evt === 'pagehide') {
           onUnload = handler
         }
       })
@@ -104,7 +104,7 @@ describe('Persisted Priority Queue', () => {
       jest
         .spyOn(window, 'addEventListener')
         .mockImplementation((evt, handler) => {
-          if (evt === 'beforeunload') {
+          if (evt === 'pagehide') {
             onUnload = handler
           }
         })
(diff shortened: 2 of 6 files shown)


silesky:
Thanks for your PR. I think this is an important point, but it appears to only be part of a solution.

 Per the MDN usage notes: 
> Like the [unload](https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event) and [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) events, this event is not reliably fired by browsers, especially on mobile. For example, the pagehide event is not fired at all in the following scenario:
- A mobile user visits your page.
- The user then switches to a different app.
- Later, the user closes the browser from the app manager.

https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event#usage_notes

silesky:
In my testing, `pagehide` doesn't trigger on chrome desktop when clicking a link, etc -- unlike beforeunload. I could only get pagehide to trigger in iOS safari.

agouil:
@silesky, I agree that `pagehide` is not the perfect solution and it will not work for all use cases on mobile devices. 

Further to the article you shared, the `beforeunload` event should not be used for listening for the end of the user's session but only for checking of unsaved work, according to https://developer.chrome.com/blog/page-lifecycle-api/#the-beforeunload-event.

Below, I tried to list out the main differences between `pagehide` and `visibilitychange`.

`pagehide`
- Fires mostly on navigation. Does not always fire when (current/background) tabs are closed or when switching to another app on mobile devices.
- Per [MDN's notes](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event#usage_notes): "The best event to use to signal the end of a user's session is the visibilitychange event. In browsers that don't support visibilitychange the pagehide event is the next-best alternative."

`visibilitychange`
- Fires on navigation, switching tabs, closing tabs, minimizing or closing the browser, or, on mobile, switching from the browser to a different app. 
- Per [MDN's notes](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event#usage_notes): "Transitioning to hidden visibility is the last event that's reliably observable by the page, so developers should treat it as the likely end of the user's session (for example, for sending analytics data)."
- There are known bugs with some older versions of Safari (e.g. 14.0.x) not firing this event reliably when navigating or closing the current tab. However, `pagehide` is triggered instead.

I think we can come up with a solution by using both of the events above which will cover most of the use cases on both desktop and mobile for all recent versions of browsers. My suggestion is to use the `visibilitychange` as the primary event to  send the last batch of requests with a fallback to `pagehide` when the event isn't triggered or not supported.

silesky:
@agouil Thanks! Do you mind taking a looking at my PR https://github.com/segmentio/analytics-next/pull/613

PR #613, opened 2022-10-05
fix unload detection

silesky:
This is a bug and enhancement
- fixes https://github.com/segmentio/analytics-next/issues/607  reported by @agouil  and @jonathan-fielding 
- now saves / fires events on loss of tab focus and tab close (both mobile and desktop), reducing event loss because of sudden exit by probably 50% or more.

This hybrid approach is because "beforeunload" is unreliable, and "pagehide" by itself will not fire on tab switch / close. visibilitychange works well, but it doesn't have as much support for older browsers as pagehide, and there are still some bugs that can cause it to not fire. https://stackoverflow.com/questions/3239834/window-onbeforeunload-not-working-on-the-ipad/52864508#52864

### DIFF
--- a/packages/browser/src/lib/on-page-leave.ts
+++ b/packages/browser/src/lib/on-page-leave.ts
@@ -1,0 +1,29 @@
+/**
+ * Register event listener on document that fires when:
+ * * tab change or tab close (in mobile or desktop)
+ * * click back / forward button
+ * * click any link or perform any other navigation action
+ * * soft refresh / hard refresh
+ *
+ * adapted from https://stackoverflow.com/questions/3239834/window-onbeforeunload-not-working-on-the-ipad/52864508#52864508,
+ */
+export const onPageLeave = (cb: (...args: unknown[]) => void) => {
+  let unloaded = false // prevents double firing if both are supported
+
+  // using document instead of window because of bug affecting browsers before safari 14 (detail in footnotes https://caniuse.com/?search=visibilitychange)
+  document.addEventListener('pagehide', () => {
+    if (unloaded) return
+    unloaded = true
+    cb()
+  })
+
+  document.addEventListener('visibilitychange', () => {
+    if (document.visibilityState == 'hidden') {
+      if (unloaded) return
+      unloaded = true
+      cb()
+    } else {
+      unloaded = false
+    }
+  })
+}
--- a/packages/browser/src/plugins/segmentio/batched-dispatcher.ts
+++ b/packages/browser/src/plugins/segmentio/batched-dispatcher.ts
@@ -1,5 +1,6 @@
 import unfetch from 'unfetch'
 import { SegmentEvent } from '../../core/events'
+import { onPageLeave } from '../../lib/on-page-leave'
 
 let fetch = unfetch
 if (typeof window !== 'undefined') {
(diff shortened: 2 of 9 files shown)


packages/core/src/priority-queue/persisted.ts
```diff
@@ -1,6 +1,7 @@
 import { PriorityQueue } from '.'
```
silesky, line 1: This file is not in use FYI, ended up not needing to bring it into core and will remove it when I get a chance.

packages/browser/src/lib/on-page-leave.ts
```diff
@@ -0,0 +1,27 @@
...
+ * * click relative link (same domain)
+ * * soft refresh / hard refresh
+ *
+ * adapted from https://stackoverflow.com/questions/3239834/window-onbeforeunload-not-working-on-the-ipad/52864508#52864508,
+ */
+export const onPageLeave = (cb: (...args: any[]) => void) => {
+  let unloaded = false // prevents double firing if both are supported
+
+  // using document instead of window because of bug affecting browsers before safari 14 (detail in footnotes https://caniuse.com/?search=visibilitychange)
+  document.addEventListener('pagehide', () => {
```
silesky, line 14: document instead of window FYI because of an old event bubbling issue (see comment). This _shouldn't_ cause issues that I an forsee, and it's IMO as a rule of thumb it's better to use document rather than window when possible.

pooyaj (approved):
Looks great 🚢

agouil (commented):
Code changes look good to me. I suggest adding a test where you fire both events and you test that the `cb()` is only called once.

packages/browser/src/lib/on-page-leave.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (unloaded) return
+    unloaded = true
+    cb()
+  })
+
+  document.addEventListener('visibilitychange', () => {
```
chrisradek, line 20: Should we reset the `unloaded` back to false if the visibilityState is not `hidden`? That way if someone comes back to the tab we still support the same unload behaviour if they close later on?

packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -92,7 +93,7 @@ export default function batch(apiHost: string, config?: BatchingConfig) {
     }, timeout)
   }
 
-  window.addEventListener('beforeunload', () => {
+  onPageLeave(() => {
     pageUnloaded = true
```
chrisradek, line 97: Is `pageUnloaded` still needed now that `onPageLeave()` has its own check?

silesky: I don't think anything changed in that regard, the variable really should be called "keepAlive"

packages/browser/src/lib/on-page-leave.ts
silesky: > Should we reset the unloaded back to false if the visibilityState is not hidden? That way if someone comes back to the tab we still support the same unload behaviour if they close later on?This is a good call out!

> Edit: Not sure if that also means we need to rehydrate the events and put them in the queue to ensure they're sent if we return to the existing tab, or if a.js will already just try to resend them.

This is a good point --  this PR is focused on the batched-dispatcher, which leaves the scenario of events in the main event queue -- we don't want to save to localStorage on tab away since everything (including unsent events in the queue) should be suspended in-memory anyway.

chrisradek: Ah awesome, that makes sense then :phew:

chrisradek (approved):
Looks great, ship it!

PR #616, opened 2022-10-06
Add bundled analytics support in Inspector

zikaari:

Corresponding Inspector PR - https://github.com/segmentio/segment-inspector/pull/24

### DIFF
--- a/.changeset/serious-shirts-crash.md
+++ b/.changeset/serious-shirts-crash.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': minor
+---
+
+Bundled analytics support in Inspector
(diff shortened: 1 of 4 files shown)


silesky:
Looks good, except for the build error -- I assume this isn't done?

silesky:
I think the build is causing a typescript error because of this line, which is conflicting with ambient declaration merging:
image

This is the correct declaration:
```ts
declare global {
  interface Window {
    analytics: AnalyticsSnippet
  }
  ```
  
All this is complicated by the fact that there's a circular dependency here with analytics-next and  '@segment/inspector-webext'. Publishing inspector-webext's types as a separate dependency or just removing that dependency altogether from analytics-next might be a good idea (and doing the typing locally).

silesky (approved):

PR #619, opened 2022-10-08
consolidate global typings

silesky:
- gets rid of JQuery / Zepto global which can have side effects on anyone importing our lib


zikaari (approved):
👍

PR #629, opened 2022-10-18
Fix action destination middleware bug

danieljackins:
This PR fixes a bug where destination filters (and other destination middleware) were being applied to all action destination types instead only those of type 'destination'. This was breaking behaviour such as amplitude adding a sessionId to the integrations object.

### DIFF
--- a/packages/browser/src/plugins/remote-loader/index.ts
+++ b/packages/browser/src/plugins/remote-loader/index.ts
@@ -44,7 +44,9 @@ export class ActionDestination implements Plugin {
   }
 
   addMiddleware(...fn: DestinationMiddlewareFunction[]): void {
-    this.middleware.push(...fn)
+    if (this.type === 'destination') {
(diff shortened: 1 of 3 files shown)


pooyaj (approved):

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -72,7 +72,11 @@ export class ActionDestination implements Plugin {
     return async (ctx: Context): Promise<Context> => {
       if (!this.action[methodName]) return ctx
 
-      const transformedContext = await this.transform(ctx)
+      let transformedContext: Context = ctx
+      if (this.type === 'destination') {
```
mericsson, line 76: Should we add a comment here for why we have this if statement?

chrisradek (approved):

chrisradek (approved):

PR #630, opened 2022-10-19
fix: can init with settings [wip]

hbrls:
The browser load related code hard-code only the `writeKey` param. Other settings such as `cdn` has no way to pass in.

I'm not sure if this pr fix it correctly. It's a wip showcase how I expect the function signature behavior.

Please help me to clarify. Thanks.


silesky:
@hbrls  we don't generally accept unsolicited feature PRs. Please create a feature request issue and we can discuss possible solutions there.

hbrls:
Yes, this pr is actually a feature request. I just thought the code diff would explain it better.

Should I create another feature request issue to discuss there?

PR #634, opened 2022-10-20
Add smoke tests for integrations

chrisradek:
Adds a new package for performing smoke tests using the UMD bundle (what's served over CDN) and actual browser action destinations.

This uses playwright to load actual browsers to run the tests. The tests I've written also check the HTTP requests for correctness. Locally, the tests will run in firefox, chromium, and webkit browsers. I had a lot of trouble getting the correct dependencies installed to run firefox/webkit in CI so for now just running against chromium there.

I have one test that really needed playwright - segment retries across page loads. This currently fails since it requires #714 before it will pass, so I have it marked to skip for now. Will remove the skip once this is merged with #714.

No changesets since this doesn't change any user-facing packages.


packages/browser-integration-tests/src/segment-retries.test.ts
```diff
@@ -0,0 +1,64 @@
...
+            headers: {
+              'Content-Type': 'application/json',
+            },
+            body: JSON.stringify(new SettingsBuilder().build()),
+          })
+        }
+      )
+    })
+
+    test.skip('supports retries on page navigation', async ({ page }) => {
```
arielsilvestri, line 28: Did you mean to leave this skip in?

chrisradek: Yes - I wrote this test for #714 and verified locally that this test passes once #714 is merged. It's functionality that should work today but doesn't. Basically once this gets merged in I'll rebase #714 and remove the `skip` here. I left it in this PR just so that if anything were to happen the test case wouldn't be lost to time :grin:

silesky (approved):

PR #635, opened 2022-10-21
[feature] add option to disable converting ISO strings to dates

chrisradek:
This adds a new load option - `disableAutoISOConversions`. This turns off the automatic conversion of event fields where any strings that appear as an ISO8601 string (including just dates, e.g. `YYYY-MM-DD`) become `Date` objects.

## Background

Through `@segment/facade`, any events passed to classic integrations had any ISO strings converted to Dates. I suspect this was intentional behavior because the way we cloned events (`JSON.parse(JSON.stringify(event))`) converted Dates to ISO8601 strings, so facade was converting them back to dates.

However, some users want their ISO strings to be passed to integrations as is, especially when the string represents a calendar date and not a time.

## PR details
`@segment/facade` supports turning off this conversion by setting `traverse` to false, so the new option ultimately sets the `traverse` flag.

I also changed how we clone events so that users can still specify a property as a `Date` object and have that carried over to the cloned event regardless of the new setting. It looks like analytics.js classic also did a deep clone of the event instead of our `JSON.parse(JSON.stringify(event))` trick, so this shouldn't cause any unintended consequences.

I made this field default to `false`, meaning users need to opt-in to this new behavior. I made it opt-in to prevent potentially impacting user-defined destination middleware.

### DIFF
--- a/.changeset/khaki-mayflies-punch.md
+++ b/.changeset/khaki-mayflies-punch.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': minor
+---
+
+Adds a new load option `disableAutoISOConversions` that turns off converting ISO strings in event fields to Dates for integrations.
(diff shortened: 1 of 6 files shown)


silesky:
Very happy to get rid of JSON.parse(JSON.stringify(event))!

silesky (approved):

PR #637, opened 2022-10-24
Delay initialization (lazy loading)

silesky:
```ts

const analytics = new AnalyticsBrowser() // nothing loaded yet

// queue events
analytics.track('foo')
analytics.identify('bar')

if (userConsentsToTracking) {
  analytics.load({ writeKey: 'foo' }) // loaded. queued events should now be sent.
}


```

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ document.body?.addEventListener('click', () => {
 ```
 
 ## Lazy / Delayed Loading
-You can load a buffered version of analytics that requires `.load` to be explicitly called before initatiating any network activity. This can be useful if you want to wait for a user to consent before fetching any tracking destinations or sending buffered events to segment.
+You can load a buffered version of analytics that requires `.load` to be explicitly called before initiating any network activity. This can be useful if you want to wait for a user to consent before fetching any tracking destinations or sending buffered events to segment.
 
 - ⚠️ ️`.load` should only be called _once_.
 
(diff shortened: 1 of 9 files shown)


zikaari:
I'm curious wouldn't this have worked anyway? We've always buffered up events until `load`, and flushed them when loaded. Can you explain behavioral differences between the existing implementation and this one?

packages/browser/src/browser/index.ts
```diff
@@ -309,12 +309,46 @@ async function loadAnalytics(
...
+        loadAnalytics(settings, options, buffer)
+      )
+    )
+
+    this._resolveLoadStart = (settings, options) =>
+      resolveLoadStart([settings, options])
+  }
+
+  /**
+   * Starts loading an analytics instance including:
```
chrisradek, line 340: Can you add a note that this should only be called once? Want to make it clear that calling this with different settings won't essentially 'reload' analytics.

packages/browser/src/lib/create-deferred.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Return a promise that can be externally resolved
+ */
+export const createDeferred = <T>() => {
```
chrisradek, line 4: This is a better name than `extractPromiseParts` :grin:

README.md
```diff
@@ -53,6 +53,29 @@ document.body?.addEventListener('click', () => {
 })
 ```
 
+## Lazy / Delayed Loading
```
chrisradek, line 56: Thanks for adding this section!

chrisradek (approved):

PR #638, opened 2022-10-24
RFC - Storage Layer for Plugins V2 💿 

pooyaj:
Second version of the RFC ( first version [here](https://github.com/segmentio/analytics-next/pull/594) ) 

Problem: If a plugin needs to use browser storage ( cookies, LS, etc. ), it has to implement the storage handling functionality, and Cookie handling in particular requires decent amount of code. Furthermore, plugins aren't aware of AJS storage configuration, so they may use storage regardless of how users configured the AJS.

In order to enable plugins ( our particular use-case is action destinations ) access the storage layer of AJS, we are creating a universalStorage object during initialization and pass it down to plugins as a dependency when registering plugins. `UniversalStorage` is an interface over a collection of different storage types like cookie, localStorage, and memory that can persist to or read data from those collection of storage objects. 

Few things of note:
- Universal storage can be configured with a set of targets: `cookie` `localStorage` `memory` 
- For `User` storage, we create three instance of universal storage: one for identity, one for legacy identity, and one for traits 
- For passing storage to plugins, we create a universal storage with all three targets ( unless storage is disabled ) and pass that down to plugins. Plugins can use all the targets, or they can override the targets when calling `set`/`get` methods

### DIFF
--- a/packages/browser/src/core/user/index.ts
+++ b/packages/browser/src/core/user/index.ts
@@ -251,20 +251,22 @@ export class UniversalStorage {
   }
 
   static getUniversalStorage(
-    hasBrowserStorage: boolean,
+    defaultTargets: StoreType[] = ['cookie', 'localStorage', 'memory'],
     cookieOptions?: CookieOptions
   ): UniversalStorage {
     const stores = []
 
-    if (hasBrowserStorage && Cookie.available()) {
+    if (defaultTargets.includes('cookie') && Cookie.available()) {
       stores.push(new Cookie(cookieOptions))
     }
 
-    if (hasBrowserStorage && LocalStorage.available()) {
+    if (defaultTargets.includes('localStorage') && LocalStorage.available()) {
       stores.push(new LocalStorage())
     }
 
-    stores.push(new Store())
+    if (defaultTargets.includes('memory')) {
+      stores.push(new Store())
+    }
 
     return new UniversalStorage(stores)
   }
@@ -277,19 +279,14 @@ export class User {
   private traitsKey: string
   private anonKey: string
   private cookieOptions?: CookieOptions
-  private universalStore: UniversalStorage
 
-  private legacyUserStoreTargets: StoreType[] = []
-  private traitsStoreTargets: StoreType[] = []
-  private identityStoreTypes: StoreType[] = []
+  private legacyUserStore: UniversalStorage
+  private traitsStore: UniversalStorage
+  private identityStore: UniversalStorage
 
   options: UserOptions = {}
 
-  constructor(
-    options: UserOptions = defaults,
-    cookieOptions?: CookieOptions,
-    universalStore?: UniversalStorage
-  ) {
+  constructor(options: UserOptions = defaults, cookieOptions?: CookieOptions) {
     this.options = options
     this.cookieOptions = cookieOptions
 
@@ -300,24 +297,41 @@ export class User {
     const isDisabled = options.disable === true
     const shouldPersist = options.persist !== false
 
-    this.universalStore =
-      universalStore ||
-      UniversalStorage.getUniversalStorage(shouldPersist, cookieOptions)
+    let defaultStorageTargets: StoreType[] = isDisabled
+      ? []
+      : shouldPersist
+      ? ['cookie', 'localStorage', 'memory']
+      : ['memory']
 
-    if (!isDisabled) {
-      this.identityStoreTypes.push('memory', 'cookie')
-      this.legacyUserStoreTargets.push('cookie')
-      this.traitsStoreTargets.push('memory')
-      if (!options.localStorageFallbackDisabled) {
-        this.traitsStoreTargets.push('localStorage')
-        this.identityStoreTypes.push('localStorage')
-      }
+    if (options.localStorageFallbackDisabled) {
+      defaultStorageTargets = defaultStorageTargets.filter(
+        (t) => t !== 'localStorage'
+      )
     }
 
-    const legacyUser = this.universalStore.get<{
+    this.identityStore = UniversalStorage.getUniversalStorage(
+      defaultStorageTargets,
+      cookieOptions
+    )
+
+    // using only cookies for legacy user store
+    this.legacyUserStore = UniversalStorage.getUniversalStorage(
+      defaultStorageTargets.filter(
+        (t) => t !== 'localStorage' && t !== 'memory'
+      ),
+      cookieOptions
+    )
+
+    // using only localStorage / memory for traits store
+    this.traitsStore = UniversalStorage.getUniversalStorage(
+      defaultStorageTargets.filter((t) => t !== 'cookie'),
+      cookieOptions
+    )
+
+    const legacyUser = this.legacyUserStore.get<{
       id?: string
       traits?: Traits
-    }>(defaults.cookie.oldKey, this.legacyUserStoreTargets)
+    }>(defaults.cookie.oldKey)
     if (legacyUser) {
       legacyUser.id && this.id(legacyUser.id)
       legacyUser.traits && this.traits(legacyUser.traits)
@@ -330,13 +344,10 @@ export class User {
       return null
     }
 
-    const prevId = this.universalStore.getAndSync(
-      this.idKey,
-      this.identityStoreTypes
-    )
+    const prevId = this.identityStore.getAndSync(this.idKey)
 
     if (id !== undefined) {
-      this.universalStore.set(this.idKey, id, this.identityStoreTypes)
+      this.identityStore.set(this.idKey, id)
 
       const changingIdentity = id !== prevId && prevId !== null && id !== null
       if (changingIdentity) {
@@ -345,20 +356,14 @@ export class User {
     }
 
     return (
-      this.universalStore.getAndSync(this.idKey, this.identityStoreTypes) ??
-      this.universalStore.get(
-        defaults.cookie.oldKey,
-        this.legacyUserStoreTargets
-      ) ??
+      this.identityStore.getAndSync(this.idKey) ??
+      this.legacyUserStore.get(defaults.cookie.oldKey) ??
       null
     )
   }
 
   private legacySIO(): [string, string] | null {
-    const val = this.universalStore.get(
-      '_sio',
-      this.legacyUserStoreTargets
-    ) as string
+    const val = this.legacyUserStore.get('_sio') as string
     if (!val) {
       return null
     }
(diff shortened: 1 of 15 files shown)


packages/browser/src/core/user/index.ts
```diff
@@ -183,18 +200,92 @@ export interface CookieOptions {
   sameSite?: string
 }
```
silesky, line 202: This is neat. Is there a way we can make this fully generic -- like I did with the [event emitter](https://github.com/segmentio/analytics-next/blob/3358219e1757384f3be66109971a1c9e5f384dda/packages/core/src/emitter/index.ts#L18)?


```ts
export class UniversalStorage<Data extends Record<string, unknown>> {
..

public get<K extends keyof Data>(
    key: K,
    storeTypes?: StoreType[]
  ): Data[K] | null {
    let val = null
    for (const store of this.getStores(storeTypes)) {
      val = store.get<Data[K]>(key as string) // store can be made generic too
      if (val) {
        return val
      }
    }
    return null
  }
```

would lead to something like:

image
image

packages/browser/src/core/user/index.ts
```diff
@@ -183,18 +200,92 @@ export interface CookieOptions {
...
+  public push(store: Store) {
+    this.stores.push(store)
+  }
+
+  public getAndSync<T>(key: string, storeTypes?: StoreType[]): T | null {
+    const val = this.get(key, storeTypes)
+
+    return this.set(
+      key,
+      typeof val === 'number' ? val.toString() : val,
```
silesky, line 225: What's the reasoning behind coercing to number here rather than deferring to set? Just flagging this, as it feels like this might be an unintended side effect?

packages/browser/src/core/user/index.ts
```diff
@@ -183,18 +200,92 @@ export interface CookieOptions {
...
+    return storeTypes
+      ? this.stores.filter((s) => storeTypes.indexOf(s.getType()) !== -1)
+      : this.stores
+  }
+
+  public push(store: Store) {
+    this.stores.push(store)
+  }
+
+  public getAndSync<T>(key: string, storeTypes?: StoreType[]): T | null {
```
silesky, line 220: Something feels a bit off about this method -- it breaks command-query separation, but that's maybe not it (returning data in an update isn't that uncommon) -- maybe calling it `syncAndGet` would be a bit better, to show that its primary function should be to sync.

packages/browser/src/core/user/index.ts
pooyaj: Great suggestion. Added. Could you please if there is any TS smells in my changes or anything can be improved?

packages/browser/src/core/user/index.ts
pooyaj: This is a legacy behavior. I think the reason seems to be AJS classic may set number cookies, but in AJS next we want to convert those to string ( not sure why, maybe because our cookie library doesn't support them 🤔 ) ... I think the problem with this code shows even more with stronger typings that I just added, hence had to add the `@ts-ignore`. It breaks the assumption that a given cookie key will always have a certain value type :(

packages/browser/src/core/user/index.ts
pooyaj: Same, not a fan of this method either. The reason it exists is probably two folds: 
- If a value exists in one storage but not others ( for whatever reason, i.e., cookies are cleared but localStorage is not ) it syncs the identities across the storages again 
- There is some formatting mismatch between AJS classic and next cookies, and this makes sure that old AJS cookies are re-written in the new format as soon as we call `getAndSync` 

The reason not refactoring this is to reduce the blast reduce of this PR. This touches fairly critical part of the code, and don't want to break things in one huge PR.

packages/browser/src/core/user/index.ts
silesky: I don't see any of the changes yet -- did you push?

pooyaj: Yes, here is the commit: https://github.com/segmentio/analytics-next/pull/638/commits/c17e5fc572f8071dfa31572e4be6145db8e670f4

packages/browser/src/core/user/index.ts
```diff
@@ -183,18 +200,115 @@ export interface CookieOptions {
...
+      val = store.get<Data[K]>(key as string)
+      if (val) {
+        return val
+      }
+    }
+    return null
+  }
+
+  public set<K extends keyof Data>(
+    key: string,
```
silesky, line 250: should be "K" rather than string?

packages/browser/src/core/user/index.ts
```diff
@@ -183,18 +200,115 @@ export interface CookieOptions {
...
+      ? this.stores.filter((s) => storeTypes.indexOf(s.getType()) !== -1)
+      : this.stores
+  }
+
+  public push(store: Store) {
+    this.stores.push(store)
+  }
+
+  public getAndSync<K extends keyof Data>(
+    key: string,
```
silesky, line 221: was this meant to be "K" rather than string?

packages/browser/src/core/user/index.ts
silesky: Gotcha, I kind of missed that this logic already existed -- all of this context would be great as a comment!

packages/browser/src/core/user/index.ts
silesky: Gotcha, I kind of missed that this logic already existed -- all of this context would be great as a comment!

packages/browser/src/core/user/index.ts
silesky: I see the build error --  you could force people to always explicitly pass data or, if we don't care, we might want to add add a default generic parameter like so:
`UniversalStorage<Data extends Record<string, unknown> = Record<string, unknown>>`

chrisradek: > I know one person that really thinks default generic parameters can be sometimes "harmful"

What? But they're so handy!

packages/browser/src/core/user/index.ts
```diff
@@ -183,18 +203,123 @@ export interface CookieOptions {
...
+    this.stores = stores || []
+  }
+
+  private getStores(storeTypes: StoreType[] | undefined): Store[] {
+    return storeTypes
+      ? this.stores.filter((s) => storeTypes.indexOf(s.getType()) !== -1)
+      : this.stores
+  }
+
+  public push(store: Store) {
```
chrisradek, line 219: Is `push` public so that anyone can write their own implementation and add it?

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -109,7 +113,7 @@ function referrerId(
   }
 
   if (!disablePersistance) {
```
chrisradek, line 115: We probably don't need this check anymore since the universal storage handles this for us now right?

packages/browser/src/core/user/index.ts
pooyaj: Good catch, removed it as it is not used anywhere. I think if we see a need to add an storage on the fly after creation of a UniversalStorage object we can add it back.

chrisradek (approved):
:ship::it:

silesky (approved):
🇮🇷 🇨🇦 🚀

zikaari:
Carrying over the comment from previous PR about scoping the storage access - https://github.com/segmentio/analytics-next/pull/594#pullrequestreview-1109959751

packages/browser/src/core/user/index.ts
```diff
@@ -128,22 +141,26 @@ export class Cookie extends Store {
...
-}
 
-class NullStorage extends Store {
-  get = (_key: string): null => null
-  set = (_key: string, _val: unknown): null => null
-  remove = (_key: string): void => {}
+  getType = (): StoreType => 'cookie'
 }
 
 export class LocalStorage extends Store {
```
silesky, line 148: Maybe we can remove some of the WTFs from this class:
```ts
  get<T>(key: string): T | null {
    const val = localStorage.getItem(key) // this should be wrapped in a try/catch in case storage is not available
    if (val) {
      try {
        return JSON.parse(val)
      } catch (e) {
          return JSON.parse(JSON.stringify(val)) // this is a serious code smell =( I have no idea what this is.
         }
    }
    return null
  }
```

packages/browser/src/core/user/index.ts
```diff
@@ -128,22 +141,26 @@ export class Cookie extends Store {
...
+  static _available: boolean | undefined
+
   static available(): boolean {
+    if (LocalStorage._available !== undefined) {
```
silesky, line 152: If we're wrapping localStorage.getItem .setItem in a try catch, do we need extra "availability" logic? I feel like it'd be cleaner here to just fail silently.

packages/browser/src/core/user/index.ts
```diff
@@ -173,6 +190,8 @@ export class LocalStorage extends Store {
   remove(key: string): void {
     return localStorage.removeItem(key)
   }
+
+  getType = (): StoreType => 'localStorage'
```
silesky, line 194: Hey, do we really need this? We can already check instanceOf -- At the very least, I think it would make more sense as a `readonly field` (e.g. readonly type = 'localStorage'), rather than a method?

pooyaj: Revising this comment, after delay. The UniversalStorage will be passed into the action-destinations, and I want the action-destinations to have a nice way to choose their storage targets: `i.e., storage.get('test', ['cookies'])`  ... Not sure possible cleanly here 🤔

packages/browser/src/core/user/index.ts
pooyaj: We can probably refactor in a subsequent PR. This code has changed since the comment, and don't want to complicate this PR further.

packages/browser/src/core/user/index.ts
pooyaj: Yes, good point. Probably the available can be replaced with try/catch. Another one of those changes that I want to avoid in the scope of this PR. Maybe we can improve in a subsequent work. Also as for the minification, I think this is a bigger problem across the code-base. The regular class methods/properties are all remain un-minified unfortunately. I think bigger piece of work would be to refactor this whole thing to not be class based 😆

pooyaj: Refactored to not rely on caching anymore, but keep availability logic to avoid breaking anything.

packages/browser/src/core/user/index.ts
silesky: Nice!

silesky: 👍

PR #640, opened 2022-10-25
Adds batching support to AnalyticsNode

chrisradek:
This PR adds batching support to AnalyticsNode.

## Plugin behavior

This plugin supports batching events and sending batches when 1 of 3 conditions is met:
1. Number of events in batch matches the maximum allowed number of events (configurable, default 15)
2. Adding an event to a batch would exceed the max payload size in bytes (not configurable, ~500 KB)
3. The amount of time since the batch was created exceeds the max wait time (configurable, default 1000 ms)

The plugin methods (track et. al.) resolve once the associated event has been sent to Segment or retries have been exhausted. This differs from the batching implementation in AnalyticsBrowser which instead resolves as soon as an event is enqueued. I decided to wait to resolve until the event was done being sent because it significantly simplifies the `EventQueue's` ability to keep track of in-flight events, which in turn makes it easier to support graceful shutdowns.

## Implementation details

The plugin is broken into 2 main components:
- `ContextBatch`
- `Publisher`

The `ContextBatch` ensures that events are batched properly. It won't allow events to be added if they exceed the event count or byte count.

The `Publisher` does the heavy lifting. It is responsible for receiving events, creating batches, and sending those batches to Segment.


packages/node/src/__tests__/graceful-shutdown-integration.test.ts
```diff
@@ -30,14 +34,14 @@ describe('Ability for users to exit without losing events', () => {
   }
 
   describe('drained emitted event', () => {
-    test('emits a drained event if only one event is dispatched', async () => {
+    it('emits a drained event if only one event is dispatched', async () => {
```
chrisradek, line 37: I changed `test` to `it` just because my IDE thought `test()` had to return a promise, so I had tons of red squiggles. Might be a transient type merge occurring, but `it` matches what we use in out other packages and fixed my issue :grin:

packages/node/src/__tests__/integration.test.ts
```diff
@@ -39,16 +39,19 @@ describe('Error handling', () => {
...
-  test('property validation errors are surfaced as thrown errors', async () => {
+  it('property validation errors are surfaced as thrown errors', async () => {
     const analytics = new AnalyticsNode({
       writeKey,
     })
     expect(() => analytics.track({} as any)).toThrowError(/event/i)
   })
 
-  test('http delivery errors are accessed through the emitter', (done) => {
+  it.skip('http delivery errors are accessed through the emitter', (done) => {
```
chrisradek, line 49: I skipped this for now - it looked like any error in the previous plugin would end up being treated as an `http delivery error`, so I wanted to confirm if the intent was for this just to apply to errors thrown by `fetch` and not non-200 status codes. @silesky

silesky:
Perf testing
```
 New SDK vs old SDK ->
 new: 7054.5, old: 6639.55, difference: 414.9499999999998 (Requests per second) 
 Diff performance: 6.25 %
 ```

packages/node/src/__tests__/integration.test.ts
silesky: `http delivery` represents any non ok response along with any thrown   errors (that can usually contain a response)

Initially, almost all errors will be http delivery, but we may have things other error types like validation in the future, so it's a good future-proof pattern to have, and it also lets you add type-safe specific metadata based on the code (which is the discriminant key).

silesky (approved):

PR #643, opened 2022-10-26
refactor get cdn to have smaller footprint

silesky:
got rid of some smells, like testing loadLegacySettings in tests.


danieljackins (approved):

packages/browser/src/browser/index.ts
```diff
@@ -79,11 +79,9 @@ export interface AnalyticsBrowserSettings extends AnalyticsSettings {
 
 export function loadLegacySettings(
```
chrisradek, line 80: I believe this a breaking change. I have seen some users actually use this function in their own code.

PR #647, opened 2022-10-28
add perf tests

silesky:
`yarn turbo run --filter=node-integration-tests test`

This PR adds load testing with autocannon as part of CI. It verifies that there are no performance regressions relative to the old analytics-node SDK, and allows the configuration of an arbitrary threshold that also tests overall overhead.

The test harness logs a human-readable report that looks like this:
```
SDK Overhead 
 with: 4611, without: 14039.64, difference: -9428.64 (Requests per second) 
 Diff performance: -67.16 %
-------------------------------------------------------
New vs Old SDK 
 new: 4611, old: 6542.28, difference: -1931.2799999999997 (Requests per second) 
 Diff performance: -29.52 %
 ```

Note: The threshold is set quite high at the moment the batching PR has not been merged in. The batching PR has been performance tested, and improves performance by 50% by making fewer network requests, so it will be marginally faster than the original SDK.


pooyaj (approved):
lfg 🚢

chrisradek (commented):
Whoo! Thanks for adding this!

chrisradek (approved):

PR #655, opened 2022-11-04
bug/analytics reset no longer clears anonoymousId

silesky:
There was a race condition where the segment "normalize" plugin would asyncronously instantiate / overwrite the anonymousId field in browser storage right before the http request, thus making it impossible to call analytics.reset() after a typical method call. (for example, in a .track call). Moved this to the EventFactory so at the very least this browser storage action happens syncronously.

The only behavioral change is that if someone mutates event.anonymousId in a plug-in, it's not going to also implicitly update the global user anonymousId when that event gets sent.

this seems... fine to me? I wish we could go a step further and force everyone to call "setAnonymousId" explicitly if they wanted to override the global anonymousId for all events.


packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -322,6 +322,45 @@ describe('Event Factory', () => {
...
+
+      test('if two different anonymousIds are passed, should use one on the event', () => {
+        const track = factory.track('Order Completed', shoes, {
+          anonymousId: 'bar',
+          context: {
+            anonymousId: 'foo',
+          },
+        })
+        expect(track.context?.anonymousId).toBe('bar')
+        expect(track.anonymousId).toBe('bar')
```
silesky, line 351: Failure!

packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -322,6 +322,45 @@ describe('Event Factory', () => {
...
+      })
+
+      test('should set an anonymousId passed from the context on the event', () => {
```
silesky, line 354: Failure

packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -322,6 +322,45 @@ describe('Event Factory', () => {
...
+        const track = factory.track('Order Completed', shoes, {
+          context: {
+            anonymousId: 'foo',
+          },
+        })
+        expect(track.context?.anonymousId).toBe('foo')
+        expect(track.anonymousId).toBe('foo')
```
silesky, line 361: Failure

packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -322,6 +322,45 @@ describe('Event Factory', () => {
...
       })
     })
+
+    describe('anonymousId', () => {
+      test('accepts an anonymousId', () => {
+        const track = factory.track('Order Completed', shoes, {
+          anonymousId: 'foo',
+        })
+        expect(track.anonymousId).toBe('foo')
+        expect(track.context?.anonymousId).toBe('foo')
```
silesky, line 332: Failure

packages/browser/src/core/analytics/__tests__/integration.test.ts
```diff
@@ -224,6 +224,19 @@ describe('Analytics', () => {
...
+      const analytics = new Analytics({ writeKey: '' })
+      const getAnonId = () =>
+        retrieveStoredData({
+          cookieNames: ['ajs_anonymous_id'],
+          localStorageKeys: ['ajs_anonymous_id'],
+        }).ajs_anonymous_id
+      void analytics.track('foo')
+      expect(getAnonId()).toBeTruthy()
+      analytics.reset()
+      expect(getAnonId()).toBeFalsy()
```
chrisradek, line 237: Since the race condition was that the `track` call would eventually set anonId again, I think checking that it is falsey after the track call completes would cover it (and fail without these changes)

silesky: I actually moved the test to a different file, tested both failure and success condition!

packages/browser/src/test-helpers/browser-storage.ts
```diff
@@ -0,0 +1,70 @@
+import cookie from 'js-cookie'
```
silesky, line 1: Refactored to make type safe

packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -322,6 +322,48 @@ describe('Event Factory', () => {
         innerProp: '👻',
       })
     })
+
+    describe.skip('anonymousId', () => {
```
chrisradek, line 326: These tests currently don't work? Are they regressions or existing issues?

silesky: Not regressions, they are just existing edge cases / undefined behavior that I uncovered when working on this bug.

chrisradek (approved):
:fire:

packages/browser/src/core/events/index.ts
```diff
@@ -205,6 +205,15 @@ export class EventFactory {
   }
 
   public normalize(event: SegmentEvent): SegmentEvent {
+    const anonymousIdOverride = event.options?.anonymousId
```
pooyaj, line 208: maybe this is more readble?  
```
event.options?.anonymousId && this.user.anonymousId(event.options.anonymousId)
```
as `event.anonymousId = id` seems to be a noop and happens regardless of this line of code.

pooyaj (approved):
looks great!

packages/browser/src/core/events/index.ts
silesky: Good catch!

PR #659, opened 2022-11-06
node: "maxWaitTimeInMs" -> "flushInterval" (and update default value)

silesky:
- renames `maxWaitTimeInMs` -> `flushInterval` (easy migration from old SDK + in googling, idiomatic around queue flushing)
- change `flushInterval` default from 1 sec to old SDK default of 10 seconds (that was the old default -- 5 secs might be better?)

### DIFF
--- a/packages/node/src/app/analytics-node.ts
+++ b/packages/node/src/app/analytics-node.ts
@@ -91,7 +91,7 @@ export class Analytics
         path: settings.path,
         maxRetries: settings.maxRetries ?? 3,
         maxEventsInBatch: settings.maxEventsInBatch ?? 15,
-        flushInterval: settings.flushInterval ?? 1000,
+        flushInterval: settings.flushInterval ?? 10000,
       })
     ).then(() => undefined)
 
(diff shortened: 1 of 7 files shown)


arielsilvestri (approved):
I think 10 seconds is a good amount of time to account for different levels of latency, 5 seconds might be too short.

packages/node/src/app/analytics-node.ts
```diff
@@ -91,7 +91,7 @@ export class Analytics
         path: settings.path,
         maxAttempts: settings.maxAttempts ?? 4,
         maxEventsInBatch: settings.maxEventsInBatch ?? 15,
-        maxWaitTimeInMs: settings.maxWaitTimeInMs ?? 1000,
+        flushInterval: settings.flushInterval ?? 1000,
```
chrisradek, line 94: It doesn't look like the default value changed.

silesky: We can talk offline! I was thinking about changing the default to something in the middle...

PR #660, opened 2022-11-06
node rename max attempts

silesky:
Describing flush in terms of "retries" is more common/natural when it comes to SDK / client configuration.


arielsilvestri (approved):
LGTM!

chrisradek (approved):

PR #661, opened 2022-11-06
rename examples package, simplify changeset config and pipeline

silesky:
- clean up examples
- ignore internal packages in changesets


arielsilvestri (approved):
Working for me!

chrisradek (approved):
Is the main benefit that we won't have to keep updating config as we add examples/internal packages?

silesky:
> Is the main benefit that we won't have to keep updating config as we add examples/internal packages?

correct!

PR #666, opened 2022-11-09
Update LocalStorage error handling

### LINKED ISSUE #631: Localstorage throws on Private Session Safari iOS
Got an error delivered through Sentry.io just now that seemed strange.

image

image

A bit of googling led me to the following StackOverflow page https://stackoverflow.com/a/66704807/11889925 which suggests that this was probably a private session in which localStorage is unavailable and throws an error on getItem or setItem. 

Checking `isLocalStorageSupported()` would seemingly fix this issue 👍

ryder-wendt:
This PR is created to fix #631

LocalStorage was reported as becoming unavailable during an IOS private safari session. While unable to reproduce, try/catch have been added enclosing the line where the error was thrown with the intention to catch the error before surfacing on downstream tools. 

### Testing
---
[x] Minor unit test added

### DIFF
--- a/.changeset/itchy-points-flash.md
+++ b/.changeset/itchy-points-flash.md
@@ -2,4 +2,4 @@
 '@segment/analytics-next': patch
 ---
 
-Update LocalStorage expected behavior
+Do not throw errors if localStorage becomes unavailable


packages/browser/src/core/user/index.ts
```diff
@@ -171,7 +176,11 @@ export class LocalStorage extends Store {
   }
 
   remove(key: string): void {
-    return localStorage.removeItem(key)
+    try {
+      return localStorage.removeItem(key)
+    } catch (err) {
+      console.warn(err)
```
chrisradek, line 182: I think we should match the warnings we have for `set` and `get` here as well.

packages/browser/src/core/user/__tests__/index.test.ts
```diff
@@ -871,14 +871,19 @@ describe('store', function () {
   })
 
   describe('#get', function () {
-    it('should not not get an empty record', function () {
-      expect(store.get('abc') === undefined)
```
chrisradek, line 875: Interesting, I wonder how this test was passing before when we should have been returning a value or null.

silesky: `expect(store.get('abc') === undefined)` will always pass (because it lacks an assertion). It's basically `expect(true)`

chrisradek: Ah of course, that makes sense.

packages/browser/src/core/user/__tests__/index.test.ts
```diff
@@ -871,14 +871,19 @@ describe('store', function () {
...
 
     it('should get an existing record', function () {
       store.set('x', { a: 'b' })
       expect(store.get('x')).toStrictEqual({ a: 'b' })
     })
+
+    it('should return null values', function () {
+      store.set('y', { a: null })
+      expect(store.get('y')).toStrictEqual({ a: null })
+    })
```
silesky, line 886: I would also do an assertion around if I set a value to:
"foo"
""
"null"
undefined
null
etc

packages/browser/src/core/user/index.ts
silesky: ^ to follow up, we probably want to create a function so we don't repeat ourselves and avoid duplicate strings -- they take up precious bytes.

packages/browser/src/core/user/index.ts
```diff
@@ -149,15 +149,20 @@ export class LocalStorage extends Store {
   }
 
   get<T>(key: string): T | null {
-    const val = localStorage.getItem(key)
-    if (val) {
+    try {
+      const val = localStorage.getItem(key)
+      if (val === null) {
```
silesky, line 154: I'd also check for undefined here

packages/browser/src/core/user/__tests__/index.test.ts
ryder-wendt: I was able to cover all the values but having some trouble with making localStorage.getItem unavailable. 
I'm only able to throw by messing with `Storage.prototype` but it has to be placed at bottom of test file, doesn't feel like best practice.

``` describe('localStorage catch', function () {
  const store = new LocalStorage()
  beforeEach(function () {
    clear()
  })

  it('should catch without localStorage.getItem', function () {
    store.set('x', 'test')
    Storage.prototype.getItem = null as any
    expect(store.get('x')).toBe(null)
  })
})

Comment (silesky) [reply packages/browser/src/core/user/__tests__/index.test.ts:886]: Use jest [spies](https://jestjs.io/docs/jest-object#jestspyonobject-methodname) to mock global methods.
````  
   it('should throw an error', function () {
      const getItemSpy = jest
        .spyOn(global.Storage.prototype, 'getItem')
        .mockImplementation(() => {
          throw new Error('getItem fail.')
        })
      store.set('foo', 'some value')
      expect(store.get('foo')).toBeNull()
      expect(getItemSpy).toBeCalled()
    })
```

packages/browser/src/core/user/__tests__/index.test.ts
```diff
@@ -865,19 +851,36 @@ describe('group', () => {
...
 describe('store', function () {
-  const store = new LocalStorage()
-  beforeEach(function () {
-    clear()
-  })
-
   describe('#get', function () {
-    it('should not not get an empty record', function () {
-      expect(store.get('abc') === undefined)
+    it('should throw an error', function () {
```
silesky, line 855: "should return null if localStorage throws" would be an improved description

ryder-wendt: Agreed!

silesky:
@ryder-wendt this PR should have a changeset (minor version bump).

.changeset/itchy-points-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@segment/analytics-next': patch
+---
+
+Update LocalStorage expected behavior
```
silesky, line 5: A changeset can be as long as you want -- it should be nicer than a commit, as it will go in the changelog. I would say "Try not to throw errors if localStorage is not available".

ryder-wendt: This makes sense, I was using @1.46.0 as a reference but now looking at all changeset's, they do appear to be more descriptive.

silesky (approved):

chrisradek (approved):

PR #667, opened 2022-11-09
update CDN upload to always occur on merges to master

silesky:
This PR attempts to address unexpected behavior per conversations with @pooyaj  and @danieljackins. Currently, https://cdn.segment.com/analytics-next/br/master/latest/standalone.js _only_ tracks npm releases. 

Work
-  our cdn's master folder gets updated rather than only on an npm publish.
-  simplification of the CDN release step, so a publish failure doesn't mean that we fail to upload to the CDN. Thus, we treat master like any other branch with regards to uploading to CDN i.e. automatically upload to CDN  on all merges, 
not just npm publish.

If a customer wants to only track our npm releases, they are also free to use: https://unpkg.com/@segment/analytics-next/dist/umd/standalone.js


danieljackins (approved):
:shipit:

pooyaj (approved):

PR #669, opened 2022-11-10
Allow use of integrations directly from NPM

zikaari:
This patch allows customers to import classic integrations directly from NPM, and use those instead of them being loaded from CDN.

This can be useful in a case where one can't afford to have too many concurrent network requests _(due to limits imposed by browsers)_ , and would rather prefer few larger requests instead.

**API**

```ts
import GoogleAnalyticsIntegration from '@segment/analytics.js-integration-google-analytics'

// the following rig assumes configuration for Google Analytics will be available in the fetched settings
const analytics = AnalyticsBrowser.load({
	writeKey: '<WRITE_KEY>',
	classicIntegrations: [ GoogleAnalyticsIntegration ]
}),

// if remote settings don't include the integration configuration, it must be supplied directly like so:
const analytics = AnalyticsBrowser.load({
	writeKey: '<WRITE_KEY>',
	classicIntegrations: [ GoogleAnalyticsIntegration ]
}, {
	integrations: {
		'Google Analytics': {
			// all the necessary must be set here
		}
	}
}),
```

### DIFF
--- a/packages/browser/src/browser/index.ts
+++ b/packages/browser/src/browser/index.ts
@@ -24,6 +24,7 @@ import {
 } from '../core/buffer'
 import { popSnippetWindowBuffer } from '../core/buffer/snippet'
 import { inspectorHost } from '../core/inspector'
+import { LegacyIntegrationSource } from '../plugins/ajs-destination/types'
 
 export interface LegacyIntegrationConfiguration {
   /* @deprecated - This does not indicate browser types anymore */
@@ -152,7 +153,8 @@ async function registerPlugins(
   analytics: Analytics,
   opts: InitOptions,
   options: InitOptions,
-  plugins: Plugin[]
+  plugins: Plugin[],
+  legacyIntegrationSources?: LegacyIntegrationSource[]
 ): Promise<Context> {
   const tsubMiddleware = hasTsubMiddleware(legacySettings)
     ? await import(
@@ -164,18 +166,20 @@ async function registerPlugins(
       })
     : undefined
 
-  const legacyDestinations = hasLegacyDestinations(legacySettings)
-    ? await import(
-        /* webpackChunkName: "ajs-destination" */ '../plugins/ajs-destination'
-      ).then((mod) => {
-        return mod.ajsDestinations(
-          legacySettings,
-          analytics.integrations,
-          opts,
-          tsubMiddleware
-        )
-      })
-    : []
+  const legacyDestinations =
+    hasLegacyDestinations(legacySettings) || legacyIntegrationSources
+      ? await import(
+          /* webpackChunkName: "ajs-destination" */ '../plugins/ajs-destination'
+        ).then((mod) => {
+          return mod.ajsDestinations(
+            legacySettings,
+            analytics.integrations,
+            opts,
+            tsubMiddleware,
+            legacyIntegrationSources
+          )
+        })
+      : []
 
   if (legacySettings.legacyVideoPluginsEnabled) {
     await import(
@@ -279,12 +283,21 @@ async function loadAnalytics(
   // needs to be flushed before plugins are registered
   flushPreBuffer(analytics, preInitBuffer)
 
+  const legacyIntegrationSources = plugins.filter(
+    (pluginOrIntegration) => typeof pluginOrIntegration === 'function'
+  ) as LegacyIntegrationSource[]
+
+  const simplePlugins = plugins.filter(
+    (pluginOrIntegration) => typeof pluginOrIntegration !== 'function'
+  ) as Plugin[]
+
   const ctx = await registerPlugins(
     legacySettings,
     analytics,
     opts,
     options,
-    plugins
+    simplePlugins,
+    legacyIntegrationSources
   )
 
   const search = window.location.search ?? ''
--- a/packages/browser/src/browser/index.ts
+++ b/packages/browser/src/browser/index.ts
@@ -154,7 +154,7 @@ async function registerPlugins(
   opts: InitOptions,
   options: InitOptions,
   plugins: Plugin[],
-  legacyIntegrationSources?: LegacyIntegrationSource[]
+  legacyIntegrationSources: LegacyIntegrationSource[]
 ): Promise<Context> {
   const tsubMiddleware = hasTsubMiddleware(legacySettings)
     ? await import(
@@ -167,7 +167,7 @@ async function registerPlugins(
     : undefined
 
   const legacyDestinations =
-    hasLegacyDestinations(legacySettings) || legacyIntegrationSources
+    hasLegacyDestinations(legacySettings) || legacyIntegrationSources.length > 0
       ? await import(
           /* webpackChunkName: "ajs-destination" */ '../plugins/ajs-destination'
         ).then((mod) => {
@@ -278,26 +278,19 @@ async function loadAnalytics(
   inspectorHost.attach?.(analytics as any)
 
   const plugins = settings.plugins ?? []
+  const classicIntegrations = settings.classicIntegrations ?? []
   Context.initMetrics(legacySettings.metrics)
 
   // needs to be flushed before plugins are registered
   flushPreBuffer(analytics, preInitBuffer)
-
-  const legacyIntegrationSources = plugins.filter(
-    (pluginOrIntegration) => typeof pluginOrIntegration === 'function'
-  ) as LegacyIntegrationSource[]
-
-  const simplePlugins = plugins.filter(
-    (pluginOrIntegration) => typeof pluginOrIntegration !== 'function'
-  ) as Plugin[]
 
   const ctx = await registerPlugins(
     legacySettings,
     analytics,
     opts,
     options,
-    simplePlugins,
-    legacyIntegrationSources
+    plugins,
+    classicIntegrations
   )
 
   const search = window.location.search ?? ''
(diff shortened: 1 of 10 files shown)


zikaari:
I would like to hear opinions on `plugins` being `(Plugin | LegacyIntegrationSource)[]`, should they be separated? For technical and/or semantic reasons?

silesky:
I like that you don't have to expose the adapter -- but on the negative side, the adapter is now extra code that will always be bundled -- where, if we forced it be imported, it wouldn't be included in the bundle unless it was being used /shrug

silesky:
Honestly, this PR looks really freaking good!

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -110,13 +118,19 @@ export class LegacyDestination implements Plugin {
...
     }
 
-    this.integration = await loadIntegration(
-      ctx,
-      analyticsInstance,
-      this.name,
-      this.version,
+    const integrationSource =
+      this.integrationSource ??
+      (await loadIntegration(
```
silesky, line 123: we might want to rename "loadIntegration" to something like "loadRemoteDestination" -- since load and build have overlapping meanings, and I could see this getting confusing.

packages/browser/src/plugins/ajs-destination/loader.ts
```diff
@@ -29,14 +33,37 @@ function recordLoadMetrics(fullPath: string, ctx: Context, name: string): void {
   }
 }
 
+export function buildIntegration(
```
silesky, line 36: I'd name this function "buildLegacyDestination"

packages/browser/src/plugins/ajs-destination/loader.ts
```diff
@@ -29,14 +33,37 @@ function recordLoadMetrics(fullPath: string, ctx: Context, name: string): void {
...
+    integrationCtr = integrationSource
+  }
+
+  const integration = new integrationCtr(integrationSettings)
+  integration.analytics = analyticsInstance
+
+  return integration
+}
+
 export async function loadIntegration(
```
silesky, line 61: This is a good opportunity  to rename "loadIntegration" to something like "loadRemoteDestination" -- since now we have multiple ways to load destinations. The word "build" vs "load" can have overlapping meanings.

packages/browser/src/browser/index.ts
```diff
@@ -279,12 +283,21 @@ async function loadAnalytics(
   // needs to be flushed before plugins are registered
   flushPreBuffer(analytics, preInitBuffer)
 
+  const legacyIntegrationSources = plugins.filter(
```
silesky, line 286: This means that we can't support a plugin "creator" function that injects an analytics instance... mmm, this makes me thinks that "integrations" should not be passed in the plugins key.

silesky:
I did some review that brought me to some reservations about the API, https://github.com/segmentio/analytics-next/pull/669#discussion_r1027363852 --

packages/browser/src/plugins/ajs-destination/loader.ts
silesky, line 36: This is a pretty nice refactor to break this up!

packages/browser/src/plugins/ajs-destination/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+  // We use `!== 'unbundled'` (versus `=== 'bundled'`) to be inclusive of
+  // destinations without a defined value for `bundlingStatus`
+  const deviceMode =
+    bundlingStatus !== 'unbundled' &&
+    (type === 'browser' || versionSettings?.componentTypes?.includes('browser'))
+
+  // checking for iterable is a quick fix we need in place to prevent
+  // errors showing Iterable as a failed destiantion. Ideally, we should
+  // fix the Iterable metadata instead, but that's a longer process.
+  return (deviceMode || name === 'Segment.io') && name !== 'Iterable'
```
chrisradek, line 18: `Segment.io` is its own plugin, so shouldn't be included as one of the ajs-destination plugins. 

The previous code had the follow check that filters out the `Segment.io` plugin:
```
if (name.startsWith('Segment')) {
  return
}
```

Suspect we can change this to `return (deviceMode && name !== 'Iterable')` and add a check to exclude `Segment` at the top of the function.

packages/browser/src/plugins/ajs-destination/loader.ts
```diff
@@ -101,8 +123,8 @@ export function resolveVersion(
   settings: LegacyIntegrationConfiguration
```
chrisradek, line 123: `settings` is optional now right? Can we update the signature to reflect that?

chrisradek (commented):
PR is looking great!

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -316,47 +337,49 @@ export function ajsDestinations(
...
-      if ((!deviceMode && name !== 'Segment.io') || name === 'Iterable') {
-        return
-      }
+  const installableIntegrations = new Set([
+    // Remotely configured installable integrations
+    ...Object.entries(remoteIntegrationsConfig)
+      .filter(([name, integrationSettings]) =>
+        isInstallableIntegration(name, integrationSettings)
+      )
+      .map(([name]) => name),
```
arielsilvestri, line 362: Minor optimization here: might be able to write this as a `reduce` to only have one scan occur:

```javascript
...Object.entries(remoteIntegrationsConfig).reduce((accum, [name, integrationSettings]) =>
     isInstallableIntegration(name, integrationSettings) ? [...accum, name] : accum
  ,[])
```

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -316,47 +337,49 @@ export function ajsDestinations(
...
+    .map((name) => {
+      const integrationSettings = remoteIntegrationsConfig[name]
       const version = resolveVersion(integrationSettings)
       const destination = new LegacyDestination(
         name,
         version,
         integrationOptions[name],
-        options
+        options,
+        adhocIntegrationSources?.[name]
```
arielsilvestri, line 382: Possible same opportunity as above to use `reduce`

packages/browser/src/plugins/ajs-destination/utils.ts
zikaari: I suggest we do this in a separate pull request. Probably not a safe bet to add too many behavourial changes at once, especially when a behaviour has been there for a long time.

packages/browser/src/plugins/ajs-destination/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+export const isDisabledIntegration = (
+  integrationName: string,
+  globalIntegrations: Integrations
+) => {
+  const allDisableAndNotDefined =
+    globalIntegrations.All === false &&
+    globalIntegrations[integrationName] === undefined
+
+  return (
+    integrationName.startsWith('Segment') ||
```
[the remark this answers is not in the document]
zikaari: https://github.com/segmentio/analytics-next/pull/669/files/316f28152ddd76eab30d522c80d9cede525fc336#r1035384414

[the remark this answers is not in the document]
zikaari: https://github.com/segmentio/analytics-next/pull/669#discussion_r1035384414

packages/browser/src/plugins/ajs-destination/utils.ts
chrisradek: What you have is a behavioral change though. The code is now saying `Segment.io` is an installable integration. Previously, no `Segment` plugin would ever make it to this line because we did an early `startsWith('Segment')` check.

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: I'd skip it here, just for a "linear" top-down readability sake

packages/browser/src/plugins/ajs-destination/utils.ts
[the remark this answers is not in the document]
chrisradek: To me, it makes more sense to have the `Segment` check in `isInstallableIntegrations` since `Segment` is not a classic integration. This function seems to imply you're simply disabling integrations.

True the end result is the same but makes `isInstallableIntegrations` more truthy.

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: Actually, this is even cleaner, and matches the block right after this one:

```ts
    ...Object.keys(remoteIntegrationsConfig).filter((name) =>
      isInstallableIntegration(name, remoteIntegrationsConfig[name])
    ),
```

packages/browser/src/plugins/ajs-destination/utils.ts
zikaari: I see, just checked. Since there was an early return does this mean that [this ](https://github.com/segmentio/analytics-next/blob/222d4ecd4db7376839aa816b4b8923cae4f1487c/packages/browser/src/plugins/ajs-destination/index.ts#L351)has been redundant all along?

chrisradek: Yeah, the Segment.io check was always redundant.

chrisradek (approved):
Sweet, looks good to me!

silesky (approved):

PR #670, opened 2022-11-10
Change node messageId, and update core in order to remove transient dependencies from nodejs

silesky:
- Allow consumers to inject custom messageId into EventFactory, allowing us to remove node transient dependency on md5 library.

- node messageId looks like: `node-next-1668208232027-743be593-7789-4b74-8078-cbcc8894c586` ("node-next-[date]-[uuidv4]")
  - notice that there is no final hashing of this messageId -- [per stackoverflow,](https://stackoverflow.com/questions/36369506/is-using-the-java-util-uuid-after-hashing-it-with-md5-a-good-option) hashing a UUID (or any unique value in general) is a no-no as it increases the chances of collisions (and adds needless cpu overhead). It seems like a smell we should avoid if we can. Open to more discussion -- cc @pooyaj

### DIFF
--- a/packages/node/src/lib/__tests__/get-message-id.test.ts
+++ b/packages/node/src/lib/__tests__/get-message-id.test.ts
@@ -1,0 +1,21 @@
+import { getMessageId } from '../get-message-id'
+
+// https://gist.github.com/johnelliott/cf77003f72f889abbc3f32785fa3df8d
+const uuidv4Regex =
+  /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i
+
+describe(getMessageId, () => {
+  it('creates a messageId that may look like "node-next-166820896773a9-c7404832-feea-4e70-ac7f-2dc94d96acca"', () => {
+    const msg = getMessageId().split('-')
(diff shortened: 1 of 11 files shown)


pooyaj (approved):

packages/node/src/lib/__tests__/get-message-id.test.ts
```diff
@@ -0,0 +1,21 @@
+import { createMessageId } from '../get-message-id'
+
+// https://gist.github.com/johnelliott/cf77003f72f889abbc3f32785fa3df8d
+const uuidv4Regex =
+  /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i
+
+describe(createMessageId, () => {
+  it('creates a messageId that may look like "node-next-166820896773a9-c7404832-feea-4e70-ac7f-2dc94d96acca"', () => {
```
chrisradek, line 8: I think the messageId in this example is incorrect - Date.now() and the first part of the uuid look smoosed.

chrisradek (approved):
Overall looks good - just looks like 1 test case has an incorrect example in the description.

chrisradek (approved):

PR #678, opened 2022-11-16
show actual package version in node request body

silesky:


packages/node/src/__tests__/http-integration.test.ts
```diff
@@ -50,6 +51,9 @@ describe('Analytics Node', () => {
     `
     )
     const body = JSON.parse(httpRes.body)
+    const event = body.batch[0]
+    expect(event.context.library.version).toBe(version)
```
chrisradek, line 55: Nit: This could be part of line 67 since that's doing a version check.

packages/node/src/plugins/segmentio/index.ts
```diff
@@ -1,9 +1,10 @@
 import { CoreContext, CorePlugin } from '@segment/analytics-core'
 import { Publisher, PublisherProps } from './publisher'
+import { version } from '../../../package.json'
```
chrisradek, line 3: Does this end up changing the structure of our dist build in a way that breaks out imports or is this already covered?

silesky: correct, it was already covered by a previous import. If the structure changes, we'll also see an integration test failure.

chrisradek (approved):

packages/node/src/__tests__/http-integration.test.ts
silesky: Hmm, adding expressions to the inline-snapshot itself works -- feels weird, but sure!

silesky: Ooh, actually you can't do that... =(

PR #679, opened 2022-11-16
fall back to node-fetch only if native fetch is unsupported

silesky:


packages/node/src/lib/fetch.ts
```diff
@@ -0,0 +1,6 @@
+import { default as nodeFetch } from 'node-fetch'
+
+// made this a default export so it mocks the same as node-fetch in tests.
+const fetch = globalThis.fetch || nodeFetch
```
chrisradek, line 4: This looks fine - the type of `fetch` would end up being the union of these 2 right? That would be ideal so we don't accidentally use node-specific functionality (like node streams)

chrisradek (approved):

packages/node/src/lib/fetch.ts
silesky: You would think, but they actually overlap (node-fetch does not seem to be polluting the global space):
image

As you pointed out, node-fetch does have differences, but realistically we're not going to encounter them. If we expose a public API (e.g. to allow passing in a custom client), we'll create a simple interface that's not tightly coupled to a concrete fetch implementation.

PR #681, opened 2022-11-16
refactor node tests

silesky:


packages/node/src/__tests__/http-integration.test.ts
```diff
@@ -73,12 +70,12 @@ describe('Analytics Node', () => {
             "context": Object {
               "library": Object {
                 "name": "AnalyticsNode",
-                "version": "${version}",
```
silesky, line 76: This trick does not work, as it gets dynamically re-written on test runs

danieljackins (approved):

PR #682, opened 2022-11-16
do not trigger a retry on thrown errors in before plugins

silesky:
I feel like this retry queue ("before-plugin only!") behavior is something we want to eventually deprecate in the browser. IMO, we shouldn't bring it over.


silesky:
cc @chrisradek  @pooyaj

chrisradek (approved):
Honestly I thought we had gotten rid of retries at the EventQueue level in node.js.

PR #683, opened 2022-11-17
Adds integration test for action destinations in addDestinationMiddleware

arielsilvestri:
Per the work done in https://github.com/segmentio/analytics-next/pull/597, there was a question of whether or not the existing `addDestinationMiddleware` function would support action destinations. It appears that the logic that used to only allow `LegacyDestinations` into the plugins array was removed at some point, and this small addition to the existing integration test for the existing function confirms that action destinations are not filtered out if `addDestinationMiddleware` is invoked.

### DIFF
--- a/packages/browser/src/browser/__tests__/integration.test.ts
+++ b/packages/browser/src/browser/__tests__/integration.test.ts
@@ -671,18 +671,9 @@ describe('addDestinationMiddleware', () => {
   })
 
   it('supports registering destination middlewares', async () => {
-    const testPlugin: Plugin = {
-      name: 'test',
-      type: 'destination',
-      version: '0.1.0',
-      load: () => Promise.resolve(),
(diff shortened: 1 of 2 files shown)


packages/browser/src/browser/__tests__/integration.test.ts
```diff
@@ -670,10 +671,19 @@ describe('addDestinationMiddleware', () => {
   })
 
   it('supports registering destination middlewares', async () => {
```
silesky, line 673: Might makes sense to break this into a specific test case instead of overloading this test case with this scenario -- 
otherwise, it's easy for this test case // bug to get refactored away. 

Something like: "addDestinationMiddleware should include action destinations" etc. I've seen some test cases that actually include words like 'this tests a bug where XYZ'

PR #686, opened 2022-11-19
browser: refactor/remove asPromise

silesky:
This `asPromise` method seems to add  unneccessary complexity both in the way it's used and by virtue of its existence. If there's a galaxy brain reason for why we'd do this, I can't find it.

A util method that's just an alias to Promise.resolve?... like whyyyyy 😄  And there are actual unit tests for this method?


packages/browser/src/plugins/middleware/index.ts
```diff
@@ -39,26 +41,24 @@ export async function applyDestinationMiddleware(
...
-            returnedEvent = null
-          }
-
-          if (evt) {
-            returnedEvent = evt.obj
-          }
-        },
-      })
-    )
+    await fn({
```
silesky, line 44: I assume that the reason you need to wrap `fn` in an asPromise wrapper was introduced is because `MiddlewareFunction` and `SourceMiddleware` could actually return a Promise<void> -- but maybe they didn't realize  you could have a return type of void | Promise<void>. And, they might not have understood that there's no requirement to "normalize" all values to a promise in order to await -- that you can just as easily await a regular value as a promise.

danieljackins (approved):

chrisradek (approved):
Sweet, love removing code!

Issue #649, opened 2022-11-03
Track function does not pick up Options


culda:
Hi guys, I'm not able to override the context object in the track payload. I've tried the below code among other things and the `context.page` in the payload remains unchanged, filled with the default values.

I'm not using any plugins. I just can't work out how to change the context. 
The properties object works

What am I doing wrong?

```
      void analytics?.track(eventName, properties, {
        context: {
          page: {
            path: '/custom',            
          },
          x: 5,
        },
      });
```

culda:
Found out what it was. 
`properties` was resolving to undefined, essentially making the function call
`track(event, undefined, options)` -> In this case, the options object does not get picked up. 
Passing an empty object instead of undefined does the trick. `track(event, {}, options)`

Is this by design or a bug?

silesky:
Thanks @culda, this looks like a bug.

PR #688, opened 2022-11-21
Undefined options bug

arielsilvestri:
Addressing #649, there was a gap in the `resolveArguments` logic that wouldn't let `options` be set if `properties` were `undefined`, as the original logic did not account for the case where `properties` wouldn't be defined.

### DIFF
--- a/packages/browser/src/core/arguments-resolver/index.ts
+++ b/packages/browser/src/core/arguments-resolver/index.ts
@@ -38,7 +38,7 @@ export function resolveArguments(
     : {}
 
   let opts: Options = {}
-  if (isPlainObject(properties) && !isFunction(options)) {
+  if (!isFunction(options)) {
     opts = options ?? {}
   }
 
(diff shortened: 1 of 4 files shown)


packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -156,6 +156,14 @@ describe('Event Factory', () => {
       expect(track.context).toEqual({ opt1: true })
     })
 
+    test('sets context correctly if property arg is undefined', () => {
```
arielsilvestri, line 159: The bug didn't make it this far per se, but I wanted to harden this test suite again as it was a possible pain point.

packages/browser/src/core/arguments-resolver/index.ts
```diff
@@ -38,7 +38,7 @@ export function resolveArguments(
     : {}
 
   let opts: Options = {}
-  if (isPlainObject(properties) && !isFunction(options)) {
+  if (!isFunction(options)) {
```
arielsilvestri, line 41: Originally had this as `isPlainObject(data)` but `data` will always be an object at this point so it felt unnecessary here.

silesky:
Tested, this appears to work! Less complicated than I expected! Thanks!

silesky (approved):

silesky (changes requested):
Add a changeset please!

PR #694, opened 2022-11-23
node: add context, timestamp, integration options

silesky:
Adding options to top level of events to make consistent with old SDK (and flat is better!).
- context
- integrations
- timestamp


arielsilvestri (approved):
LGTM!

PR #697, opened 2022-11-27
node: eliminate flush delay

silesky:
If closeAndFlush is called, we to wait for pending events to make it to the segment plugin before flushing, but not wait any additional time for the flushInterval to expire.

### DIFF
--- a/packages/node/src/plugins/segmentio/publisher.ts
+++ b/packages/node/src/plugins/segmentio/publisher.ts
@@ -166,8 +166,10 @@ export class Publisher {
     const events = batch.getEvents()
     const payload = JSON.stringify({ batch: events })
     const maxAttempts = this._maxRetries + 1
-
     let currentAttempt = 0
+    if (this._closeAndFlushPendingItemsCount) {
+      this._closeAndFlushPendingItemsCount -= batch.length
+    }
     while (currentAttempt < maxAttempts) {
       currentAttempt++
 
--- a/packages/node/src/plugins/segmentio/publisher.ts
+++ b/packages/node/src/plugins/segmentio/publisher.ts
@@ -123,35 +123,42 @@ export class Publisher {
             Success: Check if batch is full and send if it is.
             Failure: Event exceeds maximum size (it will never fit), fail the event.
     */
-
-    if (batch.tryAdd(pendingItem)) {
-      const isFull = batch.length === this._maxEventsInBatch
+    const addStatus = batch.tryAdd(pendingItem)
+    if (addStatus.success) {
       const isExpectingNoMoreItems =
         batch.length === this._closeAndFlushPendingItemsCount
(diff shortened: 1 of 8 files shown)


packages/node/src/plugins/segmentio/publisher.ts
```diff
@@ -96,15 +117,19 @@ export class Publisher {
...
         Failure: Assume event is too big to fit in current batch - send existing batch.
           Add an event to the new batch.
             Success: Check if batch is full and send if it is.
             Failure: Event exceeds maximum size (it will never fit), fail the event.
     */
 
     if (batch.tryAdd(pendingItem)) {
-      if (batch.length === this._maxEventsInBatch) {
+      const isFull = batch.length === this._maxEventsInBatch
+      const isExpectingNoMoreItems =
```
chrisradek, line 129: I think you could do the `isExpectingNoMoreItems` check outside of this if statement, then re-use it here:
> if (fallbackBatch.length === this._maxEventsInBatch) {

That way it adding the last event to the batch forces a new batch (because the size in bytes is too large to fit in the existing one), we can still send it quickly.

chrisradek (commented):
Overall this looks like a solid improvement! Just had some minor comments!

packages/node/src/plugins/segmentio/publisher.ts
```diff
@@ -96,37 +117,48 @@ export class Publisher {
...
-        reason: new Error(`Event exceeds maximum event size of 32 kb`),
-      })
-      return Promise.resolve(ctx)
-    } else {
-      if (fallbackBatch.length === this._maxEventsInBatch) {
+    const fbAddStatus = fallbackBatch.tryAdd(pendingItem)
+
+    if (fbAddStatus.success) {
+      const isExpectingNoMoreItems =
+        fallbackBatch.length === this._closeAndFlushPendingItemsCount
```
chrisradek, line 150: One thing I just thought of - wonder if we should decrement `_closeAndFlushPendingItemsCount` when we send a batch. Because if we had 10 events and only 5 fit in a batch, then we'll never hit the pending items count.

silesky: That's an interesting scenario where batch size exceeds the pending item count. I will write a test for it.

chrisradek: > That's an interesting scenario where batch size exceeds the pending item count.

Opposite actually - where max batch size is smaller than the pending item count. Imagine if batch size was 5 but you sent 6 events you'd always be expecting more currently :grin:

silesky: You missed my edit!

chrisradek (approved):
Whoo! Let's ship it!

PR #704, opened 2022-11-30
Emit Reset Events

pooyaj:
* Emitting a `reset` event when user calls `analytics.reset()`. This allows for external code to hook into the reset events and do further logic. 

- [X] I've included a changeset (psst. run `yarn changeset`. Read about changesets [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)).

### DIFF
--- a/.changeset/curvy-lamps-behave.md
+++ b/.changeset/curvy-lamps-behave.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': minor
+---
+
+Emit a `reset` event when analytics.reset() is called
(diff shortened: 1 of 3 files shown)


silesky:
Looks good. Seems like the event emitter is kind of an afterthought -- in node, the contract is very well defined and every event gets emitted on: https://github.com/segmentio/analytics-next/blob/03e6383ac6cc2411226887cbb8851ddbd03ae32d/packages/node/src/app/dispatch-emit.ts#L35

silesky (approved):

pooyaj:
Yeah, agreed. The browser version emits separately in each of the `track`/`identify`/etc. calls with different parameters passed as extra context.

PR #709, opened 2022-12-02
node: improve and document Context and Traits interface

silesky:
This PR is meant to improve the DX for typescript users who want to have hints about what kind of context or traits they may want to pass.

Customer issue: https://github.com/segmentio/analytics-node/issues/349#issuecomment-1241527417

```ts
analytics.track({ event: "foo", context: { ...... } })
```
Used this as the source of truth: https://segment.com/docs/connections/spec/common/#context

I did not remove any types.

### DIFF
--- a/packages/core/src/events/interfaces.ts
+++ b/packages/core/src/events/interfaces.ts
@@ -16,8 +16,6 @@ export type JSONValue = JSONPrimitive | JSONObject | JSONArray
 export type JSONObject = { [member: string]: JSONValue }
 export type JSONArray = JSONValue[]
 
-export type Traits = Record<string, any>
-
 export type EventProperties = Record<string, any>
 
 export type Integrations = {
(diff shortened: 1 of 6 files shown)


packages/core/src/events/interfaces.ts
```diff
@@ -38,71 +38,137 @@ export interface CoreOptions {
...
    */
   referrer?: {
-    btid?: string
-    urid?: string
+    type?: string
+    name?: string
+    url?: string
+    link?: string
+
+    btid?: string // TODO? where did this come from?
```
silesky, line 166: have no idea where this came from -- it's not in https://segment.com/docs/connections/spec/common/#context

packages/core/src/events/interfaces.ts
```diff
@@ -38,71 +38,137 @@ export interface CoreOptions {
...
+    urid?: string // TODO? where did this come from?
```
silesky, line 167: not documented in https://segment.com/docs/connections/spec/common/#context

packages/core/src/events/interfaces.ts
```diff
@@ -38,71 +38,136 @@ export interface CoreOptions {
...
+    country?: string
+    /**
+     * @example 40.2964197
+     */
+    latitude?: string
+    /**
+     * @example -76.9411617
+     */
+    longitude?: string
+    region?: string
```
silesky, line 84: This is listed in the docs, but there is no example of this.

chrisradek: Just my guess, but I'd see `region` as a more global friendly version of `state`. Province, State, whatever else you call the equivalent.

packages/core/src/events/interfaces.ts
```diff
@@ -28,64 +26,131 @@ export type Integrations = {
...
+ */
+export interface CoreExtraContext {
+  /**
+   * This is usually used to flag an .identify() call to just update the trait, rather than "last seen".
+   */
+  active?: boolean
+
+  /**
+   * Current user’s IP address.
+   * (node only, collected automatically in browser).
```
pooyaj, line 51: You can still override IP on the browser as well

packages/core/src/events/interfaces.ts
```diff
@@ -28,64 +26,131 @@ export type Integrations = {
...
+   */
+  ip?: string
+
+  /**
+   * 	Locale string for the current user, for example en-US.
+   *  (node only, collected automatically in browser).
```
pooyaj, line 57: similarly can override on browser as well

pooyaj (approved):
Looks great

packages/core/src/events/interfaces.ts
```diff
@@ -179,34 +249,129 @@ interface PlanEvent {
...
+   * Email address of a user/group.
+   */
+  email?: string
+
+  /**
+   * @example female
+   */
+  gender?: string
+
+  crossDomainId?: string // undocumented?
```
chrisradek, line 374: I think we can exclude this.

packages/core/src/utils/ts-helpers.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Remove Index Signature
+ */
+export type RemoveIndexSignature<T> = {
+  [K in keyof T as {} extends Record<K, 1> ? never : K]: T[K]
```
chrisradek, line 5: I do not fully understand how this works (why the `1`?) but that's really neat that it does.

chrisradek (approved):
Love these sorts of updates!

packages/core/src/utils/ts-helpers.ts
silesky: Us both!

PR #710, opened 2022-12-02
Fix actions loading

chrisradek:
This PR fixes an issue that occurs when explicitly enabling/disabling action destinations.

A very small handful of action destinations have a different plugin name than their creation name. Amplitude is one of these.
Our current remote loader implementation only takes into account the plugin's name and not the creation name when deciding whether to load the destination.

That means there are 2 scenarios where loading Amplitude would not do the expected thing.
1. `analytics.load(writeKey, { integrations: { All: false, "Actions Amplitude": true }})`
With this scenario, all destinations are disabled by default and only if one is explicitly enabled will it be loaded. In this specific example, the plugin name is `Amplitude (Actions)` despite the integration (creationName) being `Actions Amplitude`. This mismatch caused the plugin to fail to load.

2. `analytics.load(writeKey, { integrations: { "Actions Amplitude": false }})`
With this scenario, the destination is explicitly disabled. However without these changes users would have had to specify `Amplitude (Actions)` instead.

### DIFF
--- a/.changeset/breezy-kids-clean.md
+++ b/.changeset/breezy-kids-clean.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': patch
+---
+
+Fixes an issue impacting a small number of destinations where explicitly enabling or disabling an integration on load would not work as expected.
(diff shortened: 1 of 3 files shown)


pooyaj (approved):
💯

PR #714, opened 2022-12-07
improve Segment.io retries

chrisradek:
This PR improves retry handling for the `Segment.io` plugin. It also fixes an issue where `PersistedPriorityQueue` was no longer storing events in `localStorage` on `pagehide`.

### Fix `PersistedPriorityQueue`

The PersistedPriorityQueue listens for the `pagehide` event then stores all of its events in localStorage. Currently, it listens for the `pagehide` event to be emitted on `document`, but `pagehide` is a `Window` event. This caused the queue to stop working with `pagehide`.

### Apply backoff to segment retries

Currently, when fetch errors are encountered events are pushed back onto the buffer via `buffer.push(ctx)`. This does not apply any backoff, so this is updated to `buffer.pushWithBackoff(ctx)`. This causes the event to be part of the buffer's `todo` length. The `schedule-flush` function already checks `buffer.todo > 0`, so we're covered here.

### Store canceled events

In the scenario where the user navigates away from the page between while there are pending HTTP requests, those events are not currently persisted in localStorage. This change keeps track of events that are 'in-flight' and adds them to the persisted priority queue. 

## Testing

I was able to add tests for this using playwright. Also tested manually with Chrome and Firefox.

### DIFF
--- a/packages/browser/src/lib/on-page-leave.ts
+++ b/packages/browser/src/lib/on-page-leave.ts
@@ -10,13 +10,13 @@
 export const onPageLeave = (cb: (...args: unknown[]) => void) => {
   let unloaded = false // prevents double firing if both are supported
 
-  // using document instead of window because of bug affecting browsers before safari 14 (detail in footnotes https://caniuse.com/?search=visibilitychange)
-  document.addEventListener('pagehide', () => {
+  window.addEventListener('pagehide', () => {
     if (unloaded) return
     unloaded = true
     cb()
   })
 
+  // using document instead of window because of bug affecting browsers before safari 14 (detail in footnotes https://caniuse.com/?search=visibilitychange)
   document.addEventListener('visibilitychange', () => {
     if (document.visibilityState == 'hidden') {
       if (unloaded) return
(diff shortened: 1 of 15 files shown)


packages/browser/src/lib/on-page-leave.ts
```diff
@@ -10,13 +10,13 @@
 export const onPageLeave = (cb: (...args: unknown[]) => void) => {
   let unloaded = false // prevents double firing if both are supported
 
-  // using document instead of window because of bug affecting browsers before safari 14 (detail in footnotes https://caniuse.com/?search=visibilitychange)
-  document.addEventListener('pagehide', () => {
+  window.addEventListener('pagehide', () => {
```
silesky, line 13: Good catch on this! All the browsers I used supported visibility change, so didn't even realize when manual testing.

packages/browser/src/plugins/segmentio/index.ts
```diff
@@ -115,5 +127,11 @@ export function segmentio(
     group: send,
   }
 
+  // Buffer may already have items if they were previously stored in localStorage.
+  // Start flushing them immediately.
+  if (buffer.todo) {
+    scheduleFlush(flushing, buffer, segmentio, scheduleFlush)
```
silesky, line 133: nice!

silesky (approved):

PR #716, opened 2022-12-07
Help analytics-node work with other runtimes

silesky:
- removed dependency on process.env.
- added a _metadata.runtime field.

### DIFF
--- a/packages/node/src/__tests__/http-integration.test.ts
+++ b/packages/node/src/__tests__/http-integration.test.ts
@@ -16,10 +16,7 @@ const snapshotMatchers = {
           version: expect.any(String),
         },
       },
-
-      _metadata: {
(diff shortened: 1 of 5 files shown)


chrisradek:
> removed dependency on process.env.

Sweet!

> added a _metadata.runtime field.

I'm not sure if any of our systems are looking at specific fields or not, were you seeing this used elsewhere?

silesky:
> I'm not sure if any of our systems are looking at specific fields or not, were you seeing this used elsewhere?
@chrisradek nope, my sense is that the __metadata field is just for our own tracing/debugging, and we can use as we please?

pooyaj (approved):
Looks great! I think the `_metadata` field should be fine to use here.

PR #718, opened 2022-12-07
Refactor inspector logic to use emitter

silesky:
In order to bring over shared logic like the EventQueue in from  analytics-core, we need to remove its browser-specific dependencies.


pooyaj (approved):
Love it 🤩 code is cleaner without inspectorHost calls peppered in

chrisradek (approved):
Looks great!

PR #721, opened 2022-12-08
Exposing Universal Storage on the Analytics object rather than passing it directly to plugins

pooyaj:
I think the old style where we pass down the storage as an argument to each plugin wasn't elegant. Exposing it via the analytics object have few benefits: 
- We don't have to change the way we register plugins
- We don't have to make changes to action destination runtime to pass storage as a new parameter, and hence changing docs, etc. 
- Makes storage more accessible for future use  


silesky (approved):
Thank you!

PR #722, opened 2022-12-12
Refactor node / browser to share Context, EventQueue,  attempt(), Stats, dispatch(), Plugins, and more

silesky:
This PR makes some important improvements and cleanups to our codebase, as well as some light refactoring and improving browser types. This moves us close to a single source of truth for node and browser.🔥 

> Note: I've avoided big file structure changes  in this PR to make it easy/possible to code review, so you'll see files that now just export code modules. Many of those will go away later.

### Change summary
-  update the browser methods to use the improved types in core (ExtraContext and Traits)
- adds much shared code
- Add proper type heirarchy for Plugin, Context. We do not want to expose core CoreContext, etc directly to customers, we want to create a derived class so each SDK can have different properties and implementations if needed.
- Remove duplicate event queue test
- Refactor dispatch to take a context instance. Move offline logic into private browser class.
-  Fix various test warnings


packages/core/src/queue/event-queue.ts
```diff
@@ -232,16 +223,24 @@ export class EventQueue extends Emitter<EventQueueEmitterContract> {
     return true
   }
 
-  private availableExtensions(denyList: Integrations): CorePluginsByType {
+  private availableExtensions(denyList: Integrations) {
     const available = this.plugins.filter((p) => {
       // Only filter out destination plugins or the Segment.io plugin
       if (p.type !== 'destination' && p.name !== 'Segment.io') {
         return true
       }
```
silesky, line 232: This code was recently added to the browser event queue by neek for the destination actions work, just moving it here.

packages/core/src/stats/index.ts
```diff
@@ -87,3 +92,13 @@ export default class Stats {
     })
   }
 }
+
```
silesky, line 95: Node doesn't currently use stats / metrics, but we might want to capture it in the future. Having this pattern lets us avoid using optional chaining, which decreases compilation size overall.

packages/core/src/stats/index.ts
```diff
@@ -87,3 +92,13 @@ export default class Stats {
...
+export class NullStats implements Stats {
```
silesky, line 96: Node doesn't currently use stats / metrics, but we might want to capture it in the future. Using this as the default instance when adding a context lets us avoid using optional chaining, which decreases compilation size overall.

packages/node-integration-tests/package.json
```diff
@@ -16,10 +16,8 @@
     "@segment/analytics-node": "workspace:^",
     "@types/analytics-node": "^3.1.9",
     "@types/autocannon": "^7",
-    "@types/express": "^4",
```
silesky, line 19: This is a dev dependency that is already used by other packages, so we might as well just share it.

packages/browser/src/core/analytics/index.ts
```diff
@@ -434,7 +411,9 @@ export class Analytics
     ) as LegacyDestination[]
 
     legacyDestinations.forEach((destination) => {
-      destination.addMiddleware(...middlewares)
+      if ('addMiddleware' in destination) {
```
silesky, line 414: This check technically might not be necessary, but it was revealed when better types were added.

packages/browser/src/core/context/index.ts
```diff
@@ -1,143 +1,26 @@
...
-  }
-
-  public setFailedDelivery(options: ContextFailedDelivery) {
-    this._failedDelivery = options
-  }
+let _remoteMetrics: RemoteMetrics | undefined
 
-  public logs(): LogMessage[] {
-    return this.logger.logs
+export class Context extends CoreContext<SegmentEvent> {
```
silesky, line 14: Context subclasses CoreContext now.

packages/browser/src/core/context/index.ts
```diff
@@ -1,143 +1,26 @@
...
+  static override system() {
+    return new this({ type: 'track', event: 'system' })
   }
-
-  public flush(): void {
-    this.logger.flush()
-    this.stats.flush()
+  static initRemoteMetrics(options?: MetricsOptions) {
```
silesky, line 18: simply preserving the previous 'initMetrics' static method

packages/browser/src/core/events/interfaces.ts
```diff
@@ -1,229 +1,35 @@
...
+export type {
+  Integrations,
+  Plan,
+  TrackPlan,
+  PlanEvent,
+  Callback,
+  JSONArray,
+  JSONValue,
+  JSONPrimitive,
+  JSONObject,
```
silesky, line 34: To make this PR reviewable, I'm try to avoid changing the file structure too much. Thus, I'm simply exporting a lot of types that were previously declared here. Elsewhere I do the same thing.

packages/core/src/priority-queue/index.ts
```diff
@@ -7,22 +6,20 @@ import { backoff } from './backoff'
...
+interface QueueItem {
   id: string
 }
 
-export class PriorityQueue<
-  Context extends WithID = CoreContext
-> extends Emitter {
-  protected future: Context[] = []
-  protected queue: Context[]
+export class PriorityQueue<Item extends QueueItem = QueueItem> extends Emitter {
```
silesky, line 13: There are no logic changes, just typescript changes to clarify that PriorityQueue is completely decoupled from the analytics domain, as this has proven to be confusing to a few people (me included).

danieljackins (approved):
Looks good to me! Thanks for the additional comments to help review.

pooyaj (approved):
💥

packages/browser/src/core/plugin/index.ts
```diff
@@ -1,36 +1,16 @@
...
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  options?: any
-  priority?: 'critical' | 'non-critical' // whether AJS should expect this plugin to be loaded before starting event delivery
-}
+import type { CorePlugin } from '@segment/analytics-core'
+import type { DestinationMiddlewareFunction } from '../../plugins/middleware'
+import type { Analytics } from '../analytics'
+import type { Context } from '../context'
 
 // enrichment - modifies the event. Enrichment can happen in parallel, by reducing all changes in the final event. Failures in this stage could halt event delivery.
```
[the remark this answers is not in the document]
silesky: Good point

packages/browser/src/core/stats/index.ts
```diff
@@ -1,89 +1,12 @@
...
-    return this.metrics.map((m) => {
-      return {
-        m: m.metric,
-        v: m.value,
-        t: m.tags,
-        k: compactMetricType(m.type),
-        e: m.timestamp,
-      }
-    })
+  override increment(metric: string, by?: number, tags?: string[]): void {
```
silesky, line 8: Remote metrics added via override.

packages/node/src/app/dispatch-emit.ts
```diff
@@ -1,27 +1,26 @@
...
-  queue: EventQueue,
+  event: SegmentEvent,
+  queue: NodeEventQueue,
   emitter: NodeEmitter,
   callback?: Callback
 ): Promise<void> => {
   try {
-    const ctx = await dispatch(event, queue, emitter, {
+    const context = new Context(event)
+    const ctx = await dispatch(context, queue, emitter, {
```
chrisradek, line 23: It looks like dispatch always took a context - now I'm wondering how this was working before!

packages/browser/src/core/inspector/index.ts
```diff
@@ -20,12 +20,14 @@ const inspectorHost: Partial<InspectBroker> = ((env as any)[
 export const attachInspector = (analytics: Analytics) => {
   inspectorHost.attach?.(analytics as any)
 
-  analytics.on('dispatch_start', (ctx) => inspectorHost.triggered?.(ctx))
+  analytics.on('dispatch_start', (ctx) => inspectorHost.triggered?.(ctx as any))
 
-  analytics.queue.on('message_enriched', (ctx) => inspectorHost.enriched?.(ctx))
+  analytics.queue.on('message_enriched', (ctx) =>
+    inspectorHost.enriched?.(ctx as any)
```
chrisradek, line 26: What changed that required these to be casted to `any` now?

packages/node/src/app/dispatch-emit.ts
silesky: Previously, the dispatch method actually instantiated the context object inside.

packages/browser/src/core/inspector/index.ts
silesky: Previously, ctx was `any` here. Now, the `queue` [emitter is typed](https://github.com/segmentio/analytics-next/blob/dfb712f822c769eaef27b403f5743d9a02f96007/packages/core/src/queue/event-queue.ts#L13).

None of the context types line up with those of inspector, because we [don't share our Context type with inspector](https://github.com/segmentio/segment-inspector/blob/a175e0b37eab25676b27e5395c131b7ec6d14c23/packages/webext/src/broker/index.ts#L26). 

@zikaari's set the precedent for doing the coercion rather than say adding any mapping logic, so I just followed that pattern.

packages/browser/src/core/stats/remote-metrics.ts
```diff
@@ -85,7 +89,7 @@ export class RemoteMetrics {
     }
 
     this.queue.push({
-      type: 'Counter',
+      type: 'counter',
```
chrisradek, line 92: I think this change might break some of our metrics.

packages/browser/src/core/plugin/index.ts
```diff
@@ -1,36 +1,12 @@
...
-  unload?: (ctx: Context, instance: Analytics) => Promise<unknown> | unknown
-
-  ready?: () => Promise<unknown>
-  track?: (ctx: Context) => Promise<Context> | Context
-  identify?: (ctx: Context) => Promise<Context> | Context
-  page?: (ctx: Context) => Promise<Context> | Context
-  group?: (ctx: Context) => Promise<Context> | Context
-  alias?: (ctx: Context) => Promise<Context> | Context
-  screen?: (ctx: Context) => Promise<Context> | Context
+export interface DestinationPlugin extends Plugin {
```
chrisradek, line 8: Is this type only used internally?

Since users can write their own plugins, making `addMiddleware` a required method would be breaking (and not sure we want to expose this functionality directly.) If it's just internally used then no concerns as long as we aren't casting user-plugins to this type.

packages/browser/src/core/inspector/index.ts
chrisradek: Ah gotcha, makes sense!

packages/browser/src/core/stats/remote-metrics.ts
silesky: oh yeah, I just thought of that =S I was going to check if our API required title case or not.

packages/browser/src/core/plugin/index.ts
silesky: This is only used internally

PR #727, opened 2022-12-15
refactor remote metrics

silesky:
some clean up!


packages/browser/src/core/stats/remote-metrics.ts
```diff
@@ -14,9 +14,37 @@ export interface MetricsOptions {
...
+    [key: string]: string
+  }
+}
+
+const createRemoteMetric = (
+  metric: string,
+  tags: string[],
+  versionType: 'web' | 'npm'
+): RemoteMetric => {
+  const formattedTags = tags.reduce((acc, t) => {
```
silesky, line 31: a modern way of writing this is `Object.fromEntries(["foo:bar", "baz:bill"].map(el => el.split(":")))` -- but dang browser support

packages/browser/src/core/stats/remote-metrics.ts
```diff
@@ -14,9 +14,37 @@ export interface MetricsOptions {
...
+    const [k, v] = t.split(':')
+    acc[k] = v
+    return acc
+  }, {} as Record<string, string>)
```
silesky, line 35: This could be condensed to `Object.fromEntries(tags.map(el => el.split(":")))` -- but dang browser support =/

arielsilvestri (approved):

PR #729, opened 2022-12-15
Add / improve offline integration tests

silesky:
I had to removed a couple offline tests in the previous large refactoring PR since the event queue is no longer responsible for that logic. I'm adding those scenarios back in integration.

### DIFF
--- a/packages/browser/src/browser/__tests__/integration.test.ts
+++ b/packages/browser/src/browser/__tests__/integration.test.ts
@@ -16,15 +16,17 @@ import jar from 'js-cookie'
 import { PriorityQueue } from '../../lib/priority-queue'
 import { getCDN, setGlobalCDNUrl } from '../../lib/parse-cdn'
 import { clearAjsBrowserStorage } from '../../test-helpers/browser-storage'
+import { parseFetchCall } from '../../test-helpers/fetch-parse'
 import { ActionDestination } from '@/plugins/remote-loader'
 
-let fetchCalls: Array<any>[] = []
+let fetchCalls: ReturnType<typeof parseFetchCall>[] = []
 
 jest.mock('unfetch', () => {
   return {
     __esModule: true,
     default: (url: RequestInfo, body?: RequestInit) => {
-      fetchCalls.push([url, body])
+      const call = parseFetchCall([url, body])
+      fetchCalls.push(call)
       return createMockFetchImplementation(cdnSettingsKitchenSink)(url, body)
     },
   }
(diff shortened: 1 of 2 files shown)


packages/browser/src/browser/__tests__/integration.test.ts
```diff
@@ -897,26 +899,56 @@ describe('retries', () => {
     expect(ajs.queue.queue.getAttempts(fruitBasketEvent)).toEqual(2)
   })
 
-  it('does not queue up events when offline if retryQueue setting is set to false', async () => {
-    const [ajs] = await AnalyticsBrowser.load({ writeKey })
+  it('does not queue events / dispatch when offline if retryQueue setting is set to false', async () => {
+    const [ajs] = await AnalyticsBrowser.load(
+      { writeKey },
+      { retryQueue: false }
```
chrisradek, line 905: Were we actually testing with retryQueue true before? Or is this making the test explicit?

packages/browser/src/browser/__tests__/integration.test.ts
```diff
@@ -897,26 +899,56 @@ describe('retries', () => {
...
+      {
+        ...testPlugin,
+        ready: () => Promise.resolve(true),
+        track: trackSpy,
+      },
+      ajs
+    )
+
+    // @ts-ignore ignore reassining function
+    isOffline = jest.fn().mockReturnValue(true)
```
chrisradek, line 945: Is this only returning `true` once? I'm wondering how come the plugin is called if `isOffline` is always true.

chrisradek (commented):
Changes look good! I'm just having a hard time understanding why something works!

packages/browser/src/browser/__tests__/integration.test.ts
silesky: Yes, previously we were testing retryQueue with true here (which is the default). It only passed because the logic of the test itself was broken.

packages/browser/src/browser/__tests__/integration.test.ts
silesky: The plugin will be called unless both isOffline = true _and_ retryQueue = false. So basically, it still goes through the plugins no matter what unless a user explicitly sets retryQueue to false.

chrisradek: Oh interesting, I always thought if you were offline it would just go immediately to the PersistedQueue.

chrisradek (approved):
:shipit:

Issue #735, opened 2022-12-22
Node: "window is not defined"


hatton:
Using 1.48.0 with this code in the main process of an Electron application:

```
import { Analytics, AnalyticsNode } from "@segment/analytics-next";
let analytics: Analytics | undefined;
export async function initializeAnalyticsMain() {
  const key = ...
  const [analytics, context] = await AnalyticsNode.load({ writeKey: key });
```

I get

```
(node:29264) UnhandledPromiseRejectionWarning: ReferenceError: window is not defined
    at module.exports../node_modules/@segment/analytics-next/dist/pkg/core/user/index.js.Cookie.available (C:\dev\lameta\app\main-bundle.js:5828:29)
    at getAvailableStorageOptions (C:\dev\lameta\app\main-bundle.js:6021:24)
    at new Analytics (C:\dev\lameta\app\main-bundle.js:3640:238)        
    at Function.<anonymous> (C:\dev\lameta\app\main-bundle.js:7125:37)  
    at step (C:\dev\lameta\app\main-bundle.js:8457:23)
```

Presumably `window` is never available in NodeJS. So am I calling something incorrectly?

hatton:
I'm closing and will start again.

PR #737, opened 2022-12-26
Check that window variable is available before using it to detect cookies

dnna:
Version 1.48.0 introduces a breaking change that did not exist in 1.47.0. It assumes that the window variable is available and uses it to access `window.navigator.cookieEnabled` to determine whether the cookie storage mechanism exists. However, on environments that do not have the window variable (like node), this causes it to crash, causing issues such as https://github.com/segmentio/analytics-next/issues/735. A minor version upgrade IMHO should not break existing implementations (which used to be part of the readme) or require people to refactor their codebase to import a different package (which also has slightly different method signatures, so it's not a search-and-replace refactor).

This PR attempts to fix the specific breakage by checking the existence of the window variable in `Cookie.available()` method and assuming that cookies are disabled if it does not exist.

### DIFF
--- a/packages/browser/src/core/user/index.ts
+++ b/packages/browser/src/core/user/index.ts
@@ -63,7 +63,10 @@ const ONE_YEAR = 365
 
 export class Cookie extends Store {
   static available(): boolean {
-    let cookieEnabled = window && window.navigator.cookieEnabled
+    if (typeof window === 'undefined') {
+      return false;
+    }
+    let cookieEnabled = window.navigator.cookieEnabled
 
     if (!cookieEnabled) {
       jar.set('ajs:cookies', 'test')
--- a/packages/browser/src/core/user/index.ts
+++ b/packages/browser/src/core/user/index.ts
@@ -63,7 +63,7 @@ const ONE_YEAR = 365
 
 export class Cookie extends Store {
   static available(): boolean {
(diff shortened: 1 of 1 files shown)


silesky:
Thank you @dnna. We certainly didn't intend to break legacy consumers of the legacy 'undocumented' analytics-node included here, so this is a good PR/ issue.


As a general recommendation -- if you're on the fence, I would migrate to the new analytics node (stable, though in beta) ASAP, as the one here will be deprecated shortly.

packages/browser/src/core/user/index.ts
```diff
@@ -63,7 +63,7 @@ const ONE_YEAR = 365
 
 export class Cookie extends Store {
   static available(): boolean {
-    let cookieEnabled = window.navigator.cookieEnabled
+    let cookieEnabled = window && window.navigator.cookieEnabled
```
silesky, line 66: Would be better to use the getGlobal() helper

 https://github.com/segmentio/analytics-next/blob/master/packages/core/src/utils/get-global.ts

dnna: Sorry for taking so long, we ended up migrating away from the legacy analytics-node so this was no longer a breaking problem for us, but I'm including the recommended fix here since others might still have it.

silesky:
This should be fixed by the Universal Storage addition.

PR #746, opened 2023-01-04
Make storage .set() consistent with browser storage API

silesky:
We want to be more / less consistent with the [browser storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem)


pooyaj (approved):
Looks great 🚢

PR #764, opened 2023-01-10
Add comprehensive message lifecycle events

zikaari:
Add `delivery_retry` lifecycle event and update `message_enriched` to include the Plugin that enriched the message

### DIFF
--- a/packages/browser/src/browser/__tests__/inspector.integration.test.ts
+++ b/packages/browser/src/browser/__tests__/inspector.integration.test.ts
@@ -55,7 +55,67 @@ describe('Inspector', () => {
 
     await deliveryPromise
 
-    expect(enrichedFn).toHaveBeenCalledTimes(1)
+    expect(enrichedFn).toHaveBeenCalledTimes(2)
     expect(deliveredFn).toHaveBeenCalledTimes(1)
+  })
+
+  it('emits message enrichment events with enricher information', async () => {
+    const [analytics] = await AnalyticsBrowser.load({
+      writeKey,
+    })
+
+    const enrichedFn = jest.fn()
+
+    analytics.queue.on('message_enriched', enrichedFn)
+
+    await analytics.register({
+      name: 'Test plugin',
+      version: '1.0.0',
+      type: 'before',
+      isLoaded: () => true,
+      load: () => Promise.resolve(),
+      track: (ctx) => ctx,
+    })
+
+    const ctx = await analytics.track('Test event').catch(() => {})
+
+    expect(enrichedFn).toHaveBeenLastCalledWith(
+      ctx,
+      expect.objectContaining({
+        name: 'Test plugin',
+        type: 'before',
+      })
+    )
+  })
+
+  it('emits expected events if before plugin throws error ', async () => {
+    const [analytics] = await AnalyticsBrowser.load(
+      {
+        writeKey,
+      },
+      { retryQueue: false }
+    )
+
+    const failedFn = jest.fn()
+    const deliveredFn = jest.fn()
+
+    analytics.queue.on('delivery_failure', failedFn)
+    analytics.queue.on('message_delivered', deliveredFn)
+
+    await analytics.register({
+      name: 'Faulty Plugin',
+      version: '1.0.0',
+      type: 'before',
+      isLoaded: () => true,
+      load: () => Promise.resolve(),
+      track: () => {
+        throw new Error()
+      },
+    })
+
+    await analytics.track('Faulty event')
+
+    expect(failedFn).toHaveBeenCalledTimes(1)
+    expect(deliveredFn).not.toHaveBeenCalled()
   })
 })
--- a/packages/core/src/queue/event-queue.ts
+++ b/packages/core/src/queue/event-queue.ts
@@ -12,8 +12,9 @@ import { isOffline } from '../connection'
 
 export type EventQueueEmitterContract<Ctx extends CoreContext> = {
   message_delivered: [ctx: Ctx]
-  message_enriched: [ctx: Ctx]
+  message_enriched: [ctx: Ctx, plugin: CorePlugin<Ctx>]
   delivery_success: [ctx: Ctx]
+  delivery_retry: [ctx: Ctx]
   delivery_failure: [ctx: Ctx, err: Ctx | Error | ContextCancelation]
   flush: [ctx: Ctx, delivered: boolean]
 }
@@ -265,6 +266,10 @@ export abstract class CoreEventQueue<
       throw new Error('Not ready')
     }
 
+    if (ctx.attempts > 1) {
+      this.emit('delivery_retry', ctx)
+    }
+
     const { before, enrichment } = this.availableExtensions(
       ctx.event.integrations ?? {}
     )
@@ -274,6 +279,8 @@ export abstract class CoreEventQueue<
       if (temp instanceof CoreContext) {
         ctx = temp
       }
+
+      this.emit('message_enriched', ctx, beforeWare)
     }
 
     for (const enrichmentWare of enrichment) {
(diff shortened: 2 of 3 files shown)


packages/core/src/queue/event-queue.ts
```diff
@@ -265,6 +266,10 @@ export abstract class CoreEventQueue<
       throw new Error('Not ready')
     }
 
+    if (ctx.attempts > 1) {
+      this.emit('delivery_retry', ctx)
```
chrisradek, line 270: How valuable is this event? Most (all?) of the delivery and segment.io plugins handle their own retries separate from the main event queue, and I think only `before` plugins/source middleware can trigger a pipeline retry.

zikaari: It exists just to make the Inspector trace look linear and chorological

image

It will go away when we fix retries once and for all

zikaari: At least for now, regardless of what real value the "retry" mechanism provides, just because it exists, we need to accommodate for it. Part of the reason that explains inclinations towards removing the "global retries"

packages/core/src/queue/event-queue.ts
```diff
@@ -274,16 +279,18 @@ export abstract class CoreEventQueue<
       if (temp instanceof CoreContext) {
         ctx = temp
       }
+
+      this.emit('message_enriched', ctx, beforeWare)
```
silesky, line 283: We probably want to get into the habit of writing some tests for these emitted events since we are moving into  treating them as a first class API.

packages/browser/src/browser/__tests__/inspector.integration.test.ts
```diff
@@ -55,7 +56,51 @@ describe('Inspector', () => {
...
+    const triggeredFn = jest.fn()
+    const failedFn = jest.fn()
+    const deliveredFn = jest.fn()
+
+    analytics.on('dispatch_start', triggeredFn)
+    analytics.queue.on('delivery_failure', failedFn)
+    analytics.queue.on('message_delivered', deliveredFn)
+
+    await analytics.register({
+      name: 'Faulty Plugin',
```
silesky, line 77: A typical "faulty" before plugin  would probably just throw an error, rather than a contextCancellation. I would probably add a test for it like so (setting retryQueue to false):
```ts
**
  it('emits message failure event', async () => {
    const [analytics] = await AnalyticsBrowser.load(
      {
        writeKey,
      },
      { retryQueue: false }
    )

    const triggeredFn = jest.fn()
    const failedFn = jest.fn()
    const deliveredFn = jest.fn()

    analytics.on('dispatch_start', triggeredFn)
    analytics.queue.on('delivery_failure', failedFn)
    analytics.queue.on('message_delivered', deliveredFn)

    await analytics.register({
      name: 'Faulty Plugin',
      version: '1.0.0',
      type: 'before',
      isLoaded: () => true,
      load: () => Promise.resolve(),
      track: () => {
        throw new Error('sup')
      },
    })

    await analytics.track('Faulty event')

    expect(triggeredFn).toHaveBeenCalledTimes(1)

    expect(triggeredFn).toHaveBeenCalledWith(
      expect.objectContaining({
        id: expect.any(String),
        event: expect.objectContaining({
          event: 'Faulty event',
          type: 'track',
        }),
      })
    )
    expect(failedFn).toHaveBeenCalledTimes(1)
    expect(deliveredFn).not.toHaveBeenCalled()
  })**
```

packages/browser/src/browser/__tests__/inspector.integration.test.ts
```diff
@@ -55,7 +56,51 @@ describe('Inspector', () => {
 
     await deliveryPromise
 
-    expect(enrichedFn).toHaveBeenCalledTimes(1)
+    expect(enrichedFn).toHaveBeenCalledTimes(2)
     expect(deliveredFn).toHaveBeenCalledTimes(1)
   })
+
+  it('emits message failure event', async () => {
```
silesky, line 63: Nitpick: Seems like this is testing a few things, we might want to change the name to "emits expected events" if we aren't breaking it down into subtests

packages/browser/src/browser/__tests__/inspector.integration.test.ts
```diff
@@ -1,5 +1,6 @@
+import { waitFor } from '@testing-library/dom'
```
silesky, line 1: I love this lib, but I feel like this helper (while useful) is a bit misleading here, as there's no relevant DOM manipulation going on in these tests.

silesky: I mentioned the alternative https://github.com/segmentio/analytics-next/pull/764#discussion_r1066397210

packages/browser/src/browser/__tests__/inspector.integration.test.ts
zikaari: Other things are not being tested for, they are there to act as an "experiment control"

packages/browser/src/browser/__tests__/inspector.integration.test.ts
zikaari: The pipeline is asynchronous

packages/browser/src/browser/__tests__/inspector.integration.test.ts
silesky: FYI, other tests in the suite already provide that "control" through their assertions, so my preference is to remove assertions that aren't directly relevant to the test, as they make the test less focused and more brittle. /my 2 cents

packages/browser/src/browser/__tests__/inspector.integration.test.ts
silesky: > The pipeline is asynchronous
If the ultimate goal is test are testing whether or not `failedFn` gets called and `deliveredFn` does not get called, I feel like they're the same, no (they have the same assertions)?

zikaari: It tests for exactly that - failed does get called and delivered does not

Both being triggered is a contradictory scenario

packages/browser/src/browser/__tests__/inspector.integration.test.ts
zikaari: It won't work. The returned promise can't be `await`ed because it never resolves, or throws.

packages/browser/src/browser/__tests__/inspector.integration.test.ts
zikaari: Ok. Removed

packages/browser/src/browser/__tests__/inspector.integration.test.ts
silesky: > The pipeline is asynchronous

I guess I'm not following!

packages/browser/src/browser/__tests__/inspector.integration.test.ts
silesky: > It won't work. The returned promise can't be awaited because it never resolves, or throws.

Did you try running the test? It _does pass_, I just checked.

packages/browser/src/browser/__tests__/inspector.integration.test.ts
zikaari: Oh man! My bad, I forgot to disable `retryQueue` 🤦‍♂️

Thanks!

packages/browser/src/browser/__tests__/inspector.integration.test.ts
zikaari: https://github.com/segmentio/analytics-next/pull/764#discussion_r1070112271

silesky (approved):

PR #765, opened 2023-01-11
Attempting to refactor schema filter functionality

arielsilvestri:
This is the point where the least amount of failing tests occur. None of the failures are surprising to me given the changes (you can pull and run tests if you'd like to see which are failing).

I've been trying to find a way to restore the intended schema filtering functionality throughout our application, and I'm wondering if we should be taking a different approach for this problem. 

For a brief description of the problem, schema filtering should not affect cloud-mode destinations. I've tried a number of different filtering approaches, including: filtering the destinations array before mapping it to attempt delivery of analytics calls, only filtering before enrichment, only filtering after enrichment, and various hierarchy re-arrangements within `availableExtensions` .

I'm wondering if the solution here should look similar to how schema filter is currently implemented, but with a different format for `All: false`. Perhaps we want to enable and disable certain integrations based on whether or not they are device mode destinations when unplanned events are blocked? Either way, it sounds like there's a desire to move away from relyign on `All: false` at all.

### DIFF
--- a/packages/browser/src/core/context/index.ts
+++ b/packages/browser/src/core/context/index.ts
@@ -49,6 +49,7 @@ export class Context implements AbstractContext {
   public stats: Stats
   private _id: string
   private _failedDelivery?: ContextFailedDelivery
+  private _disabledIntegrations: Set<string> = new Set()
 
   constructor(event: SegmentEvent, id?: string) {
     this._attempts = 0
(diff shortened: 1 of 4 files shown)


silesky:
@arielsilvestri  I think putting this into an SDD on Dropbox Paper would be helpful, since it sounds like we want to do some more ideation. IMO, this is exactly the kind of thing an SDD is good for, since it allows a forum for people to discuss API design topics.

arielsilvestri:
@silesky Good point, I'll move to that.

Issue #766, opened 2023-01-11
Export types from `@segment/analytics-node`?


mpalmerlee:
I'm working on integrating the public beta for the `@segment/analytics-node` package (`^0.0.1-beta.15`), it appears that the types for the core method parameters such as `IdentifyParams`, `TrackParams` and `AliasParams` are not exported. It would be nice to be able to use these types in my service so I can more easily type the parameters to my service functions that will call these methods.

silesky:
Fixed in @segment/analytics-node@0.0.1-beta.16!

mpalmerlee:
Thank you for the quick turn around @silesky 🙌 ❗  Confirmed this is working for me and I can now use these types with `0.0.1-beta.16` 🙏

PR #767, opened 2023-01-11
node: export Param types

silesky:
Addresses #766


packages/node/src/index.ts
```diff
@@ -1,6 +1,15 @@
...
+export type {
+  Plugin,
+  GroupTraits,
+  UserTraits,
+  TrackParams,
+  IdentifyParams,
+  AliasParams,
+  GroupParams,
+  PageParams,
+} from './app/types'
```
mpalmerlee, line 12: Nice! Thank you for the quick change 😍  🙌

PR #773, opened 2023-01-18
add web worker (partytown.io) example

silesky:
By popular demand 🤷 

It can send events to the tracking endpoint, and even some device mode destinations like mixpanel appear to work out of the box (fullstory looks broken). 

Anyone who wants to kick the tires is welcome!


chrisradek:
How much work would it be to create a new example for this scenario? It feels like the nextjs example is starting to get pretty complex with the addition of partytown (which we haven't fully vetted to see the tradeoffs.) I'm be worried a new user might think partytown is a requirement.

silesky:
> How much work would it be to create a new example for this scenario? It feels like the nextjs example is starting to get pretty complex with the addition of partytown (which we haven't fully vetted to see the tradeoffs.) I'm be worried a new user might think partytown is a requirement.

Not much work, I was just feeling lazy :-)

examples/with-next-js/public/~partytown/debug/partytown-atomics.js
```diff
@@ -0,0 +1,572 @@
+/* Partytown 0.7.4 - MIT builder.io */
```
chrisradek, line 1: Is the debug folder meant to be included?

examples/with-next-js/public/~partytown/partytown-atomics.js
```diff
@@ -0,0 +1,2 @@
+/* Partytown 0.7.4 - MIT builder.io */
```
chrisradek, line 1: Actually, should any of the `~partytown` directory be committed? It's built as part of the build step right? Otherwise we'll see updates anytime we update partytown?

examples/with-next-js/pages/web-worker/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+            'analytics.track',
+            'analytics.identify',
+            'analytics.group',
+            'analytics.reset',
+            'analytics.alias',
+          ]}
+        />
+        {/* copied the snippet from app.segment.com */}
+        <script
+          type="text/partytown"
```
chrisradek, line 24: I looked at the partytown docs and _think_ I understand what's going on here...

So because of the script type, the JS is actually ran in a web worker right? Though party town will interact with the main thread's DOM and APIs as needed?

And then the `Partytown` component is just forwarding main thread `window.analytics` calls to the web worker?

If that's correct (or wrong!) might be helpful to add a comment or 2 on how this works.

chrisradek (commented):
Just had some comments on what needs to be committed, and a sanity check on my own understanding of how this works!

examples/with-next-js/public/~partytown/debug/partytown-atomics.js
silesky: Yes

examples/with-next-js/public/~partytown/partytown-atomics.js
silesky: The docs are not super clear on this (searching sourcegraph, I see this both ways). I think you're correct, though, we don't need to commit the ~partytown folder.

chrisradek (approved):

packages/browser/README.md
```diff
@@ -192,11 +192,21 @@ export default defineComponent({
 </script>
 ```
 
+## Support for Web Workers (Experimental)
+ While this package does not support web workers out of the box, there are options:
 
+1. Run analytics.js in a web worker via [partytown.io](https://partytown.builder.io/). See [our partytown example](../../examples/with-next-js/pages/partytown). **Supports both cloud and device mode destinations, but not all device mode destinations may work.**
 
-### For snippet users only: How to add typescript support
+2. Try the [Segment Public API SDK](https://github.com/segmentio/public-api-sdk-typescript). **Warning: cloud destinations only!**
```
pooyaj, line 200: What is this for? Given the SDK is not relevant to data collection, does it need to be listed here?

pooyaj (approved):

PR #776, opened 2023-01-27
fix: support typescript "NodeNext" module resolution

thomasskk:
Currently typescript cant resolve types when `{ moduleResolution: "NodeNext" }` is in tsconfig.

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -7,7 +7,8 @@
   "types": "./dist/types/index.d.ts",
   "exports": {
     "require": "./dist/cjs/index.js",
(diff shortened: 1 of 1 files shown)


silesky:
@thomasskk -- 
1. I wasn't aware that putting "types" in the export map was something that was common. Is there a bug that you can link to in the typescript repo?

2. How are you using this package where you need NodeNext module resolution? 

3. Would this be more appropriate for https://github.com/segmentio/analytics-next/tree/master/packages/node?

silesky (approved):

Issue #768, opened 2023-01-12
Add keepalive flag for standard transport (not only batching)


arm1n:
Hi there,

I'm facing an issue when using the standard transport (fetch-dispatcher), where there is no `keepalive: true` setting applied on the fetch request, whereas it's put on the request when using the batch dispatcher. Is there any reason why this mechanism is not attached in the standard transport?

In my case I've to call `identify()` + `group()` right after users has entered a group, but there's a hard redirect happening right after that process, thus leading to cancelling these requests. If they would be configured with `keepalive` the problem could be circumvented.

Thanks for any feedback!

chrisradek:
@arm1n Thanks for opening this issue! It sounds reasonable to me to support setting `keepalive: true` for all requests - we'll try to dig into why that wasn't done to begin with when batched dispatcher was introduced.

We did make a recent update (that's rolling out to the CDN soon) that will store events in localstorage if the`fetch` call hasn't completed and you leave the page - then get sent from the new page so long as the domain is the same. But the keepalive option should be even more reliable for browsers that support it.

arm1n:
Thanks for the heads up @chrisradek - yes I agree, the `keepalive` option should be the safer approach and it shouldn't have a negative impact when just setting it in the standard transport as well 👍

arm1n:
Awesome, thanks for the update and implementation!

PR #787, opened 2023-02-10
add keep alive option to fetch standard transport

silesky:
https://github.com/segmentio/analytics-next/issues/768


chrisradek (approved):
:boom:

PR #788, opened 2023-02-10
make keepalive configurable -- defaults to false

silesky:
add a keepalive configuration option to deliveryStrategy.


chrisradek (approved):
Looks good to me!

PR #790, opened 2023-02-11
improve !hasUser error message

znwhite:
I was sending the `userId` as a number (because that is how they are stored in our database) and was receiving an error that the `userId` was missing. This led me on a wild goose chase that could have been easily solved if the error message had also included the information that it must be a string.

If the `hasUser` error check is going to fail if `userId` is not a string, then it should be included in the error message when it fails.


silesky:
That's annoying. I would like to know if user ID actually has to be a string, or if we can change the type to string | number. acc @chrisradek

znwhite:
@silesky Wen merge?

PR #800, opened 2023-02-24
 Fix staging-only bug where integrations URL would not respect CDN URL overrides

silesky:
Fixes a bug where staging CDNUrl is not properly overridden.

repro: https://sethsilesky.com/ajs-cd.html

image


pooyaj (approved):
💯

Issue #834, opened 2023-04-10
Unclear error message when `userId` is numeric - (Error integrating with Auth.js Login Events)


0xCLARITY:
I was getting the following error: `Must have userId or anonymousId or previousId or groupId`

With a setup like this:
```ts
// [...nextauth].ts

  events: {
    createUser: async (message) => {
      analytics.identify({
        userId: message.user.id,
        traits: {
          email: message.user.email,
        },
      });
    },
    signIn: async (message) => {
      analytics.track({
        userId: message.user.id,
        event: 'Developer Signed In',
      });
    },
```
    
See AuthJS docs on events for audit purposes here: https://next-auth.js.org/configuration/events

Switching back to `analytics-node v6` fixed the issue.

silesky:
Is your ID a string?

0xCLARITY:
Ah. Yup - that would do it I guess.

silesky:
PR here: https://github.com/segmentio/analytics-next/pull/835

PR #835, opened 2023-04-11
Fix ID Validation Error Message / refactor

silesky:
This should not have any behavior change.

I'm keeping the string requirement to be consistent with mobile SDK: I think it's better to force users to coerce IDs to string on their end, rather than have them surprised when their IDs are coerced for them. However, it's important to improve the error message. 

Anyone upgrading from analytics-node who is now facing more strict ID validation will continue to get a validation error, but that error won't be misleading.

https://segment.atlassian.net/browse/LIBWEB-1377


silesky:
Fixes https://github.com/segmentio/analytics-next/issues/834

pooyaj (approved):
Looks great! love the bundle size savings 📉

PR #838, opened 2023-04-15
Refactor page enrichment plugin

silesky:
Page Enrichment plugin was messy and written in a very imperative style, and and was hard to grok with so much overriding. This is a major improvement in readability by using a functional style.

-  pageDefault() was called twice for every page() call, now it's called once
- canonical() is now type safe and uses `querySelector` rather than `getElementsByTagName`
- passing page properties with an empty string should now properly override fields (tests in other PR -- related to semantics of the pick function).

This work is related to the work around fixing stale / wrong UTM params / page properties, but I want to keep my branches clean if possible.

### DIFF
--- a/.changeset/quiet-islands-kneel.md
+++ b/.changeset/quiet-islands-kneel.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': patch
+---
+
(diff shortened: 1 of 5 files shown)


danieljackins (approved):
Awesome 🎉

PR #839, opened 2023-04-16
Fix wrong utm parsing by making context.page.search the source of truth for query params.

silesky:
Fixes a utm-parameter parsing bug where overridden page.search properties would not be reflected in the context.campaign object
```ts
analytics.page(undefined, undefined, {search: "?utm_source=123&utm_content=content" )
analytics.track("foo", {url: "....", search: "?utm_source=123&utm_content=content" )

// should result in a context.campaign of:
{ source: 123, content: 'content'}
```
This fix is part 2 --
See: https://github.com/segmentio/analytics-next/pull/838

### DIFF
--- a/packages/browser/src/plugins/segmentio/normalize.ts
+++ b/packages/browser/src/plugins/segmentio/normalize.ts
@@ -126,13 +126,11 @@ export function normalize(
 ): object {
   const user = analytics.user()
 
+  // context should always exist here (see page enrichment)? ... and why would we default to json.options? todo: delete this
   json.context = json.context ?? json.options ?? {}
   const ctx = json.context
 
-  // This guard should not be neccessary -- why would context not exist here? Ditto ^ --
-  // page enrichment should add a context to an event by default.
-  // In any case, adding an empty string 'default'.
-  // We do not use the current search parameters, as they might be stale by this point.
+  // This guard against missing ctx.page should not be neccessary, since context.page is always defined
   const query: string = ctx.page?.search || ''
 
   delete json.options
--- a/packages/browser/src/plugins/segmentio/normalize.ts
+++ b/packages/browser/src/plugins/segmentio/normalize.ts
@@ -125,10 +125,15 @@ export function normalize(
   integrations?: LegacySettings['integrations']
 ): object {
   const user = analytics.user()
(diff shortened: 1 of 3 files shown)


packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -125,11 +125,16 @@ export function normalize(
   integrations?: LegacySettings['integrations']
 ): object {
   const user = analytics.user()
-  const query = window.location.search
 
   json.context = json.context ?? json.options ?? {}
   const ctx = json.context
 
+  // This guard should not be neccessary -- why would context not exist here? Ditto ^ --
```
silesky, line 132: @pooyaj might have some insight here? 

Thoughts about a broader refactor: It would be simpler / make sense if the page enrichment plugin was was also responsible for this logic:
```
  if (query && !ctx.campaign) {
    ctx.campaign = utm(query)
  }
```

If the current bugfix is safe, then moving the logic altogether should be equally safe / rational.

pooyaj: Yeah, I think this is mostly legacy behavior from AJS classic where the UTM params were injected in the Segment integration rather than in AJS core by accessing `window.location.search` directly. I think this change is great. But moving the UTM logic to the page enrichment plugin has some consequences (i.e., UTM params will be sent as a part of the context to all destinations ). I'm not sure if that would be ok or not tho.

pooyaj (approved):

packages/browser/src/plugins/segmentio/normalize.ts
silesky: Thanks, that makes sense. I guess we've decided that, while we pass the query string, we don't want to pass ctx.campaign to all destinations.   

Do you have any insight into this line:
```
json.context = json.context ?? json.options ?? {}
```
The page enrichment plugin already sets context to empty object if it doesn't exist.

danieljackins (approved):
Changes look good to me, might be worth testing it out on one of the affected customer pages as an override

silesky:
> Changes look good to me, might be worth testing it out on one of the affected customer pages as an override

Yep, I've tested it.

PR #846, opened 2023-04-23
Prevent usage of node or browser APIs in @segment/analytics-core and @segment/analytics-node

silesky:
1. Remove isOffline() from core
2. eslint error if any native node modules are used.
3. eslint error any of the following are used in source code of core/node (since isomorphic):
         'document',
          'window',
          'self',
          'process',
          'global',
          'navigator',


oscb (approved):
This lgtm.

Correct me if I'm getting this wrong: `packages/core` is the isomorphic core whereas `browser/core` will retain the browser specific implementation, right?

silesky:
> This lgtm.
> 
> Correct me if I'm getting this wrong: `packages/core` is the isomorphic core whereas `browser/core` will retain the browser specific implementation, right?

Yep.

PR #852, opened 2023-04-28
Make page context resilient to quick navigation changes

silesky:
#### What is this????
Previously, we only captured page information at event flush time, and it was possible for page information like URL / params to be wrong by the time analytics was fully loaded.  Now, we capture page information at function call time and store it in the arguments buffer, as if it was any other explicitly passed function argument like traits or properties.

Basically, every time there's a buffered method call (track, identify, etc), we gather the current page information like document.title, location.href, and use it to construct a `BufferedPageContext` object, that gets added as the last argument of the buffered argument list. Then, when we flush events, that last argument acts a kind of hidden parameter that we pop off the arguments list when the 'real' Analytics method is called -- and we use that stored page information to augment the event with the url, search params, title, etc.

#### Other Change list summary 
* Move page enrichment step to event factory, rename page enrichment plugin 'EnvironmentEnrichment'. 
* Parse UTM params into context.campaign if consumers override the default search params with an object.
* No changes to auto-track, but fixed tests and refactored poorly configured, brittle tests with hidden dependencies 
* Fix warnings in `UniversalStorage`

In order for the snippets to get the same resilliency, they require this update
```js
// using es5 to maximize bw compat
   !function () {
      // add this function - using single key names to save bytes
     function p() {
        var  c = document.querySelector("link[rel='canonical']");
        return {
           __t: 'bpc',  // discriminant key for extra uniqueness guarantee
          c: (c && c.getAttribute('href')) || undefined,
          p: location.pathname,
          u: location.href,
          s: location.search,
          t:  document.title,
          r:  document.referrer,
        }
      }
   
         // full snippet example in `index-buffered-page-ctx.html`
          analytics.factory = function (e) {
            return function () {
            ....
              var t = Array.prototype.slice.call(arguments);
             // add this statement
              if (['track', 'screen', 'alias', 'group', 'page', 'identify'].indexOf(e) > -1) t.push(p());
   
```

## Manual Test
1. cd examples/standalone-playground && yarn dev
2. open `index-buffered-page-ctx.html` (and read the description)


packages/browser/src/core/events/index.ts
```diff
@@ -235,21 +292,18 @@ export class EventFactory {
...
 
-    const body = {
+    const newEvent: SegmentEvent = {
       timestamp: new Date(),
       ...rest,
       context,
       integrations: allIntegrations,
       ...overrides,
     }
+    // NOTE() This is a solution that was implemented a long time ago to prevent some hash collision issue.
```
silesky, line 302: Slight opportunistic refactor -- this is one of the most confusing files in the code base.

oscb (dismissed):
lgtm! Just added a few comments about types

packages/browser/src/core/analytics/index.ts
```diff
@@ -109,6 +111,13 @@ function _stub(this: never) {
   console.warn(deprecationWarning)
 }
 
+const popPageContext = (args: any[]): PageContext | undefined => {
```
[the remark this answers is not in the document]
silesky: `unknown[]` would be work, but its the same level of typechecking. I am choosing to not make this because this final argument is "hidden" from typescript, as the way that positional arguments work here (see how these Params types are actually made to understand) make it impossible to safely add an "any last argument" (which I actually tried). The actual type signature is insane, and can't even be represented with function overloading AFAIK.

packages/browser/src/lib/pick.ts
```diff
@@ -3,15 +3,12 @@
...
-  object: T,
-  keys: readonly K[]
-): Pick<T, K> =>
+export const pick = <T>(object: T, keys: string[]): Partial<T> =>
   Object.assign(
     {},
     ...keys.map((key) => {
       if (object && Object.prototype.hasOwnProperty.call(object, key)) {
-        return { [key]: object[key] }
+        return { [key]: (object as any)[key] }
```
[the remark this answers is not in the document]
silesky: Fair call oput. I think  rather than adding more runtime code, I'd just have T extends the `Record<string, unknown>` type

packages/browser/src/browser/index.ts
```diff
@@ -169,9 +167,7 @@ async function flushFinalBuffer(
 ): Promise<void> {
   // Call popSnippetWindowBuffer before each flush task since there may be
   // analytics calls during async function calls.
-  buffer.push(...popSnippetWindowBuffer())
```
silesky, line 172: Refactor: encapsulated this into the Buffer logic itself, so it occurs every time "buffer.getCalls()" is called

packages/browser/src/plugins/page-enrichment/__tests__/index.test.ts
```diff
@@ -212,53 +212,6 @@ describe('Page Enrichment', () => {
...
-  afterEach(() => {
-    jest.restoreAllMocks()
-  })
-
-  it('handles no canonical links', () => {
-    const defs = pageDefaults()
-    expect(defs.url).not.toBeNull()
-  })
-
-  it('handles canonical links', () => {
```
silesky, line 233: FYI: This PR does not remove any tests, it just moves a few around. For example, these are moved to: packages/browser/src/core/page/__tests__/index.test.ts

packages/browser/src/core/page/add-page-context.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  return cookieOptions
+}
+
+type Ad = { id: string; type: string }
+
+/**
+ *
+ * @param obj e.g. { foo: 'b', bar: 'd'}
```
silesky, line 35: This adds a "feature" / fixes a bug where now we will calculate context.campaign if a user sets context.page.search to an object (as people sometimes do).

packages/browser/src/browser/index.ts
```diff
@@ -169,9 +167,7 @@ async function flushFinalBuffer(
...
   await flushAddSourceMiddleware(analytics, buffer)
-  buffer.push(...popSnippetWindowBuffer())
   flushAnalyticsCallsInNewTask(analytics, buffer)
```
silesky, line 171: ensure that buffer pops here -- also pop on toArray()

packages/browser/src/plugins/env-enrichment/index.ts
```diff
@@ -198,29 +143,9 @@ class PageEnrichmentPlugin implements Plugin {
...
-    }
-
-    evtCtx.page = {
-      ...defaultPageContext,
-      ...pageContextFromEventProps,
-      ...evtCtx.page,
-    }
-
-    const query: string = evtCtx.page.search || ''
+    const search = evtCtx.page!.search || ''
```
silesky, line 146: Note: `page` should never be empty since the page is created as part of the event factory. We probably can/should update the SegmentEvent type to reflect this.

danieljackins (approved):
Looks great! Test page is awesome 👌

packages/browser/src/core/page/add-page-context.ts
```diff
@@ -0,0 +1,33 @@
...
+      ...event.properties,
+      ...(event.name ? { name: event.name } : {}),
+    }
+  }
+
+  evtCtx.page = {
+    ...pageCtx,
+    ...pageContextFromEventProps,
+    ...evtCtx.page,
+  }
```
oscb, line 32: In general I prefer to keep these kind of functions immutable so that there's no invisible sideeffects

packages/browser/src/plugins/env-enrichment/index.ts
oscb: If we have an integration test that can guarantee that we don't move this order I think this is fine. If we don't then it might be better to add the extra lines to check it and prevent a crash.

packages/browser/src/core/page/__tests__/index.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    },
+    writable: true,
+  })
+})
+
+describe(isBufferedPageContext, () => {
+  it('should return true if object is page context', () => {
+    expect(isBufferedPageContext(getDefaultBufferedPageContext())).toBe(true)
+  })
+  it('should returm false if object is not page context', () => {
```
oscb, line 22: nit: typo `return`

oscb (approved):
Left a few comments. Some architectural suggestins, but I don't think they are a blocker. good work

packages/browser/src/core/buffer/index.ts
```diff
@@ -107,34 +143,81 @@ type ReturnTypeUnwrap<Fn> = Fn extends (...args: any[]) => infer ReturnT
...
-    return this
+  clear(): void {
+    // clear calls in the global snippet buffered array.
+    this._pushSnippetWindowBuffer()
+    // clear calls in this instance
+    this.calls = {}
+  }
+
+  toArray(): PreInitMethodCall[] {
+    return ([] as PreInitMethodCall[]).concat(...Object.values(this.calls))
```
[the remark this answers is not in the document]
silesky: Flattening: `Object.values` will return PreInitMethodCall[][] 😉

packages/browser/src/core/page/get-page-context.ts
```diff
@@ -0,0 +1,141 @@
...
+const removeHash = (href: string) => {
+  const hashIdx = href.indexOf('#')
+  return hashIdx === -1 ? href : href.slice(0, hashIdx)
+}
+
+const formatCanonicalPath = (canonicalUrl: string) => {
+  const a = document.createElement('a')
+  a.href = canonicalUrl
+  // remove leading slash from canonical URL path
+  return a.pathname[0] === '/' ? a.pathname : '/' + a.pathname
```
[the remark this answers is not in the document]
silesky: This is super legacy code -- I had the same question and did some testing, and a main difference is that new URL will throw errors (e.g. foo.com will throw an error because it doesn't have https))

packages/browser/src/core/page/add-page-context.ts
silesky: I more or less copied and pasted this from the plugin, which is why it mutates. You mean have `addPageContext` return a new event, rather than mutate?

packages/browser/src/core/buffer/index.ts
```diff
@@ -80,21 +86,51 @@ export const flushAnalyticsCallsInNewTask = (
...
+
+    /**
+     * For specific events, we want to add page context here
+     */
+    const shouldAddPageContext = (
+      ['track', 'screen', 'alias', 'group', 'page', 'identify'] as MethodName[]
+    ).includes(method)
+    this.args =
+      shouldAddPageContext && !hasBufferedPageContextAsLastArg(args)
+        ? [...args, getDefaultBufferedPageContext()]
```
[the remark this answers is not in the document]
silesky: That's a good refactor suggestion! done.

packages/browser/src/plugins/env-enrichment/index.ts
silesky: dit: Seth response (github bug, I'm not sure why I can't comment)
> If we have an integration test that can guarantee that we don't move this order I think this is fine. If we don't then it might be better to add the extra lines to check it and prevent a crash.

That's a good call out! IMO, there would be a runtime error in all of our integration tests here if page was not an object, so I actually felt OK about not adding a specific test. I couldn't think of a scenario where you could make that mistake in a refactor, unless you were to eliminate all integration tests. Typically I do agree that having an explicit test is better than implicit, and I could go either way!

packages/browser/src/core/analytics/index.ts
```diff
@@ -360,7 +379,8 @@ export class Analytics
       page,
       properties,
       options,
-      this.integrations
+      this.integrations,
+      pageCtx
```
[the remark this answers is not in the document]
silesky: That's a good point, but I think we can do that if we ever add another argument? It seems pretty unlikely that we will, though not totally out of the question. I would imagine any other meta page information would likely be added to `pageCtx`.

packages/browser/src/core/page/get-page-context.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  for (const k of BUFFERED_PAGE_CONTEXT_KEYS) {
+    if (!(k in bufferedPageCtx)) return false
+  }
+  // if object has both the correct key names AND the right number of keys, assume it's a BufferedPageContext
+  if (
+    Object.keys(bufferedPageCtx).length !== BUFFERED_PAGE_CONTEXT_KEYS.length
+  ) {
+    return false
+  }
```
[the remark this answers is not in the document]
silesky: Good find!

packages/browser/src/plugins/env-enrichment/index.ts
```diff
@@ -195,32 +140,13 @@ class PageEnrichmentPlugin implements Plugin {
...
-    evtCtx.page = {
-      ...defaultPageContext,
-      ...pageContextFromEventProps,
-      ...evtCtx.page,
-    }
+    const search = evtCtx.page!.search || ''
 
-    const query: string = evtCtx.page.search || ''
+    const query =
+      typeof search === 'object' ? objectToQueryString(search) : search
```
chrisradek, line 149: In what scenario is `search` and object?

silesky: Users sometimes override search with dictionary that contains parsed search params (e.g they use NextJS where it's on the router object), generally when they are making a page call.

chrisradek: I was wondering how we'd support query strings that had multiple values for the same key.  (Wasn't sure what was making the object in the first place for it to be a concern) For example `?foo=a&foo=b`

Looks like next allows strings or string arrays for the value:
```
NextApiRequest.query: Partial<{
    [key: string]: string | string[];
}>
```
But `new URLSearchParam()` won't work with `{ 'foo': ['a', 'b'] }` - it'd have to look like `[['foo', 'a'], ['foo', 'b']]`. It does accept `Record<string, string>`, but treats `Record<string, string[]>` oddly.

chrisradek: We could probably convert the object straight to the array of arrays version easily enough.

silesky: I wasn't suggesting we convert anything to  URLSearchParams, I was saying that we it would be plausible  accept that as an argument and convert them to a string here, as I'm doing with object.

silesky: @chrisradek I updated the code to handle  `[key: string]: string | string[]`.

packages/browser/src/core/page/add-page-context.ts
chrisradek: I believe the answer is "it does not", but these mutations aren't modifying the properties object the customer passes in right?

packages/browser/src/core/buffer/index.ts
```diff
@@ -107,34 +133,96 @@ type ReturnTypeUnwrap<Fn> = Fn extends (...args: any[]) => infer ReturnT
 
 type MethodCallMap = Partial<Record<PreInitMethodName, PreInitMethodCall[]>>
 
+type SnippetWindowBufferedMethodCall<
+  MethodName extends PreInitMethodName = PreInitMethodName
+> = [MethodName, ...PreInitMethodParams<MethodName>]
+
+/**
+ * A list of the method calls before initialization for snippet users
+ * For example, [["track", "foo", {bar: 123}], ["page"], ["on", "ready", function(){..}]
```
chrisradek, line 142: Appreciate the example comments!

packages/browser/src/core/buffer/index.ts
```diff
@@ -107,34 +133,96 @@ type ReturnTypeUnwrap<Fn> = Fn extends (...args: any[]) => infer ReturnT
...
-    return ([] as PreInitMethodCall[]).concat(...Object.values(this._value))
+  constructor(...calls: PreInitMethodCall[]) {
+    this.push(...calls)
+  }
+
+  /**
+   * Pull any buffered method calls from the window object, and use them to hydrate the instance buffer.
+   */
+  private get calls() {
+    this._pushSnippetWindowBuffer()
```
chrisradek, line 160: :+1: to this improvement

chrisradek (approved):
Overall think this looks great! Are you going to make a similar update to the snippet repo once this is merged?

silesky:
> Overall think this looks great! Are you going to make a similar update to the snippet repo once this is merged?

yep!

packages/browser/src/core/page/add-page-context.ts
silesky: > These mutations aren't modifying the properties object the customer passes in right?
Correct

PR #861, opened 2023-05-16
Increases specificity of localStorage keys to differentiate events by writeKey

chrisradek:
This PR adds the `writeKey` used to configure analytics.js to the `PersistedPriorityQueue` localStorage keys.

This prevents 2 analytics.js instances configured with different writeKeys but installed on the same domain (e.g. `https://foo.segment.com/app1` with writeKey `foo`, `https://foo.segment.com/app2` with writeKey `bar`) from reading each others retried events from localStorage.

Analytics.js saves events to localStorage when a user navigates away from a page that has events that haven't been sent to destinations/Segment.

### Special Notes

- classic integrations can also persist events to localStorage, so this led me to piping `writeKey` to a lot more places than I hoped.
- I thought about handling the fallback scenario where an end-user was just upgraded to the latest version of this library and had events in localStorage using the old key format. (Even wrote a test for it initially...)
I backed those changes out because it's extra tech debt that should only be needed for about a week, and I was worried it would make this issue still possible if a page in the domain was using a version of analytics.js that persisted to the old key.

### Testing

I added playwright tests that will execute the conditions that trigger events to be sent to localStorage.

### DIFF
--- a/packages/browser-integration-tests/src/segment-retries.test.ts
+++ b/packages/browser-integration-tests/src/segment-retries.test.ts
@@ -7,7 +7,7 @@ import {
   getPersistedItems,
 } from './helpers/get-persisted-items'
 
-test.describe('Standalone tests', () => {
+test.describe('Segment.io Retries', () => {
   test.beforeEach(standaloneMock)
   test.beforeEach(async ({ context }) => {
     await context.route(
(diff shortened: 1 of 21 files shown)


packages/browser/src/lib/priority-queue/persisted.ts
```diff
@@ -86,16 +86,17 @@ function mutex(key: string, onUnlock: Function, attempt = 0): void {
 }
 
 export class PersistedPriorityQueue extends PriorityQueue<Context> {
-  constructor(maxAttempts: number, key: string) {
+  constructor(maxAttempts: number, key: string, writeKey: string) {
```
silesky, line 89: Nit: I feel like it would be cleaner if this class didn't know what a writekey was (since it's part of our  domain logic, and this is hypothetically a generic). Also, we wouldn't have to repeat writekey:key an anymore!

packages/browser-integration-tests/src/segment-retries.test.ts
```diff
@@ -111,5 +119,85 @@ test.describe('Standalone tests', () => {
...
+      expect(request.method()).toBe('POST')
+      expect(request.postDataJSON().messageId).toBe(messageId1)
+      expect(request.postDataJSON().writeKey).toBe('key1')
+
+      // Make sure localStorage only has data for the 2nd writeKey - which wasn't reloaded.
+      persistedItems = await page.evaluate(getPersistedItems)
+
+      expect(persistedItems).toHaveLength(1)
+      expect(persistedItems[0].writeKey).toBe('key2')
+      expect(persistedItems[0].messageIds).toStrictEqual([messageId2])
```
silesky, line 200: This test... 🤩

packages/browser-integration-tests/src/segment-retries.test.ts
```diff
@@ -1,30 +1,38 @@
 import { Request, test, expect } from '@playwright/test'
 import { SettingsBuilder } from './fixtures/settings'
 import { standaloneMock } from './helpers/standalone-mock'
+import { extractWriteKeyFromUrl } from './helpers/extract-writekey'
+import {
+  PersistedQueueResult,
+  getPersistedItems,
+} from './helpers/get-persisted-items'
 
 test.describe('Standalone tests', () => {
```
silesky, line 10: Was this meant to be double nested in here?

packages/browser/src/core/queue/event-queue.ts
```diff
@@ -6,8 +6,14 @@ import { CoreEventQueue } from '@segment/analytics-core'
...
-  constructor(priorityQueue?: PriorityQueue<Context>) {
-    super(priorityQueue ?? new PersistedPriorityQueue(4, 'event-queue'))
+  constructor(writeKey: string)
+  constructor(priorityQueue: PriorityQueue<Context>)
+  constructor(writeKeyOrQueue: string | PriorityQueue<Context>) {
+    super(
+      typeof writeKeyOrQueue === 'string'
+        ? new PersistedPriorityQueue(4, 'event-queue', writeKeyOrQueue)
+        : writeKeyOrQueue
+    )
```
silesky, line 16: If we changed the signature, we could make this child class responsible for the writeKey.

something like:
writeKey: string, priorityQueue?: PriorityQueue

packages/browser-integration-tests/src/segment-retries.test.ts
chrisradek: Ooph, I think it was meant to signify these tests are for the standalone (CDN) version of a.js, not the NPM version. Can probably remove this describe block though.

packages/browser/src/lib/priority-queue/persisted.ts
chrisradek: Originally required it to handle the fallback case of reading keys that didn't include the `writeKey`, but I can remove that from this now. It did give me the benefit of having TypeScript yell at me if any usage didn't include the writekey though, but that work is done now.

danieljackins (approved):
LGTM!

silesky (approved):

PR #862, opened 2023-05-18
Add ability to use action-destinations from NPM

zikaari:
This pull request is the other half of https://github.com/segmentio/action-destinations/pull/1276.

This patch just adds the necessary moving parts that will let users bring in action-destinations from NPM.

### DIFF
--- a/packages/browser/src/browser/index.ts
+++ b/packages/browser/src/browser/index.ts
@@ -10,7 +10,11 @@ import { MetricsOptions } from '../core/stats/remote-metrics'
 import { mergedOptions } from '../lib/merged-options'
 import { createDeferred } from '../lib/create-deferred'
 import { pageEnrichment } from '../plugins/page-enrichment'
-import { remoteLoader, RemotePlugin } from '../plugins/remote-loader'
+import {
+  PluginFactory,
+  remoteLoader,
+  RemotePlugin
+} from '../plugins/remote-loader'
 import type { RoutingRule } from '../plugins/routing-middleware'
 import { segmentio, SegmentioSettings } from '../plugins/segmentio'
 import { validation } from '../plugins/validation'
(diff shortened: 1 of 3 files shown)


packages/browser/src/browser/index.ts
```diff
@@ -278,7 +284,16 @@ async function loadAnalytics(
 
   attachInspector(analytics)
 
-  const plugins = settings.plugins ?? []
+  const plugins = settings.plugins?.filter(
```
silesky, line 287: Nit: Can we encapsulate this into the `registerPlugins` so:
- `loadAnalytics` has a habit of getting large (it does a lot of different things, it's kind of a god function), so I've been gradually trying to break it up.
- `registerPlugins` doesn't need a new argument, since it can be derived from `plugins`. It already has a huge signature that's weird (opts + options?? LOL)

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -159,47 +160,61 @@ function isPluginDisabled(
   return false
 }
 
+async function loadPluginFactory(
+  remotePlugin: RemotePlugin,
+  obfuscate?: boolean
+): Promise<void | PluginFactory> {
```
silesky, line 166: nit: I think  `Promise<PluginFactory | undefined> ` would be a bit more idiomatic =)

packages/browser/src/browser/index.ts
zikaari: Done

silesky:
@neek Looks good, but CI is failing =)

silesky (approved):

.vscode/settings.json
```diff
@@ -24,7 +24,7 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[typescript]": {
-    "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
silesky, line 27: Was this change accidental? prettier-plugin-eslint` uses eslint to format, so vscode-eslint should be the 'formatter'.

zikaari: I didn't mean to add this change, it might've slipped in accidentally. I'll revert it back

```suggestion
    "editor.defaultFormatter": "dbaeumer.vscode-eslint"
```

PR #864, opened 2023-05-20
Introduce Client Hints API

danieljackins:
This PR introduces the Client Hints API. We will always request the low entropy hints, but each high entropy hint will only be requested as set in the options.

For example,

`AnalyticsBrowser.load(
      ...,
      { highEntropyValues: ['bitness', 'architecture', 'model', 'platformVersion', 'fullVersionList'] }
    )`
 
 Would populate `context.userAgentData` with something like
 
 `{
    "architecture": "x86",
    "bitness": "64",
    "brands": [
        {
            "brand": "Google Chrome",
            "version": "113"
        },
        {
            "brand": "Chromium",
            "version": "113"
        },
        {
            "brand": "Not-A.Brand",
            "version": "24"
        }
    ],
    "fullVersionList": [
        {
            "brand": "Google Chrome",
            "version": "113.0.5672.92"
        },
        {
            "brand": "Chromium",
            "version": "113.0.5672.92"
        },
        {
            "brand": "Not-A.Brand",
            "version": "24.0.0.0"
        }
    ],
    "mobile": false,
    "model": "",
    "platform": "macOS",
    "platformVersion": "13.1.0"
}`

### DIFF
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -103,6 +103,9 @@ export interface InitOptions {
         aid?: RegExp
         uid?: RegExp
       }
+  /**
+   * Array of high entropy Client Hints to request. These may be rejected by the user agent - only required hints should be requested.
+   */
   highEntropyValuesClientHints?: HighEntropyHint[]
 }
 
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -48,6 +48,7 @@ import { version } from '../../generated/version'
 import { PriorityQueue } from '../../lib/priority-queue'
 import { getGlobal } from '../../lib/get-global'
 import { AnalyticsClassic, AnalyticsCore } from './interfaces'
+import { HighEntropyHint } from '../../lib/client-hints/interfaces'
 
 const deprecationWarning =
   'This is being deprecated and will be not be available in future releases of Analytics JS'
@@ -102,7 +103,7 @@ export interface InitOptions {
         aid?: RegExp
         uid?: RegExp
       }
-  highEntropyValues?: string[]
+  highEntropyValuesClientHints?: HighEntropyHint[]
 }
 
 /* analytics-classic stubs */
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -48,6 +48,7 @@ import { version } from '../../generated/version'
 import { PriorityQueue } from '../../lib/priority-queue'
 import { getGlobal } from '../../lib/get-global'
 import { AnalyticsClassic, AnalyticsCore } from './interfaces'
+import { HighEntropyValues } from '../../plugins/segmentio/normalize'
 
 const deprecationWarning =
   'This is being deprecated and will be not be available in future releases of Analytics JS'
@@ -102,6 +103,7 @@ export interface InitOptions {
         aid?: RegExp
         uid?: RegExp
       }
+  highEntropyValues?: HighEntropyValues[]
 }
 
 /* analytics-classic stubs */
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -48,7 +48,6 @@ import { version } from '../../generated/version'
 import { PriorityQueue } from '../../lib/priority-queue'
 import { getGlobal } from '../../lib/get-global'
 import { AnalyticsClassic, AnalyticsCore } from './interfaces'
-import { HighEntropyValues } from '../../plugins/segmentio/normalize'
 
 const deprecationWarning =
   'This is being deprecated and will be not be available in future releases of Analytics JS'
@@ -103,7 +102,7 @@ export interface InitOptions {
         aid?: RegExp
         uid?: RegExp
       }
-  highEntropyValues?: HighEntropyValues[]
+  highEntropyValues?: string[]
 }
 
 /* analytics-classic stubs */
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -57,11 +57,15 @@ const deprecationWarning =
 const global: any = getGlobal()
 const _analytics = global?.analytics
 
-function createDefaultQueue(retryQueue = false, disablePersistance = false) {
+function createDefaultQueue(
+  name: string,
+  retryQueue = false,
+  disablePersistance = false
+) {
   const maxAttempts = retryQueue ? 4 : 1
   const priorityQueue = disablePersistance
     ? new PriorityQueue(maxAttempts, [])
-    : new PersistedPriorityQueue(maxAttempts, 'event-queue')
+    : new PersistedPriorityQueue(maxAttempts, name)
   return new EventQueue(priorityQueue)
 }
 
@@ -145,7 +149,12 @@ export class Analytics
     this.settings = settings
     this.settings.timeout = this.settings.timeout ?? 300
     this.queue =
-      queue ?? createDefaultQueue(options?.retryQueue, disablePersistance)
+      queue ??
+      createDefaultQueue(
+        `${settings.writeKey}:event-queue`,
+        options?.retryQueue,
+        disablePersistance
+      )
 
     this._universalStorage = new UniversalStorage(
       disablePersistance ? ['memory'] : ['localStorage', 'cookie', 'memory'],
--- a/packages/browser/src/lib/client-hints/index.ts
+++ b/packages/browser/src/lib/client-hints/index.ts
@@ -1,25 +1,16 @@
+import { NavigatorUAData, UADataValues } from './interfaces'
+
 export async function clientHints(
   hints?: string[]
 ): Promise<UADataValues | undefined> {
-  const userAgentData = ((): NavigatorUAData | undefined =>
-    (navigator as any).userAgentData)()
+  const userAgentData = (navigator as any).userAgentData as
+    | NavigatorUAData
+    | undefined
 
-  if (!userAgentData) return
+  if (!userAgentData) return undefined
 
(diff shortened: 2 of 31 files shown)


packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -9,6 +9,42 @@ import { SegmentioSettings } from './index'
 import { version } from '../../generated/version'
 import { getAvailableStorageOptions, UniversalStorage } from '../../core/user'
 
+declare global {
```
silesky, line 12: Couple of nit picks
1. Can we encapsulate this into a separate file/folder, maybe under core/client-hints/...? 

2. I don't think we want to declare a global; it will leak into anyone who imports our lib, so it can conflict with their typings when later versions of typescript support this out of the box, or if they have custom typings. It would actually better to just do const getUserAgentDate = (): UserAgentData =>(navigator as any).userAgentData

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -9,6 +9,42 @@ import { SegmentioSettings } from './index'
...
+}
+
+export type HighEntropyValues =
+  | 'architecture'
+  | 'bitness'
+  | 'model'
+  | 'platformVersion'
+  | 'fullVersionList'
+
+export interface UserAgentData {
```
silesky, line 30: I think we should vendor these types

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -9,6 +9,42 @@ import { SegmentioSettings } from './index'
 import { version } from '../../generated/version'
 import { getAvailableStorageOptions, UniversalStorage } from '../../core/user'
 
+declare global {
+  interface Navigator {
+    userAgentData: UserAgentData
```
silesky, line 14: Maybe we just vendor these types (not npm install, just copy)? https://www.npmjs.com/package/user-agent-data-types/v/0.3.1?activeTab=code

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -135,8 +136,12 @@ export function normalize(
 
   delete json.options
   json.writeKey = settings?.apiKey
-  ctx.userAgent = window.navigator.userAgent
 
+  ctx.userAgent = window.navigator.userAgent
```
silesky, line 140: not: I would skip window and just use "navigator", like we do on like 146

packages/browser/src/lib/client-hints/interfaces.ts
```diff
@@ -0,0 +1,43 @@
+// WICG Spec: https://wicg.github.io/ua-client-hints
+
+declare interface Navigator extends NavigatorUA {}
+declare interface WorkerNavigator extends NavigatorUA {}
+
+// https://wicg.github.io/ua-client-hints/#navigatorua
+declare interface NavigatorUA {
```
silesky, line 7: nit: I am proposing that we don't do any "declare" stuff here, since it's acts a kind of side effect for any consumer (these interfaces are global). We should just export the interfaces into our code as regular types. Thoughts?

packages/browser/src/lib/client-hints/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    try {
+      // this also includes the low entropy values, so we can overwrite here
+      userAgentDataContext = await userAgentData.getHighEntropyValues(hints)
+    } catch (_) {
+      return userAgentDataContext
+    }
+  }
+
+  return userAgentDataContext
+}
```
silesky, line 25: Nit: imo a shorter and a bit more 'functional' way to write this would be: 

```ts

export async function clientHints(...) {
  if (!(navigator as any).userAgentData) return undefined;
  
  const userAgentDataContext: UADataValues = {
      brands: userAgentData.brands,
      mobile: userAgentData.mobile,
      platform: userAgentData.platform,
    }
    
  if (!hints) return userAgentDataContext
  return userAgentData.getHighEntropyValues(hints).catch(() => userAgentDataContext)
}
```

packages/browser/src/lib/client-hints/interfaces.ts
danieljackins: Sounds good! I didn't know declare had that effect.

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -135,8 +136,12 @@ export function normalize(
...
+  const userAgentData = await clientHints(analytics.options.highEntropyValues)
```
chrisradek, line 141: Can we get the `clientHints` once when the plugin loads, then attach them to each event? This is a value that shouldn't change during the lifetime of a page, and would save us from having to make this an async function/make an 'expensive' call for each event.

packages/browser/src/core/analytics/index.ts
```diff
@@ -102,6 +102,7 @@ export interface InitOptions {
         aid?: RegExp
         uid?: RegExp
       }
+  highEntropyValues?: string[]
```
chrisradek, line 105: `requestClientHintsHighEntropyValues` or something more descriptive that ties this to Client Hints?

silesky: Is it actually necessary atm tho? I don't think this option was mentioned is the SDD. Seems like if you don't pass any hints specifically, it just gets everything it has -- seems OK for 99%, no?


... anyway, If we think we want to add it not or down the road, my vote is for "getHighEntropyValuesParams", since it's just the arguments you directly pass to https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues


 (Also -- small things if we want to add it
1. A doc string in the interface would be useful.
2. Would want a unit test to confirm the public API behavior.)

packages/browser/src/lib/client-hints/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  const userAgentData = ((): NavigatorUAData | undefined =>
+    (navigator as any).userAgentData)()
+
+  if (!userAgentData) return
+
+  let userAgentDataContext: UADataValues = {
+    brands: userAgentData.brands,
+    mobile: userAgentData.mobile,
+    platform: userAgentData.platform,
+  }
```
silesky, line 13: Rather than constructing this object, I think an idiomatic thing is to just do const lowEntropyValues = userAgentData.toJSON() -- I updated my code sample below.

packages/browser/src/core/analytics/index.ts
silesky: It's just typescript typings -- we can always expand on the list if it does change.

packages/browser/src/lib/client-hints/interfaces.ts
```diff
@@ -0,0 +1,46 @@
+// https://wicg.github.io/ua-client-hints/#dictdef-navigatoruabrandversion
+export interface NavigatorUABrandVersion {
+  readonly brand: string
+  readonly version: string
+}
+
+// https://wicg.github.io/ua-client-hints/#dictdef-uadatavalues
+export interface UADataValues {
```
silesky, line 8: we could just move this type to analytics core so we are all DRY?

packages/browser/src/lib/client-hints/interfaces.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// https://wicg.github.io/ua-client-hints/#navigatoruadata
+export interface NavigatorUAData extends UALowEntropyJSON {
+  getHighEntropyValues(hints: string[]): Promise<UADataValues>
+  toJSON(): UALowEntropyJSON
+}
+
+/**
+ * Array of high entropy Client Hints to request. These may be rejected by the user agent - only required hints should be requested.
+ * @return {Object} Object containing low entropy hints and successfully resolved high entropy hints
```
silesky, line 37: These jsdoc comments seem obsolete :-)

packages/browser/src/lib/client-hints/interfaces.ts
```diff
@@ -0,0 +1,46 @@
...
+// https://wicg.github.io/ua-client-hints/#dictdef-ualowentropyjson
+export interface UALowEntropyJSON {
+  readonly brands: NavigatorUABrandVersion[]
+  readonly mobile: boolean
+  readonly platform: string
+}
+
+// https://wicg.github.io/ua-client-hints/#navigatoruadata
+export interface NavigatorUAData extends UALowEntropyJSON {
+  getHighEntropyValues(hints: string[]): Promise<UADataValues>
```
silesky, line 31: Should we use HighEntropyHint here?

packages/browser/src/lib/client-hints/interfaces.ts
danieljackins: Oh you mentioned wanting a doc string and I thought you meant for this interface 😅 which did you mean?

silesky: Oh, I see what you did. The docstring goes in the InitOptions interface.

silesky:
FYI we should document this new userAgentData object like we do our other reserved properties

packages/browser/src/plugins/segmentio/index.ts
```diff
@@ -84,6 +85,10 @@ export function segmentio(
       ? batch(apiHost, deliveryStrategy.config)
       : standard(deliveryStrategy?.config as StandardDispatcherConfig)
 
+  const userAgentData = await clientHints(
```
chrisradek, line 88: Thoughts on wrapping this in a try/catch?

You've got good error handling in `clientHints` already - my paranoia stems from the whole client hints API still being marked experimental, and safari/firefox haven't implemented it yet. My fear is they expose a partial implementation without `toJSON` (I think super unlikely scenario but a failure would mean the segment plugin fails to load which is pretty critical)

chrisradek (approved):

PR #867, opened 2023-05-30
Add support for filtering within arrays

zikaari:


silesky:
@zikaari I assume you'll add a changeset per our offline discussion 😄

packages/browser/package.json
```diff
@@ -51,7 +51,7 @@
     "@segment/analytics-core": "1.2.5",
     "@segment/analytics.js-video-plugins": "^0.2.1",
     "@segment/facade": "^3.4.9",
-    "@segment/tsub": "1.0.1",
```
chrisradek, line 54: I'm not sure why we were pegged to a specific version - any reason not to pin to `^2.0.0`?

zikaari: I don't know why it was like this, historically. I can make it `^2.0.0` now

silesky (approved):
🔥

PR #877, opened 2023-06-29
Pluggable custom http client wrapper

MichaelGHSeg:
Creating a pluggable interface to wrap fetch calls, which will allow for custom clients to be used, such as a fetch implementation that supports http proxies.


packages/node/src/__tests__/integration.test.ts
```diff
@@ -52,11 +52,15 @@ describe('Error handling', () => {
     expect(() => analytics.track({} as any)).toThrowError(/event/i)
   })
 
-  it('should emit on an error', async () => {
-    const analytics = createTestAnalytics({ maxRetries: 0 })
-    fetcher.mockReturnValue(
-      createError({ statusText: 'Service Unavailable', status: 503 })
-    )
+  it.only('should emit on an error', async () => {
```
silesky, line 55: Oops, .only got left in :-)

packages/node/src/app/analytics-node.ts
```diff
@@ -51,6 +52,7 @@ export class Analytics extends NodeEmitter implements CoreAnalytics {
         httpRequestTimeout: settings.httpRequestTimeout,
         disable: settings.disable,
         flushInterval,
+        customclient: settings.customClient ?? new DefaultFetchClient(),
```
silesky, line 55: Thoughts on calling this setting "httpClient" instead of customClient? 

- pretty idiomatic
- Communicates that this is specifically an http client  
- Let's us add an option for more general non-http clients if we ever need to be (which would override hyypClient)

packages/node/src/lib/customhttpclient.ts
```diff
@@ -0,0 +1,20 @@
+export interface CustomHTTPClient {
+  send: (resource: any, options: any) => Promise<Response>
+}
+
+export class DefaultFetchClient implements CustomHTTPClient {
```
silesky, line 5: Sounds a little odd to me... I think this should be called `DefaultHttpClient`

packages/node/src/__tests__/disable.integration.test.ts
```diff
@@ -1,9 +1,28 @@
-const fetcher = jest.fn()
-jest.mock('../lib/fetch', () => ({ fetch: fetcher }))
-
 import { createTestAnalytics } from './test-helpers/create-test-analytics'
+import { CustomHTTPClient } from '../lib/customhttpclient'
+
+export class CheckFetchClient implements CustomHTTPClient {
```
silesky, line 4: While this works -- Instead of implementing our own custom spy logic to keep track of calls, it would be less boilerplate / more idiomatic to leverage  the existing jest library that has support for mocks/spies and all sorts of goodies. https://jestjs.io/docs/jest-object#jestspyonobject-methodname-accesstype

Ditto elsewhere. You can leverage this pattern to change from success and error between tests... let me know if you want help on that. Also, with that jest automatically clears all mock state between tests, so no need to do any reset.




e.g. 
```ts
import { createTestAnalytics } from './test-helpers/create-test-analytics'
import { createSuccess } from './test-helpers/factories'

describe('disable', () => {
  const mockSend: jest.Mock = jest.fn().mockResolvedValue(createSuccess())

  const checkFetchClient = {
     send: mockSend,
  }

  it('should dispatch callbacks and emit an http request, even if disabled', async () => {
    const analytics = createTestAnalytics({
      disable: true,
    })
    const emitterCb = jest.fn()
    analytics.on('http_request', emitterCb)
    await new Promise((resolve) =>
      analytics.track({ anonymousId: 'foo', event: 'bar' }, resolve)
    )
    expect(emitterCb).toBeCalledTimes(1)
  })

  it('should call fetch if disabled is false', async () => {
    const analytics = createTestAnalytics({
      disable: false,
      customClient: checkFetchClient,
    })
    await new Promise((resolve) =>
      analytics.track({ anonymousId: 'foo', event: 'bar' }, resolve)
    )
    expect(mockSend).toBeCalledTimes(1)
  })
  it('should not call fetch if disabled is true', async () => {
    const analytics = createTestAnalytics({
      disable: true,
      customClient: checkFetchClient,
    })
    await new Promise((resolve) =>
      analytics.track({ anonymousId: 'foo', event: 'bar' }, resolve)
    )
    expect(mockSend).toBeCalledTimes(0)
  })
})

  
```

packages/node/src/__tests__/integration.test.ts
MichaelGHSeg: Oops, thanks

packages/node/src/app/analytics-node.ts
MichaelGHSeg: Yeah, I think it's better, changing

packages/node/src/lib/customhttpclient.ts
MichaelGHSeg: Fixed

silesky:
Please see

PR #880, opened 2023-07-05
Add ability to pass custom http client into node

silesky:
Package: @segment/analytics-node.
image

Please see the types/default implementation here: `packages/node/src/lib/http-client.ts`

This is based on work / research from @MichaelGHSeg!

## Use a custom fetch-like implementation with proxy (simple, recommended)
```ts
import { HTTPFetchFn } from '../lib/http-client'

const httpClient: HTTPFetchFn = async (url, options) => {
  return axios({
    url,
    proxy: {
        protocol: 'http',
        host: 'proxy.example.com',
        port: 8886,
        auth: {
          username: 'user',
          password: 'pass',
        },
      },
    ...options,
  })
}

new Analytics({
  writeKey: '<YOUR_WRITE_KEY',
  httpClient,
})

```

## Augment the default HTTP Client
```ts
import { FetchHTTPClient, HTTPClientRequest  } from '@segment/analytics-node' 
 
class MyClient extends FetchHTTPClient {
  async makeRequest(options: HTTPClientRequest) {
    return super.makeRequest({
         ...options, 
         headers: { ...options.headers, foo: 'bar'  }
      }})
  }
}

const analytics = new Analytics({ 
  writeKey: '<YOUR_WRITE_KEY>', 
  httpClient: new MyClient() 
})
```

## Option 2: Completely override the full HTTPClient (Advanced, you probably don't need to do this)

```ts
import { HTTPClient, HTTPClientRequest } from '@segment/analytics-node'

class CustomClient implements HTTPClient {
  async makeRequest(options: HTTPClientRequest) {
    return someRequestLibrary(options.url, { 
      method: options.method,
      body: JSON.stringify(options.data) // serialize data
      headers: options.headers,
    })
  }
}
const analytics = new Analytics({ 
  writeKey: '<YOUR_WRITE_KEY>', 
  httpClient: new CustomClient() 
})
```

### DIFF
--- a/packages/node/src/app/emitter.ts
+++ b/packages/node/src/app/emitter.ts
@@ -14,7 +14,7 @@ export type NodeEmitterEvents = CoreEmitterContract<Context> & {
       url: string
       method: string
       headers: Record<string, string>
-      body: any
+      body: Record<string, unknown>
     }
   ]
   drained: []
--- a/packages/node/src/app/emitter.ts
+++ b/packages/node/src/app/emitter.ts
@@ -14,7 +14,7 @@ export type NodeEmitterEvents = CoreEmitterContract<Context> & {
       url: string
       method: string
       headers: Record<string, string>
-      body: Record<string, unknown>
+      body: Record<string, any>
     }
   ]
   drained: []
--- a/packages/node/src/app/emitter.ts
+++ b/packages/node/src/app/emitter.ts
@@ -14,7 +14,7 @@ export type NodeEmitterEvents = CoreEmitterContract<Context> & {
       url: string
       method: string
       headers: Record<string, string>
-      body: string
+      body: any
     }
   ]
   drained: []
--- a/packages/node/src/lib/http-client.ts
+++ b/packages/node/src/lib/http-client.ts
@@ -1,7 +1,9 @@
 import { abortSignalAfterTimeout } from './abort'
+import { fetch as defaultFetch } from './fetch'
 
 /**
- * This interface is meant to be compatible with different fetch implementations.
+ * This interface is meant to be compatible with different fetch implementations (node and browser).
+ * Using the ambient fetch type is not possible because the AbortSignal type is not compatible with node-fetch.
  * @link https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
  */
 export interface HTTPFetchFn {
@@ -60,14 +62,20 @@ export interface HTTPRequestOptions {
   timeout: number
 }
 
+/**
+ * HTTP client interface for making requests.cccccbenidtinulcgklueududltvedrjgeligdefjkfb
+ */
 export interface HTTPClient {
   makeRequest(_options: HTTPRequestOptions): Promise<HTTPFetchResponse>
 }
 
+/**
+ * Default HTTP client implementation using fetch.
+ */
 export class FetchHTTPClient implements HTTPClient {
   private _fetch: HTTPFetchFn
-  constructor(fetchFn: HTTPFetchFn) {
-    this._fetch = fetchFn
+  constructor(fetchFn?: HTTPFetchFn) {
+    this._fetch = fetchFn ?? defaultFetch
   }
   async makeRequest(options: HTTPRequestOptions): Promise<HTTPFetchResponse> {
     const [signal, timeoutId] = abortSignalAfterTimeout(options.timeout)
--- a/packages/node/src/lib/http-client.ts
+++ b/packages/node/src/lib/http-client.ts
@@ -35,7 +35,7 @@ export interface HTTPResponse {
  * This interface is meant to be a generic interface for making HTTP requests.
  * While it may overlap with fetch's Request interface, it is not coupled to it.
  */
-export interface HTTPClientRequestOptions {
+export interface HTTPClientRequest {
   /**
    * URL to be used for the request
    * @example 'https://api.segment.io/v1/batch'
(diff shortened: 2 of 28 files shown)


MichaelGHSeg (approved):

danieljackins (approved):

packages/node/src/plugins/segmentio/__tests__/test-helpers/validate-http.ts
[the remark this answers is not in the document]
silesky: Good catch!

packages/node/src/plugins/segmentio/publisher.ts
```diff
@@ -183,44 +186,38 @@ export class Publisher {
...
     while (currentAttempt < maxAttempts) {
       currentAttempt++
 
       let failureReason: unknown
-      const [signal, timeoutId] = abortSignalAfterTimeout(
-        this._httpRequestTimeout
-      )
       try {
-        const requestInit = {
-          signal: signal,
```
MichaelGHSeg, line 199: Did you have a specific fetch replacement library in mind that doesn't support 'signal' here?  I have questions about simultaneously allowing access to tinker with our internals and passing around non-standard objects.

silesky: All fetch libraries support AbortSignal, but not all HTTP Libraries do (e.g Request).
 
That's why signal  is only present in the HTTPFetch interface.

MichaelGHSeg: I just wanted to know what the impetus was.  Figured you had a reason for moving some of the fiddly bits into the user replaceable thing that they then have to repeat.

packages/node/src/lib/http-client.ts
```diff
@@ -0,0 +1,95 @@
...
+  method: string
+  /**
+   * Headers to be sent with the request
+   */
+  headers: Record<string, string>
+  /**
+   * JSON data to be sent with the request (will be stringified)
+   * @example { "batch": [ ... ]}
+   */
+  data: Record<string, any>
```
chrisradek, line 57: It feels a little weird for an HTTP client to have `data` instead of `body`. (headers and query params are data too!)

How come we accept an object rather than a string/byte array here? I would expect the HTTP Client to only be concerned with making the HTTP request, not transforming the body.

packages/node/src/lib/http-client.ts
```diff
@@ -0,0 +1,95 @@
...
+  method?: string
+  signal?: any // AbortSignal type does not play nicely with node-fetch
+}
+
+/**
+ * This interface is meant to very minimally conform to the Response interface.
+ * @link https://developer.mozilla.org/en-US/docs/Web/API/Response
+ */
+export interface HTTPResponse {
+  ok: boolean
```
chrisradek, line 29: Kind of wonder if we really need `ok` - it's just a boolean that is `true` if `status` if 200-299. We get `status` already so could check that ourselves (then if someone uses an HTTP client that doesn't return `ok`, they don't have to worry about it)

packages/node/src/lib/http-client.ts
```diff
@@ -0,0 +1,95 @@
...
+    const requestInit = {
+      url: options.url,
+      method: options.method,
+      headers: options.headers,
+      body: JSON.stringify(options.data),
+      signal: signal,
+    }
+
+    return this._fetch(options.url, requestInit).finally(() =>
+      clearTimeout(timeoutId)
```
chrisradek, line 92: Ah I see, timeout is time to response headers (to answer my earlier question :smile:)

packages/node/src/lib/http-client.ts
oscb: I can see the possibility of wanting to change the data before sending to the server and hence wanting to have the object and not parse from JSON.

But we have already a few places to do these operations, so I agree with this, let's keep the HTTP Client dumb to discourage having too much logic here

oscb (changes requested):
Looks good. I +1 on some of the comments before, I think the most important one would be to remove the `ok` from the responses to prevent weird issues where it's missed to turn on (or the other way around.

packages/node/src/lib/http-client.ts
silesky: `data` is used by `axios` to describe JSON data, rather than body, which is a string

packages/node/src/app/emitter.ts
```diff
@@ -14,7 +14,7 @@ export type NodeEmitterEvents = CoreEmitterContract<Context> & {
       url: string
       method: string
       headers: Record<string, string>
-      body: string
+      body: Record<string, any>
```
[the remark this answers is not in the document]
silesky: The serialization part is  the client's responsibility, and I didn't want the http client to have take an emitter -- so the emitting part now happens in the publisher code (before the request is passed to the http client). I saw no downsides to this change, and only upsides.

packages/node/src/lib/http-client.ts
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * JSON data to be sent with the request (will be stringified)
+   * @example { "batch": [ ... ]}
+   */
+  data: Record<string, any>
+  /**
+   * Timeout in milliseconds
+   * @example 10000
+   */
+  timeout: number
```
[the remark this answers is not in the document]
silesky: Good point. We currently support `httpRequestTimeout` which is "timeout for full response body" -- I can rename it to match our config.

oscb (approved):
👍

PR #883, opened 2023-07-10
Add eslint-plugin-jest and jest rules; fix tests that violate eslint rules

silesky:
This came out of a convo with @MichaelGHSeg  where we saw that it was possible to push a .only ('focused') test without CI failing.


packages/browser/src/browser/__tests__/analytics-pre-init.integration.test.ts
```diff
@@ -141,14 +141,14 @@ describe('Pre-initialization', () => {
 
   describe('Promise API', () => {
     describe('.then', () => {
-      test('.then should be called on success', (done) => {
```
oscb, line 144: I didn't even knew this was a way to async testing, good riddance!

oscb (approved):

PR #888, opened 2023-07-13
Fix query string bug

danieljackins:
Since we started looking at the ctx object for the query string, if a user passed in an object for the 'search' key an error would be thrown from trying to parse a string. This was causing these events to be dropped.

### DIFF
--- a/packages/browser/src/plugins/segmentio/normalize.ts
+++ b/packages/browser/src/plugins/segmentio/normalize.ts
@@ -131,7 +131,7 @@ export function normalize(
   const ctx = json.context
 
   // This guard against missing ctx.page should not be neccessary, since context.page is always defined
-  const query: string = ctx.page?.search || ''
+  const query: string | any = ctx.page?.search || ''
 
   delete json.options
   json.writeKey = settings?.apiKey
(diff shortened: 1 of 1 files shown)


silesky:
@danieljackins is there another PR that's following this?

danieljackins:
> @danieljackins this should always be a string https://segment.com/docs/connections/spec/page/#properties? I am not sure this is _exactly_ what we want

Right, but this bug report was from a customer who was passing in an object to it in analytics-classic (and older versions of analytics-next) successfully.

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -131,7 +131,8 @@ export function normalize(
   const ctx = json.context
 
   // This guard against missing ctx.page should not be neccessary, since context.page is always defined
-  const query: string = ctx.page?.search || ''
```
oscb, line 134: Not sure where we set `search` for ourselves, but if it's internal maybe we should try putting our stuff in a particular nested object `internal`? or use `_search` ?

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -131,7 +131,8 @@ export function normalize(
...
+  const query: string =
+    typeof ctx.page?.search === 'string' ? ctx.page?.search : ''
```
silesky, line 135: This line needs a unit tests, correct? Page calls should be able to take search as an object without throwing.

silesky (approved):

PR #891, opened 2023-07-13
[LIBWEB-1353] Fix cookie write error

zikaari:
Analytics would fail fatally if cookies are made inaccessible by overriding the cookie getter/setter by a piece of code. This patch adds resilience against such a case.

### DIFF
--- a/.changeset/pink-dancers-exist.md
+++ b/.changeset/pink-dancers-exist.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': minor
+---
+
+[LIBWEB-1353] Fix cookie write error
--- a/.changeset/pink-dancers-exist.md
+++ b/.changeset/pink-dancers-exist.md
@@ -1,5 +1,5 @@
 ---
-'@segment/analytics-next': minor
+'@segment/analytics-next': patch
 ---
 
 [LIBWEB-1353] Fix cookie write error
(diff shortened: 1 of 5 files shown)


silesky (approved):
LGTM

packages/browser/src/core/user/index.ts
```diff
@@ -62,15 +62,14 @@ const ONE_YEAR = 365
...
 export class Cookie extends Store {
   static available(): boolean {
-    let cookieEnabled = window.navigator.cookieEnabled
-
-    if (!cookieEnabled) {
-      jar.set('ajs:cookies', 'test')
-      cookieEnabled = document.cookie.includes('ajs:cookies')
-      jar.remove('ajs:cookies')
+    try {
+      jar.set('ajs_cookies', 'test')
```
oscb, line 66: nit: for quick reuse putting `ajs_cookies` as a const `AJS_COOKIE_AVAILABILITY_KEY` or similar.

I would also suggest renaming it to `ajs_cookies_check` or something to highight that it is only an availability check cookie.

oscb (approved):

packages/browser/src/core/user/index.ts
silesky: good point! All of these strings get repeated in the bundle -- using a constant saves bytes!

zikaari: Updated

.changeset/pink-dancers-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'@segment/analytics-next': minor
```
chrisradek, line 2: Nit: Probably a patch since this is a bug fix/doesn't introduce any new APIs.

chrisradek (approved):
Looks good to me! Just minor comment on versioning.

PR #895, opened 2023-07-13
Treat C001 as a regular category in OneTrust consent wrapper.

silesky:
Remove strictly necessary cookie check.

Reasoning:
- If Segment is being used for critical workflows, or loads critical plugins, `C001` (Strictly neccessary) is the right category to use. 

[✔️ ] I've included a changeset


oscb (approved):
👍

PR #896, opened 2023-07-13
fix: support screen events on AJS on Segment Destination

oscb:
Adds `screen` support to Segment's destination.

Previously `screen` events where captured and processed correctly but not sent to Segment.

Added tests.

[✔︎] I've included a changeset (psst. run `yarn changeset`. Read about changesets [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)).


silesky (approved):
LGTM

PR #898, opened 2023-07-14
Only enable integration if _all_ consent categories are consented to.

silesky:
- Only enable integration if _all_ consent categories are consented to.
- This matches the behavior on ingestion.


chrisradek (approved):
:fire:

PR #899, opened 2023-07-14
Validate categories during consent stamping

silesky:
This is work based on previous discussion with @oscb and @chrisradek.
- Add validation error if no categories are found
- Validate categories for consent stamping
- Improve tests

### DIFF
--- a/packages/consent/consent-tools/src/domain/create-wrapper.ts
+++ b/packages/consent/consent-tools/src/domain/create-wrapper.ts
@@ -12,6 +12,7 @@ import { validateCategories, validateOptions } from './validation'
 import { createConsentStampingMiddleware } from './consent-stamping'
 import { pipe, pick, uniq } from '../utils'
 import { AbortLoadError, LoadContext } from './load-cancellation'
+import { ValidationError } from './validation/validation-error'
 
 export const createWrapper: CreateWrapper = (createWrapperOptions) => {
   validateOptions(createWrapperOptions)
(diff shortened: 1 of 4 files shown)


packages/consent/consent-tools/src/domain/create-wrapper.ts
```diff
@@ -90,10 +89,15 @@ export const createWrapper: CreateWrapper = (createWrapperOptions) => {
 
         if (!allCategories.length) {
           // No configured integrations found, so no categories will be sent (should not happen unless there's a configuration error)
-          return undefined
+          throw new ValidationError(
+            'Invariant: No consent categories defined in Segment',
+            []
+          )
```
oscb, line 95: Do we handle this Exception or is this bubbled up all the way to the user?

Just curious. I'm ok with this preventing any operation as consent is not defined at this point, but something to call out in the documentation so that people set their categories first before enabling consent in their clients.

silesky: > Do we handle this Exception or is this bubbled up all the way to the user?
> 
> Just curious. I'm ok with this preventing any operation as consent is not defined at this point, but something to call out in the documentation so that people set their categories first before enabling consent in their clients.


This error stops the pipeline -- it's async, so the error is exposed to the user via the usual mechanism (I think like any other plugin, it's logged in context.failedDelivery()). I don't think it bubbles up completely.

I think that's a reasonable thing to document (we are still in private beta, so don't have to worry about that). Categories need to be set regardless -- since we're validating categories on first load -- this feature does not change that behavior.

oscb (approved):

chrisradek (approved):

PR #901, opened 2023-07-15
Do not throw error when analytics are disabled.

Pruxis:
Imagine the following case where you disable segment in your code.

const { Analytics } = require('@segment/analytics-node');
new Analytics({ writeKey: processe.env.WRITE_KEY || '', disable: true });

If WRITE_KEY is not present, but your analytics are disabled this still throws an error.

This would simplify local setups and do not force them to have a truthy string as write key

### DIFF
--- a/packages/node/src/app/settings.ts
+++ b/packages/node/src/app/settings.ts
@@ -43,7 +43,7 @@ export interface AnalyticsSettings {
 }
 
 export const validateSettings = (settings: AnalyticsSettings) => {
-  if (!settings.writeKey) {
+  if (!settings.writeKey && !settings.disable) {
     throw new ValidationError('writeKey', 'writeKey is missing.')
   }
 }


chrisradek:
@Pruxis Thanks for opening this pull request! I think from a TypeScript perspective `writeKey` will still be required with this change - @silesky do you think we should update types as well (I think we'd need a tagged union to make this nice) or really just keep this a runtime check?

silesky:
> @Pruxis Thanks for opening this pull request! I think from a TypeScript perspective `writeKey` will still be required with this change - @silesky do you think we should update types as well (I think we'd need a tagged union to make this nice) or really just keep this a runtime check?

No Typescript changes are needed here

Pruxis:
@silesky Do I need to write some tests to get this merged?

PR #903, opened 2023-07-18
Add onetrust-wrapper unit tests

silesky:
Testing is for winners. 🥇


packages/consent/consent-wrapper-onetrust/src/lib/__tests__/onetrust-api.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+beforeEach(() => {
+  // @ts-ignore
+  delete window.OneTrustActiveGroups
+  // @ts-ignore
+  delete window.OneTrust
+})
+
+describe(getConsentedGroupIds, () => {
+  it('should return an empty array if no groups are consented', () => {
```
danieljackins, line 13: Is this test named incorrectly?

danieljackins (approved):
LGTM! Just one comment about a test name

packages/consent/consent-wrapper-onetrust/src/lib/__tests__/onetrust-api.test.ts
silesky: Hmm... it looks good to me eyes. What am I missing?

danieljackins: I'm probably missing something, but the test says should return an empty array(...), and then we're expecting a non-empty array?

Issue #907, opened 2023-07-19
Support for Web Extension Service Workers


kyranjamie:
We've just upgraded our Chrome extensions to MV3 and are unable to collect analytics in our background script (service worker).

Are there any workarounds for this? I see Partytown mentioned but we're not using this tool and believe there must be a simpler solution. 

Currently, the library internally writes the anon id to `localStorage`. This presumes the environment. Ideally, the `AnalyticsBrowser` could accept some kind of storage API with a `getItem`, `setItem`, `removeItem` API? This way we could pass it the `chrome.storage.local` object.

Something like:

```ts
new AnalyticsBrowser({
  storage: chrome.storage.local
})
```

jonandgon:
Is there any movement on this? The lack of support for chrome extensions is make/break for us.

kyranjamie:
Chasing this. We have a huge gap in our extension's analytics currently owing to this issue.

markmhendrickson:
+1 this is sorely needed

PR #908, opened 2023-07-20
feat: storage options, change priority or use custom

oscb:
# Adds a `storage` option to analytics client. 

This can be used to specify a particular order for storage systems. e.g.

```js
storage: [StoreType.Cookie, StoreType.LocalStorage, StoreType.Memory]
```

or 

```js
storage: ['cookie', 'localStorage', 'memory']
```

This is useful for web apps that might use multiple domains/subdomains to use cookies over local storage.

- This PR also moves class implementations from the `User` package to `Storage` as `User` had become too bloated.
- The `UniversalStorage` option for overriding particular `StoreType` is removed as the behaviour was getting complex to maintain if the order is also switchable. It is recommended to create a `UniversalStorage` object with the particular systems instead.
- Refactored the `UniversalStorage` arguments to prevent passing around `cookieOptions` as an specific thing all over the place. New API should be more extensible if other storages require any additional settings.
- Refactored the `Store` class to reuse the same interface of `UniversalStorage` as they were pretty much the same.
- Adds additional tests.

WIP: Custom Storage support

[✔︎] I've included a changeset (psst. run `yarn changeset`. Read about changesets [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)).


silesky:
there's a related  festuren req here? https://github.com/segmentio/analytics-next/issues/907

packages/browser/src/core/storage/__tests__/prioritizedListStorage.test.ts
```diff
@@ -0,0 +1,159 @@
+import jar from 'js-cookie'
+import { CookieStorage } from '../cookieStorage'
+import { LocalStorage } from '../localStorage'
+import { MemoryStorage } from '../memoryStorage'
+import { UniversalStorage } from '../universalStorage'
+describe('prioritizedListStorage', function () {
```
danieljackins, line 6:
```suggestion
import { UniversalStorage } from '../universalStorage'

describe('prioritizedListStorage', function () {
```

packages/browser/src/core/storage/__tests__/prioritizedListStorage.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    })
+
+    it('get data from memory', function () {
+      jar.set('ajs_test_key', '🍪')
+      localStorage.setItem('ajs_test_key', '💾')
+      const us = new UniversalStorage([new MemoryStorage()])
+      expect(us.get('ajs_test_key')).toBeNull()
+    })
+
+    it('order of default targets matters!', function () {
```
danieljackins, line 48: Did you mean to change the order in this test? It's just repeating the "picks data from cookies first" test.

packages/browser/src/core/storage/__tests__/prioritizedListStorage.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    })
+
+    it('returns null if there are no storage targets', function () {
+      jar.set('ajs_test_key', '🍪')
+      localStorage.setItem('ajs_test_key', '💾')
+      const us = new UniversalStorage([])
+      expect(us.get('ajs_test_key')).toBeNull()
+    })
+
+    //   it('can override the default targets', function () {
```
danieljackins, line 62: Is this intended to be commented out?

packages/browser/src/core/storage/__tests__/prioritizedListStorage.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    //     expect(us.get('ajs_test_key', ['localStorage', 'memory'])).toEqual('💾')
+    //     expect(us.get('ajs_test_key', ['cookie', 'memory'])).toEqual('🍪')
+    //     expect(us.get('ajs_test_key', ['cookie', 'localStorage'])).toEqual('🍪')
+    //     expect(us.get('ajs_test_key', ['cookie'])).toEqual('🍪')
+    //     expect(us.get('ajs_test_key', ['memory'])).toEqual(null)
+    //   })
+  })
+
+  describe('#set', function () {
+    it('set the data in all storage types', function () {
```
danieljackins, line 78:
```suggestion
    it('sets the data in all storage types', function () {
```

packages/browser/src/core/storage/__tests__/prioritizedListStorage.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        new CookieStorage(),
+        new MemoryStorage(),
+      ])
+      us.set('ajs_test_key', '💰')
+      expect(jar.get('ajs_test_key')).toEqual('💰')
+      expect(localStorage.get('ajs_test_key')).toEqual(null)
+      expect(us.get('ajs_test_key')).toEqual('💰')
+    })
+
+    //   it('can override the default targets', function () {
```
danieljackins, line 138: Same question here

packages/browser/src/core/storage/__tests__/prioritizedListStorage.test.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    it('can save and retrieve from memory when there is no other storage', function () {
+      const us = new UniversalStorage([new MemoryStorage()])
+      us.set('ajs_test_key', '💰')
+      expect(jar.get('ajs_test_key')).toEqual(undefined)
+      expect(localStorage.getItem('ajs_test_key')).toEqual(null)
+      expect(us.get('ajs_test_key')).toEqual('💰')
+    })
+
+    it('does not write to cookies when cookies are not available', function () {
```
danieljackins, line 110: Can we use the disableCookies/disableLocalStorage helpers here?

packages/browser/src/core/storage/settings.ts
```diff
@@ -0,0 +1,28 @@
+import { Storage, StoreType, StoreTypeWithSettings } from './types'
+
+export type StorageSettings = Storage | StoreType[]
+
+export function isArrayOfStoreType(s: StorageSettings): s is StoreType[] {
+  return (
+    s !== undefined &&
+    s !== null &&
```
danieljackins, line 8:
```suggestion
    s &&
```

packages/browser/src/core/storage/settings.ts
```diff
@@ -0,0 +1,28 @@
...
+    Array.isArray(s) &&
+    s.every((e) => Object.values(StoreType).includes(e))
+  )
+}
+
+export function isStorageObject(s: StorageSettings): s is Storage {
+  return (
+    s !== undefined &&
+    s !== null &&
```
danieljackins, line 17:
```suggestion
    s &&
```

packages/browser/src/core/user/index.ts
```diff
@@ -1,8 +1,20 @@
...
+  StorageObject,
+  StorageSettings,
+  StoreType,
+  applyCookieOptions,
+  initializeStorages,
+  isArrayOfStoreType,
+  isStorageObject,
+} from '../storage'
+import { MemoryStorage } from '../storage/memoryStorage'
+import {} from '../storage/settings'
```
danieljackins, line 17: Is this intended?

packages/browser/src/core/storage/types.ts
```diff
@@ -0,0 +1,93 @@
...
+export abstract class BaseStorage<Data extends StorageObject = StorageObject>
+  implements Storage<Data>
+{
+  abstract get type(): StoreType | string
+  abstract get available(): boolean
+  abstract get<K extends keyof Data>(key: K): Data[K] | null
+  abstract set<K extends keyof Data>(key: K, value: Data[K] | null): void
+  abstract clear<K extends keyof Data>(key: K): void
+  /**
+   * By default a storage getAndSync will handle calls exactly as the
```
danieljackins, line 76: Unfinished comment here

danieljackins (approved):
Looks awesome! Just some questions/nits but nothing blocking

kyranjamie:
Great to see this being worked on. Note that `chrome.storage.local` has an asynchronous API, so ideally this will work with Promises too

```ts
getItem(): Promise<any> | any;
setItem(): Promise<any> | any;
removeItem(): Promise<any> | any;
```

packages/browser/src/core/analytics/__tests__/integration.test.ts
```diff
@@ -271,4 +273,30 @@ describe('Analytics', () => {
       expect(fn).toHaveBeenCalledTimes(1)
     })
   })
+
+  describe('storage', () => {
+    beforeEach(() => {
+      clearAjsBrowserStorage()
+    })
+
+    it('handles custom priority storage', async () => {
```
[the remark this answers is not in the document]
oscb: I might add one where we disable one. The rest of the tests are using the LocalStorage -> Cookie -> Memory. I think we can safely expect that the most important scenario to test is making `Cookie` take precedence.

packages/browser/src/core/analytics/index.ts
```diff
@@ -194,7 +205,47 @@ export class Analytics
...
+  private createStore(
+    disablePersistance: boolean,
+    storageSetting: InitOptions['storage'],
+    cookieOptions?: CookieOptions | undefined
+  ): Storage {
+    // DisablePersistance option overrides all, no storage will be used outside of memory even if specified
+    if (disablePersistance) {
+      return new MemoryStorage()
+    } else {
+      if (storageSetting !== undefined && storageSetting !== null) {
```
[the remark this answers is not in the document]
oscb: In general I prefect explicit checks to truthy/falsy, but since this one is a config and I'm doing typeguarding below it's ok!

packages/browser/src/core/storage/__tests__/prioritizedListStorage.test.ts
oscb: Safe to remove, this feature is now deleted

packages/browser/src/core/user/index.ts
oscb: Nope, no idea how this happened tbs!

silesky (changes requested):
Thanks for this! There's a lot to like about this PR, such as the refactoring. Blocking for some more discussion/review of this API, since there is no SDD. 

Can we base Storage on the idiomatic Web Storage API (and/or with AsyncStorage support per @kyranjamie comment), rather than using the same interface as UniversalStorage?  The UniversalStorage interface is both larger than I think it needs to be and confusing and I would hesitate to commit to this as our main public API for storage behavior.

Other concerns I have as a user are similarly around UniversalStorage and the ways it differs from what I'd expect from a typical storage interface.
- storage.getAndSync vs storage.get -- the former is an implementation detail. I'd be confused how to implement these and why if I'm just adding some basic storage implementatio, and what I'm expecting to be syncing with if there's only one store. Consumers can power up their storage implementation  whichever way they want (if they want to combine cookies and indexDB into one class, they can go ahead) -- we don't  need to complicate our interface.
- Why do I need  to implement 'type' if storage only accepts a single store instance? 
- can we make .available  not required?

oscb:
> 



> * storage.getAndSync vs storage.get -- the former is an implementation detail. I'd be confused how to implement these and why if I'm just adding some basic storage implementatio, and what I'm expecting to be syncing with if there's only one store. Consumers can power up their storage implementation  whichever way they want. if they want to combine cookies and indexDB into one class, they can go ahead. Maybe  we don't  need to complicate our interface.

Agree on this. `getAndSync` doesn't make much sense anymore if we're killing the weird support of overriding the Storages to use in each operation. I'm fine with leaving this up to the `Storage` implementation

> Why do I need to implement 'type' if storage only accepts a single store instance?
I think this was due to an old approach I took (?) doesn't seem like I need this anymore. Fine with removing it.

> can we make .available not required?
I actually think we should keep this one. That way you can do fallbacks of Storage systems abstracting a little bit more of how you know if they are ready.

I'm ok with switching from the `UniversalStorage` interface, just wanted to stick to it since there's at least one destination using it currently and didn't want to break stuff

packages/browser/src/core/storage/types.ts
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Defines a Storage object for use in AJS Client.
+ */
+export interface Storage<Data extends StorageObject = StorageObject> {
+  /**
+   * Returns the kind of storage.
+   * @example cookie, localStorage, custom
+   */
+  get type(): StoreType | string
```
silesky, line 24: Hmm, doesn't this just reduce to string? I guess it's for documentation...

silesky:
Just occurred to me: can we use a string union instead of enums? Using enums is bad for snippet users from a docs perspective and makes typing complicated since snippet users would have to use strings anyway (which would actually result in a TS compilation error if they used our types) -- we also don't use enums elsewhere in our public API.

silesky (approved):

PR #912, opened 2023-07-28
Filter out integrations in consent

silesky:
This was a missed requirement -- we need to prune both the integrations objects and the remotePlugins array. Support classic destinations and locally installed action destinations by removing integrations (as well as remote plugins).

### DIFF
--- a/packages/consent/consent-tools/src/domain/create-wrapper.ts
+++ b/packages/consent/consent-tools/src/domain/create-wrapper.ts
@@ -113,7 +113,7 @@ export const createWrapper: CreateWrapper = (createWrapperOptions) => {
           return cdnSettings
         }
 
-        const remotePlugins = omitDisabledRemotePlugins(
+        const { remotePlugins, integrations } = omitDisabledRemotePlugins(
           cdnSettings.remotePlugins,
           cdnSettings.integrations,
           initialCategories,
@@ -121,7 +121,7 @@ export const createWrapper: CreateWrapper = (createWrapperOptions) => {
           shouldEnableIntegration
         )
 
-        return { ...cdnSettings, remotePlugins }
+        return { ...cdnSettings, remotePlugins, integrations }
       }
 
       return ogLoad.call(analytics, settings, {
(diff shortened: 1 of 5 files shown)


packages/consent/consent-tools/src/types/settings.ts
```diff
@@ -64,9 +64,6 @@ export interface CreateWrapperSettings {
   shouldEnableIntegration?: (
     integrationCategories: string[],
     categories: Categories,
-    integrationInfo: Pick<
-      CDNSettingsRemotePlugin,
-      'creationName' | 'libraryName'
-    >
+    integrationInfo: Pick<CDNSettingsRemotePlugin, 'creationName'>
```
silesky, line 67: Unused/neccessary property that is safe to remove.

packages/consent/consent-tools/src/domain/create-wrapper.ts
```diff
@@ -189,4 +191,26 @@ const omitDisabledRemotePlugins = (
...
+          (el) => el.creationName !== creationName
+        )
+        // remove disabled classic destinations and locally-installed action destinations
+        delete acc.integrations[creationName]
+      }
+      return acc
+    },
+    {
+      remotePlugins,
+      integrations: { ...integrations }, // make shallow copy to avoid mutating original
```
oscb, line 212: nit: to not have to do a shallow copy we could construct the `integrations` part on the reducer since we're just putting everything except the disabled ones.

I'm not sure if every `remotePlugin` will have an `integration` object, if so it could also be constructed as it iterates, but I'm assuming from the way this logic is written that it's not the case.

oscb (approved):
Just a non-blocking comment.

packages/consent/consent-tools/src/domain/create-wrapper.ts
silesky: Correct, I am not certain every remotePlugin gets a corresponding integration object.

I went with this because it felt a bit easier for others read, since everything can be nested in the `isPluginEnabled` === false block, whereas with constructing integrations -- that would only happened if `isPluginEnabled` === true -- which felt a little bit of a brain teaser.

PR #915, opened 2023-08-01
Fix batching after page navigation

danieljackins:
This PR addresses a bug where if a user navigated to another page and then back, batching no longer worked. This was due to the 'pageUnloaded' variable never being set back to false.

### DIFF
--- a/packages/browser/src/lib/on-page-change.ts
+++ b/packages/browser/src/lib/on-page-change.ts
@@ -7,13 +7,13 @@
  *
  * adapted from https://stackoverflow.com/questions/3239834/window-onbeforeunload-not-working-on-the-ipad/52864508#52864508,
  */
-export const onPageLeave = (cb: (...args: unknown[]) => void) => {
+export const onPageChange = (cb: (...args: boolean[]) => void) => {
   let unloaded = false // prevents double firing if both are supported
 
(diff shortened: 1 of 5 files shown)


packages/browser/src/lib/on-page-change.ts
```diff
@@ -7,23 +7,23 @@
  *
  * adapted from https://stackoverflow.com/questions/3239834/window-onbeforeunload-not-working-on-the-ipad/52864508#52864508,
  */
-export const onPageLeave = (cb: (...args: unknown[]) => void) => {
+export const onPageChange = (cb: (...args: boolean[]) => void) => {
```
silesky, line 10: Nit:
This sig says that there are multiple bool args.
Type signature should just be (unloaded: boolean) => void

packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -91,10 +91,10 @@ export default function batch(
     }, timeout)
   }
 
-  onPageLeave(() => {
-    pageUnloaded = true
+  onPageChange((unloaded: boolean) => {
```
silesky, line 94: Nit: No need to put "boolean" here; that should be part of the fn signature type

silesky (approved):

PR #928, opened 2023-08-15
feat: support custom global analytics key

oscb:


silesky:
Hey @oscb -- slightly confused by this PR -- I thought we want to allow users to use window.foo.track() rather than window.analytics.track()? So the config option would be like 'analyticsGlobalKey'

oscb:
> I thought the goal was to allow users to use window.foo.track() rather than window.analytics.track()? So the config option would be like 'analyticsGlobalKey', and we save that key in memory -- using it whenever we want to get the global analytics, like how getCDNUrl() works. The fact that this would also be the same reference as the buffer is just an implementation detail.

Yes, it's sort of the idea. I think the name of the config option can be changed to `analyticsGlobalKey`

Per my discussion with Chris we're not yet changing the script part of it as it is going to be a challenge given that the script itself uses `window.analytics` but devs can modify the script and use their custom key and send it through the config (needs docs).

.changeset/mean-apricots-hammer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@segment/analytics-next': minor
+'@internal/consent-tools-integration-tests': minor
+---
+
+Adds `bufferKey` option for setting custom global window buffers
```
silesky, line 6: globalAnalyticsKey

packages/browser/src/browser/__tests__/integration.test.ts
```diff
@@ -212,6 +213,57 @@ describe('Initialization', () => {
     })
   })
 
+  describe('bufferKey', () => {
```
silesky, line 216: globalAnalyticsKey

packages/core/src/task/__tests__/task-group.test.ts
```diff
@@ -0,0 +1,26 @@
+import { createTaskGroup } from '../task-group'
```
silesky, line 1: thanks!

packages/browser/src/core/analytics/index.ts
```diff
@@ -117,6 +117,11 @@ export interface InitOptions {
    * Array of high entropy Client Hints to request. These may be rejected by the user agent - only required hints should be requested.
    */
   highEntropyValuesClientHints?: HighEntropyHint[]
+  /**
+   * Key for the global window property storing the buffered calls
```
silesky, line 121:
```suggestion
   * When using the snippet, this is the key that points to the global analytics instance (e.g. window.analytics).
```

packages/browser/src/browser/index.ts
```diff
@@ -162,11 +164,12 @@ async function flushFinalBuffer(
   analytics: Analytics,
   buffer: PreInitMethodCallBuffer
 ): Promise<void> {
+  const calls = getGlobalAnalytics()
```
silesky, line 167: Can we get rid of this now? restore the old function signature of "popSnippetWindowBuffer"?

packages/browser/src/browser/standalone-analytics.ts
```diff
@@ -59,15 +45,15 @@ function getWriteKey(): string | undefined {
...
       'Failed to load Write Key. Make sure to use the latest version of the Segment snippet, which can be found in your source settings.'
     )
     return
   }
 
-  window.analytics = (await AnalyticsBrowser.standalone(
+  ;(window as any).analytics = (await AnalyticsBrowser.standalone(
     writeKey,
     options
   )) as AnalyticsSnippet
```
silesky, line 59: Should this be something like this:
```ts
AnalyticsBrowser.standalone(
    writeKey,
    options
  )).then(analytics => {
       window[options?.globalAnalyticsKey || 'analytics'] = analytics
  })
```

packages/browser/src/core/analytics/index.ts
```diff
@@ -489,7 +494,7 @@ export class Analytics
 
   noConflict(): Analytics {
     console.warn(deprecationWarning)
-    window.analytics = _analytics ?? this
+    ;(window as any).analytics = _analytics ?? this
```
silesky, line 497: Should this be:
``` 
setGlobalAnalytics(_analytics ?? this)
```

packages/browser/src/browser/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Gets the global analytics instance/buffer
+ * @param key name of the window property where the buffer is stored (default: analytics)
+ * @returns AnalyticsSnippet
+ */
+export function getGlobalAnalytics(): AnalyticsSnippet | undefined {
+  return (window as any)[_globalAnalyticsKey]
+}
+
```
silesky, line 16: Can we add this?
```ts
export function setGlobalAnalytics(analytics: AnalyticsSnippet): void {
   (window as any)[_globalAnalyticsKey] = analytics
}
```

silesky (changes requested):

packages/browser/src/browser/index.ts
oscb: Yea I guess we don't ned this anymore since we store it globally

silesky (approved):

silesky (changes requested):
Small late breaking feedback just noticed!

packages/browser/src/browser/utils.ts
```diff
@@ -0,0 +1,31 @@
+import { AnalyticsSnippet } from './standalone-interface'
```
silesky, line 1: Since file has import side effects -- I think we should name it 'global-analytics-helpers.ts ' or something like that, so nobody puts anything else in it. It could also go in /lib

packages/browser/src/browser/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Stores the global window analytics key
+ */
+let _globalAnalyticsKey = 'analytics'
+
+/**
+ * Gets the global analytics instance/buffer
+ * @param key name of the window property where the buffer is stored (default: analytics)
```
silesky, line 10: Can we change the comment to say "global analytics/buffer"

packages/browser/src/browser/utils.ts
oscb: Good point. It stopped being only an `utils` file

silesky (approved):

.changeset/mean-apricots-hammer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@segment/analytics-next': minor
+'@internal/consent-tools-integration-tests': minor
```
silesky, line 3: this looks like a mistake

packages/browser/src/browser/standalone-analytics.ts
```diff
@@ -1,30 +1,19 @@
-import { Analytics, InitOptions } from '../core/analytics'
 import { AnalyticsBrowser } from '.'
```
silesky, line 1: Will changing this file name change the structure of the UMD build folder? If so, I would do this in a separate PR where we could evaluate the implications, as this is an actual entry point to the app and I would consider it to be a public API.

silesky (changes requested):
Some other stuff ;-)

packages/browser/src/tester/ajs-tester.ts
```diff
@@ -12,11 +12,13 @@ function makeStub(page: playwright.Page) {
       ...args: Parameters<Analytics['register']>
     ): Promise<SerializedContext> {
       return await page.evaluate((innerArgs) => {
-        // @ts-ignore
-        return window.analytics
-          .register(...innerArgs)
-          .then((ctx) => ctx.toJSON())
-        // @ts-ignore
+        return (
```
silesky, line 15: It's actually fine to do declare global  in test code, since it doesn't get included in the actual published library.

examples/with-next-js/pages/partytown/index.tsx
```diff
@@ -44,7 +44,7 @@ const WebWorker: React.FC = () => {
           }}
           value="Track!"
```
silesky, line 45: It's actually fine to do:
```
declare global {
  interface Window {
    analytics: AnalyticsSnippet
  }
}
```
... since this is not a published repo

oscb: I'll leave it as it is for now. But yeah, doesn't matter.

silesky (approved):

PR #932, opened 2023-08-17
fix: sentAt set at batch upload time

oscb:
`sentAt` is not set at batch upload time once per the whole batch. Individual event `sentAt` property is stripped when doing batch uploading.

- [✔︎] I've included a changeset (psst. run `yarn changeset`. Read about changesets [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)).

### DIFF
--- a/packages/browser/src/plugins/segmentio/batched-dispatcher.ts
+++ b/packages/browser/src/plugins/segmentio/batched-dispatcher.ts
@@ -60,13 +60,23 @@ export default function batch(
 
     const writeKey = (batch[0] as SegmentEvent)?.writeKey
 
+    // Remove sentAt from every event as batching only needs a single timestamp
+    const updatedBatch = batch.map((event) => {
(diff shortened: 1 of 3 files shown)


packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -60,13 +60,23 @@ export default function batch(
 
     const writeKey = (batch[0] as SegmentEvent)?.writeKey
 
+    // Remove sentAt from every event as batching only needs a single timestamp
+    const updatedBatch = batch.map((event) => {
+      const { sentAt, ...newEvent } = event as SegmentEvent
+      return newEvent
```
silesky, line 66: Oh I like destructuring to do 'omit'. Neat

silesky (approved):

PR #934, opened 2023-08-19
[Consent management] Add Segment consent changed event

silesky:
- Add `registerOnConsentChanged` setting.
- For OneTrust, dispatch a segment consent changed event whenever OneTrust consent change.
- Add ability to disable consent changed events.
- update README.

Other misc goodies!

### DIFF
--- a/packages/consent/consent-tools/src/domain/create-wrapper.ts
+++ b/packages/consent/consent-tools/src/domain/create-wrapper.ts
@@ -125,18 +125,16 @@ export const createWrapper: CreateWrapper = (createWrapperOptions) => {
         createConsentStampingMiddleware(getValidCategoriesForConsentStamping)
       )
 
-      if (registerOnConsentChanged) {
-        // whenever consent changes, dispatch a new event with the latest consent information
-        registerOnConsentChanged((categories) => {
-          try {
-            validateCategories(categories)
-            sendConsentChangedEvent(analytics, categories)
-          } catch (err) {
-            // Not sure if there's a better way to handle this, but this makes testing a bit easier.
-            console.error(err)
-          }
-        })
-      }
+      // whenever consent changes, dispatch a new event with the latest consent information
+      registerOnConsentChanged?.((categories) => {
+        try {
+          validateCategories(categories)
+          sendConsentChangedEvent(analytics, categories)
+        } catch (err) {
+          // Not sure if there's a better way to handle this, but this makes testing a bit easier.
+          console.error(err)
+        }
+      })
 
       const updateCDNSettings: InitOptions['updateCDNSettings'] = (
         cdnSettings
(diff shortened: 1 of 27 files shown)


packages/consent/consent-tools/src/domain/create-wrapper.ts
```diff
@@ -118,6 +125,19 @@ export const createWrapper: CreateWrapper = (createWrapperOptions) => {
...
+        registerOnConsentChanged((categories) => {
+          try {
+            validateCategories(categories)
+            sendConsentChangedEvent(analytics, categories)
+          } catch (err) {
+            // Not sure if there's a better way to handle this, but this makes testing a bit easier.
+            console.error(err)
+          }
+        })
+      }
```
oscb, line 139: nit: you could do `registerOnConsentChanged?.(`

packages/consent/consent-tools/src/domain/validation/common-validators.ts
```diff
@@ -11,7 +11,7 @@ export function assertIsFunction(
 
 export function assertIsObject(
   val: unknown,
-  variableName: string
+  variableName = 'value'
```
oscb, line 14: shouldn't it keep its typedef to `string`? won't TS restrict this to `value` only?

packages/consent/consent-wrapper-onetrust/src/domain/wrapper.ts
```diff
@@ -1,25 +1,30 @@
...
   integrationCategoryMappings?: CreateWrapperSettings['integrationCategoryMappings']
+  disableConsentChangedEvent?: boolean
 }
 
+/**
+ *
+ * @param analyticsInstance - An analytics instance. Either `window.analytics`, or the instance returned by `new AnalyticsBrowser()` or `AnalyticsBrowser.load({...})`
+ * @param settings - Optional settings for configuring your OneTrust wrapper
+ */
 export const oneTrust = (
```
oscb, line 24: nit: should this be named `withOneTrust` instead? To make sure people understand it's wrapping the `analytics` object?

packages/consent/consent-tools/src/types/settings.ts
```diff
@@ -27,6 +27,36 @@ export interface CreateWrapperSettings {
    **/
   getCategories: () => Categories | Promise<Categories>
 
+  /**
+   * Programatically send a "Segment Consent Preference" event to Segment when consent preferences change.
+   * An event will be sent to Segment whenever this function is called.
```
oscb, line 32: I would specify that this is a "Function to register a listener for consent changes to programatically send ..."

oscb (approved):
Left a few nits, nothing is a blocker imo

packages/consent/consent-tools/src/domain/validation/common-validators.ts
silesky: Nope, it keeps the string type (not a const)

packages/consent/consent-wrapper-onetrust/src/domain/wrapper.ts
silesky: I've been weighing that change as well, I'm glad you chimed in

PR #938, opened 2023-08-24
Consent: add bundled version that can be loaded via script tag

silesky:
This PR creates bundled versions of analytics-consent-wrapper-onetrust:
image

For more info, see the README Updates: https://github.com/segmentio/analytics-next/blob/consent-add-umd-script-build/packages/consent/consent-wrapper-onetrust/README.md#for-snippet-users-windowanalytics

1. Rename oneTrust -> withOneTrust
2. Allow chaining - this makes it easier from a docs / ergonomic perspective 
3. Add  a umd version that _also_ assigns a window.withOneTrust object. (In my research, I found that some libs like Chart.js did this, so I felt like this was an idiomatic approach).
Other benefits:
     - simple / idiomatic (only one bundle to worry about supporting / uploading to CDN)
     - supports UMD users and script users, so lessons support burden
     - having a "umd" folder is what we already do with browser , and it even is something that unpkg expects implicitly


packages/consent/consent-wrapper-onetrust/README.md
```diff
@@ -1,64 +1,124 @@
...
+  <script>
+    !function(){var analytics=window.analytics...
+    ....
+    withOneTrust(analytics).load() // replace analytics.load()
+  </script>
+</head>
 ```
 
-- Use the following initialization code
+#### ⚠️ Reminder: _delete_ `analytics.load('....')` from the original Segment snippet so that only .load() is only called after `withOneTrust` is called. See comment in example above.
```
oscb, line 89: Oh, this is great to callout things, I'm going to borrow this for an RN readme change

oscb (approved):
lgtm!

PR #939, opened 2023-08-25
Move context augmentation to Page enrichment

zikaari:
Part of https://segment.atlassian.net/browse/LIBWEB-1426

Since Page Enrichment is no longer just enriching "page" events, with page information, I wanted to rename this to something generic like "Context Enrichment", but wasn't certain if this is safe. Open to discussion.

The original implementation suggests something about falling back to "options" if "context" is not there. Not sure what that means, so please help me identify what's the intended outcome of that design so I can double-check that behaviour is not impacted.


packages/core/src/events/interfaces.ts
```diff
@@ -477,3 +471,11 @@ export type UserTraits = BaseUserTraits & {
  * Traits are pieces of information you know about a user or group.
  */
 export type Traits = UserTraits | GroupTraits
+
+export type Campaign = {
+  name: string
```
silesky, line 476: Pretty sure every single one of these should be optional

silesky (changes requested):

silesky:
@zikaari Definitely an improvement architecturally! 

 > I wanted to rename this to something generic like "Context Enrichment", but wasn't certain if this is safe. Open to discussion.

Cool! Yeah, I see this plugin is about augmenting the event with information from the user's environment-- so maybe EnvironmentAugmentation plugin would be a good candidate -- but "page" can also be interpreted as a kind of word for environment :-)

I mainly don't like the word 'enrichment' because this plugin isn't a type:enrichment plugin -- so it's  v confusing.

packages/core/src/events/interfaces.ts
```diff
@@ -477,3 +471,11 @@ export type UserTraits = BaseUserTraits & {
  * Traits are pieces of information you know about a user or group.
  */
 export type Traits = UserTraits | GroupTraits
+
+export type Campaign = {
```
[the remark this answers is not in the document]
zikaari: I copied pasted most things as-is to minimize any collateral damage, including this - but I've fixed it up now that you mentioned it

zikaari:
@pooyaj Do you know any dangers of renaming a plugin on top of your head? Not just the symbol name, but also `Plugin.name` property.

If you don't feel confident, we can leave renaming it to some other time.

packages/browser/src/plugins/page-enrichment/index.ts
```diff
@@ -70,45 +122,134 @@ export function pageDefaults(): PageDefault {
...
+    // @ts-ignore
+    const locale = navigator.userLanguage || navigator.language
+
+    if (typeof evtCtx.locale === 'undefined' && typeof locale !== 'undefined') {
+      evtCtx.locale = locale
+    }
+
+    evtCtx.library ??= {
+      name: 'analytics.js',
+      version: `${getVersionType() === 'web' ? 'next' : 'npm:next'}-${version}`,
```
silesky, line 226: For web, this should be `next-${version}` to match the old logic?

silesky (approved):

silesky:
@zikaari change to core package (interface) should have a another changeset

PR #945, opened 2023-09-01
Non-blocking lazy destinations

zikaari:
This patch changes the architecture around destinations work. Previously AJS would use the waterfall like method when loading destinations (and other plugins) where it'd wait for their loading sequence to finish before it'd start processing events. With this patch the destinations load in "background" and AJS will start processing events the moment critical plugins are ready.

The events will get forwarded to each destination and it can send the event whenever it becomes ready (assuming it does). This way if 4 out of 6 destinations are ready in 0.2s after page load, and the other 2 take 0.8s, then those 4 won't be bottlenecked by the slower ones.

Learn more: [Lazy destinations SDD](https://docs.google.com/document/d/10eVBIHKS2ivd-CFw8KHD4RU9eMYv9mYFuNuxk4gWSk0/)


packages/browser/src/core/analytics/index.ts
```diff
@@ -524,7 +525,7 @@ export class Analytics
   async ready(
     callback: Function = (res: Promise<unknown>[]): Promise<unknown>[] => res
   ): Promise<unknown> {
-    return Promise.all(
+    return Promise.allSettled(
```
silesky, line 528: Since `allSettled` is a newish browser thing, we might want to try a polyfill
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

```ts
type AllSettledResult<T> = {
  status: 'fulfilled' | 'rejected'
  reason?: Error
  value?: T
}

// promise.allSettled "polyfill"
export const allSettled = <T>(
  promises: Promise<T>[]
): Promise<AllSettledResult<T>[]> => {
  return Promise.all(
    promises.map((promise) => {
      // Per spec, Promise.allSettled([Promise.reject('oops'), 2] ->
      // [{status: "rejected", reason: "oops"}, { "status": "fulfilled", "value": 2 }}
      return Promise.resolve(promise)
        .then(
          (value) =>
            ({
              value,
              status: 'fulfilled',
            } as AllSettledResult<T>)
        )
        .catch(
          (reason) =>
            ({
              reason,
              status: 'rejected',
            } as AllSettledResult<T>)
        )
    })
  )
}



----

import { allSettled } from './all-settled'

describe('all settled', () => {
  it('should return a promise ', async () => {
    const result = allSettled([Promise.resolve('success')])
    expect(result).toBeInstanceOf(Promise)
  })

  it('should handle successes', async () => {
    const result = await allSettled([
      Promise.resolve('success'),
      Promise.resolve('success 2'),
    ])
    const response = [
      { status: 'fulfilled', value: 'success' },
      { status: 'fulfilled', value: 'success 2' },
    ]
    expect(result).toEqual(response)
  })

  it('should handle rejects', async () => {
    const result = await allSettled([
      Promise.reject('oops'),
      Promise.reject('oops 2'),
    ])
    const response = [
      { status: 'rejected', reason: 'oops' },
      { status: 'rejected', reason: 'oops 2' },
    ]
    expect(result).toEqual(response)
  })

  it('should handle mixes of fulfilled/reject', async () => {
    const result = await allSettled([
      Promise.reject('oops'),
      Promise.resolve(2),
    ])
    const response = [
      { status: 'rejected', reason: 'oops' },
      { status: 'fulfilled', value: 2 },
    ]
    expect(result).toEqual(response)
  })

  it('should handle mixes of fulfilled/reject / respect order', async () => {
    const result = await allSettled([
      Promise.resolve(2),
      Promise.reject('oops'),
    ])
    const response = [
      { status: 'fulfilled', value: 2 },
      { status: 'rejected', reason: 'oops' },
    ]
    expect(result).toEqual(response)
  })

  it('should resolve any non-promises passed', async () => {
    const result = await allSettled([2 as any, Promise.reject('oops 2')])
    const response = [
      { status: 'fulfilled', value: 2 },
      { status: 'rejected', reason: 'oops 2' },
    ]
    expect(result).toEqual(response)
  })

  it('should never call catch block', async () => {
    const result = await allSettled([
      Promise.reject('oops'),
      Promise.reject('oops 2'),
    ]).catch(() => {
      throw new Error('fail')
    })
    const response = [
      { status: 'rejected', reason: 'oops' },
      { status: 'rejected', reason: 'oops 2' },
    ]
    expect(result).toEqual(response)
  })
})


```

packages/browser/src/browser/index.ts
```diff
@@ -180,7 +180,6 @@ async function registerPlugins(
   writeKey: string,
   legacySettings: LegacySettings,
   analytics: Analytics,
-  opts: InitOptions,
   options: InitOptions,
```
oscb, line 183: Woah, this is so confusing, why was it this way?

packages/browser/src/browser/index.ts
```diff
@@ -317,8 +317,13 @@ async function loadAnalytics(
   const retryQueue: boolean =
     legacySettings.integrations['Segment.io']?.retryQueue ?? true
 
-  const opts: InitOptions = { retryQueue, ...options }
-  const analytics = new Analytics(settings, opts)
+  options = {
+    retryQueue,
+    destinationTimeout: 3000,
```
oscb, line 322: Can we put `3000` into a const at the top of the file to make it more visible?

packages/browser/src/browser/index.ts
```diff
@@ -349,8 +353,14 @@ async function loadAnalytics(
     await analytics.queryString(term).catch(console.error)
   }
 
-  analytics.initialized = true
-  analytics.emit('initialize', settings, options)
+  analytics
+    .ready()
+    .then(() => {
+      analytics.initialized = true
+      console.log('emitting initialize')
```
oscb, line 360: just a reminder to remove the console.log in the final PR

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -283,7 +283,7 @@ export class LegacyDestination implements DestinationPlugin {
       this.integration.initialize()
     }
 
-    return this.onInitialize!.then(() => {
+    return this.initializePromise.promise.then(() => {
```
oscb, line 286: Why does this one in particular triggers and waits for `initialize`? It is already done in the `load()` right?

nit: Can we convert it into `await this.initializePromise.promise` instead of this syntax to keep it consistent to where we await `ready`?

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -139,22 +140,20 @@ export class LegacyDestination implements DestinationPlugin {
...
 
-    this.onReady = new Promise((resolve) => {
-      const onReadyFn = (): void => {
-        this._ready = true
-        resolve(true)
-      }
+    setTimeout(() => {
+      const e = 'Destination timed out'
+      this.initializePromise.reject(e)
+      this.readyPromise.reject(e)
```
oscb, line 146: Can we add a metric here?

oscb: I'm also curious about what happens in JS when an already resolved promise rejects. it could happen here that `initializedPromise` was already resolved then `readyPromise` times out and makes everything reject.

oscb: Should this set `_ready` and `_initialized` to `false`?

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -139,22 +140,20 @@ export class LegacyDestination implements DestinationPlugin {
...
+    }, this.options.destinationTimeout!)
 
-      this.integration!.once('ready', onReadyFn)
+    this.integration!.once('ready', () => {
+      this._ready = true
+      this.readyPromise.resolve()
```
oscb, line 151: to make the promises more sound maybe we should chain the state of `_ready` and `_initialized` to them during construction

```ts
constructor(...) {
	//...
	this.readyPromise.then(() => { this._ready = true }).catch(() => { this._ready = false });
	// Same for initializePromise
}
```

That way you only have to care about resolving or rejecting promises throughout the code.

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -116,11 +117,11 @@ export class LegacyDestination implements DestinationPlugin {
   }
 
   ready(): Promise<unknown> {
-    return this.onReady ?? Promise.resolve()
+    return this.readyPromise.promise
   }
 
   async load(ctx: Context, analyticsInstance: Analytics): Promise<void> {
-    if (this._ready || this.onReady !== undefined) {
+    if (this._ready || this.readyPromise.settled) {
```
oscb, line 124: You could bypass the need to add `settled` by having `_ready` be `boolean | undefined`. Set it to `true` or `false` at `resolve` or `reject`. If it is `undefined` it is still unsettled.

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -30,14 +32,18 @@ export class ActionDestination implements DestinationPlugin {
...
+  options: InitOptions
+
+  private loadPromise = createDeferred<unknown>()
 
   middleware: DestinationMiddlewareFunction[] = []
 
   action: Plugin
 
-  constructor(name: string, action: Plugin) {
+  constructor(name: string, action: Plugin, options: InitOptions) {
```
oscb, line 43: Do we need the whole options object? Can we define instead a subset object of options just for plugins?
I prefer the later since it is easier to limit what we expose, and also to limit the impact if we ever have to change things in the client options.

oscb: Should also be optional

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -79,6 +85,7 @@ export class ActionDestination implements DestinationPlugin {
         transformedContext = await this.transform(ctx)
       }
 
+      await this.ready()
```
oscb, line 88: Doesn't this need to happen before the `transform` call above?

packages/browser/src/plugins/ajs-destination/index.ts
oscb: Or better yet could be modeled into an stateflow-like enum: Created -> Initialized -> Ready (+ error states)

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -196,7 +217,7 @@ export async function remoteLoader(
   settings: LegacySettings,
   userIntegrations: Integrations,
   mergedIntegrations: Record<string, JSONObject>,
-  obfuscate?: boolean,
+  options: InitOptions,
```
oscb, line 220: same as above about having a custom plugin options object (we can add obfuscate to it)

packages/browser/src/plugins/remote-loader/index.ts
oscb: I would also argue that the timeout should be an option only for the `load` method, not for the whole plugin defined at construction time (do we let `.load` be called multiple times? when is the constructor of the plugin called?)

oscb (changes requested):

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -73,8 +74,8 @@ export class LegacyDestination implements DestinationPlugin {
 
   private _ready = false
   private _initialized = false
-  private onReady: Promise<unknown> | undefined
-  private onInitialize: Promise<unknown> | undefined
+  private readyPromise = createDeferred<void>()
```
silesky, line 77: These are more like "wrappers around Promise" so maybe "readyDeferred" would be a better name? "deferred" is a common name for this object.

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: > I'm also curious about what happens in JS when an already resolved promise rejects

It's a no-op in JS. A settled promise will stay in the same state, further `resolve` or `rejects` don't do anything.

> Can we add a metric here?

Done in the other patch

packages/browser/src/lib/create-deferred.ts
```diff
@@ -4,13 +4,18 @@
...
     reject = _reject
   })
+
+  promise.finally(() => (settled = true))
+
   return {
     resolve,
     reject,
     promise,
+    settled,
```
silesky, line 19: This doesn't look like it actually works as expected, as `settled` is a value and not a function that is closing over the value. But rather than fixing it, maybe we can do without adding this to the API (per @oscb comments below?)

packages/browser/src/plugins/ajs-destination/index.ts
silesky: @oscb  Doesn't the meaning of .ready = true just mean ".ready has been fired", which should happens on either resolve _or_ reject? 

I feel like making .ready a tristate variable is a bit confusing (undefined, true, or false)?

packages/browser/src/browser/index.ts
silesky: @oscb This has been confusing me for ... checks wash... since the very first time I saw this codebase almost 2 years ago. I feel like we should break out the champagne once this gets merged 😄

packages/browser/src/core/analytics/index.ts
zikaari: Done

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: It's been like this for a long time - I didn't want to risk messing with it

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: Please see the latest push for revision to this

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: Please see the latest push for revision to this

packages/browser/src/plugins/remote-loader/index.ts
zikaari: Not necessary, `transform` is internal function, not a proxy to any of the plugin's function

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: `readyDeferred` doesn't really communicate the meaning of what's inside at a glance - `readyState`? then it'd be `readyState.promise`, `readyState.resolve()`

packages/browser/src/lib/create-deferred.ts
zikaari: very very good catch! although this will be removed since @oscb suggestion voids the purpose anyway

packages/browser/src/plugins/ajs-destination/index.ts
silesky: Sounds good to me

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -112,15 +121,15 @@ export class LegacyDestination implements DestinationPlugin {
...
   }
 
   ready(): Promise<unknown> {
-    return this.onReady ?? Promise.resolve()
+    return this.readyPromise.promise
   }
 
   async load(ctx: Context, analyticsInstance: Analytics): Promise<void> {
-    if (this._ready || this.onReady !== undefined) {
+    if (typeof this._ready === 'boolean') {
```
silesky, line 132: this._ready === undefined would be more idiomatic?

chrisradek:
I think this is a great start. I have a few high-level comments/questions:

1. Do we need to make any changes to how classic integrations load? It looks like the existing `load()` method for classic integrations doesn't wait for anything except our own 1st party code to load, and has it's own event buffering in place. Assuming I'm not misreading the existing code, I worry this might be too aggressive for classic integrations since I _think_ they aren't giving us any problems with never-ending loads.

2. It looks like we add the same timeout to all action destination actions, regardless of what lifecycle hook they are configured for. While classic integrations are always destinations, action destination actions can be any plugin type.
This has me a little worried that we might end up being too aggressive with our values here. Let's take the google-analytics-4 destination as an example - it has a `before` action (`setConfigurationFields`). We treat any non-destination load failure as fatal, so our timeout could cause analytics.js to completely fail to initialize if say it took longer than 3 seconds for the 3rd party library to load.

I think my initial inclination is to only apply the timeout for destination plugins for now. I think further enhancements could be to potentially remove enrichment plugins that fail due to the timeout error rather than let the `register` function re-throw, and then figure out what to do about `before` plugins since (spit-balling ideas - maybe set the integration associated with that plugin to false but still load analytics.js, not sure...)

packages/browser/src/core/analytics/index.ts
silesky: @zikaari is this done?

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -86,6 +89,13 @@ export class ActionDestination implements DestinationPlugin {
           methodName,
           type: 'action',
         })
+
+        if (!(await this.ready())) {
+          throw new Error(
+            'Something prevented the destination from getting ready'
+          )
+        }
```
chrisradek, line 97: I'd recommend moving this above the metric since we use the metric to understand how often destinations are called. If we introduced a regression here (where this.ready() never resolved), our metrics would look unchanged right now.

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -86,6 +89,13 @@ export class ActionDestination implements DestinationPlugin {
           methodName,
           type: 'action',
         })
+
+        if (!(await this.ready())) {
+          throw new Error(
+            'Something prevented the destination from getting ready'
```
chrisradek, line 95: Does this error get surfaced to the user somehow? ALso might be useful to include the name of the destination in this message.

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -113,25 +123,41 @@ export class ActionDestination implements DestinationPlugin {
...
         integrationName: this.action.name,
         methodName: 'load',
         type: 'action',
       })
-      return await this.action.load(ctx, analytics)
+
+      const loadP = this.action.load(ctx, analytics)
+
+      this.loadPromise.resolve(await loadP)
+      return loadP
```
chrisradek, line 151: Why do we need to create `loadPromise` from `createDeferred`? Seems like we could set it to `this.action.load`, and then at the top of this function just return `this.loadPromise` if it exists.

zikaari: The deferred promise is used in the `ready` function. This handles a case where ready might be called before `load`

packages/browser/src/lib/create-deferred.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Return a promise that can be externally resolved
+ */
+export const createDeferred = <T>() => {
```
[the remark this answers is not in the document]
zikaari: Done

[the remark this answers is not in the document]
silesky: Did you push?

packages/browser/src/plugins/remote-loader/index.ts
zikaari: Good catch! done

packages/browser/src/plugins/remote-loader/index.ts
zikaari: No it doesn't make to the surface, gets suppressed in one of the catch blocks in the call stack

packages/core/src/queue/event-queue.ts
```diff
@@ -49,25 +50,22 @@ export abstract class CoreEventQueue<
     plugin: Plugin,
     instance: CoreAnalytics
   ): Promise<void> {
-    await Promise.resolve(plugin.load(ctx, instance))
-      .then(() => {
-        this.plugins.push(plugin)
+    if (plugin.type === 'destination' && plugin.name !== 'Segment.io') {
+      plugin.load(ctx, instance).catch((err) => {
+        this.failedInitializations.push(plugin.name)
+        this.emit('initialization_failure', plugin)
```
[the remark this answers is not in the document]
zikaari: Done

zikaari:
@silesky @chrisradek Please have another look, the last 2 commits are the actual new changes I made

packages/browser/src/core/analytics/index.ts
silesky: I still see that we're not using a polyfill here! I don't think we can get away with using the raw allSettled as it doesn't get added in the normal polyfills.

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -113,25 +123,41 @@ export class ActionDestination implements DestinationPlugin {
...
     } catch (error) {
       recordIntegrationMetric(ctx, {
         integrationName: this.action.name,
         methodName: 'load',
         type: 'action',
         didError: true,
       })
+
+      this.loadPromise.resolve(error)
```
silesky, line 160: Nit: Why not reject? Looks a bit strange!

Then wouldn't need the error instanceof Error thing

zikaari: Very good catch mate. Thank you!

This was by design back in the day when I didn't know about the trick of using `Promise.allSettled([promise])` based "proxying" in `createDeffered` utility. Using `promise.finally` directly there, and then rejecting the same promise somewhere else would cause "Unhandled promise rejection" even if you added `catch` trap. That was because `catch` would've been added after `finally` hook was.

Now, with the new technique this behaviour is not exhibitted since we "fork" the promise instead of attachting traps directly.

Code has been updated and loadPromise now rejects. Thanks again for highlighting this.

packages/browser/src/core/analytics/index.ts
zikaari: Thanks for pointing out - fixed

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -113,25 +123,40 @@ export class ActionDestination implements DestinationPlugin {
...
       recordIntegrationMetric(ctx, {
         integrationName: this.action.name,
         methodName: 'load',
         type: 'action',
       })
-      return await this.action.load(ctx, analytics)
+
+      const loadP = this.action.load(ctx, analytics)
+
+      this.loadPromise.resolve(await loadP)
```
silesky, line 149: Do we want this await here, instead of just loadP? Promise<Promise<val>> unwraps to Promise<val> automatically, so should resolve to the same thing.

Looks a bit strange to await a promise result but then return the underlying promise -- I want to just make sure we aren't waiting if we don't need to.

LMK if missing something!

chrisradek: Yeah I got a bit caught up by this too. I think the await is here so that if load throws an error it's catch by the catch handler and we can record the metric.

packages/browser/src/core/analytics/index.ts
chrisradek: Can you remind me why this change is needed again? I thought it was originally because we were changing the initialization/ready ordering, but I don't actually see that being the case anymore in the analytics.js loader function.

Our docs currently say this:
> The ready method isn’t invoked if any Destination throws an error (for example, for an expired API key, incorrect settings configuration, or when a Destination is blocked by the browser) during initialization.

So, do we even need this change?

packages/browser/src/plugins/remote-loader/index.ts
silesky: Ah that makes sense

packages/browser/src/core/analytics/index.ts
zikaari: Another good catch, we made this change when `ready` could throw - but that was reverted, however this wasn't. I'm gonna revert this back.

packages/browser/src/browser/standalone.ts
```diff
@@ -83,24 +83,28 @@ if (globalAnalyticsKey) {
...
-    document.body.appendChild(script)
-  }
-
-  script.onload = function (): void {
-    attempt(install)
+    }
   }
+  // load polyfills in order to get AJS to work with old browsers
+  loadScript(
+    'https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.7.0/polyfill.min.js',
```
silesky, line 101: I think we should add allSettled in generic utils.
Extending this polyfill code has a couple issues:
1. potential CSP issues with jsdeliver
2. 
We should add it in code, the shouldPolyfill thing is not reliable -- look at the support table (allSettled was only added in FF 71, but we only polyfill below 46)
image

The entire polyfill pattern that was designed here seems very hacky/fragile/manual and I am in favor of re-examining it (I even wrote tickets around it)

packages/generic-utils/src/create-deferred/create-deferred.ts
```diff
@@ -4,13 +4,20 @@
...
   let reject!: (reason: any) => void
+  let settled = false
   const promise = new Promise<T>((_resolve, _reject) => {
     resolve = _resolve
     reject = _reject
   })
+
+  // we "fork" the original promise instead of directly attatching `finally` to
+  // it as it interferes with natural flow of then and catch in the user code
+  Promise.allSettled([promise]).then(() => (settled = true), void 0)
```
chrisradek, line 15: Can we get rid of using allSettled here by doing something like this?
```suggestion
  Promise.all([promise]).finally(() => (settled = true), void 0)
```
Doesn't seem like there's an issue with using `finally` in general, just directly on the promise.

packages/generic-utils/src/create-deferred/create-deferred.ts
```diff
@@ -4,13 +4,22 @@
...
   let reject!: (reason: any) => void
+  let settled = false
   const promise = new Promise<T>((_resolve, _reject) => {
     resolve = _resolve
     reject = _reject
   })
+
+  // we "fork" the original promise instead of directly attatching `finally` to
+  // it as it interferes with natural flow of then and catch in the user code
+  Promise.all([promise])
```
silesky, line 15: Can we add a quick unit test for this?

packages/core/src/queue/event-queue.ts
```diff
@@ -49,25 +50,24 @@ export abstract class CoreEventQueue<
...
+        console.warn(plugin.name, err)
+
+        ctx.log('warn', 'Failed to load destination', {
+          plugin: plugin.name,
+          error: err,
+        })
+
+        this.plugins = this.plugins.filter((p) => p.name !== plugin.name)
```
chrisradek, line 64: This probably works fine in practice, but we might want to do a strict equality check instead. Reason being, each action of an action destination is implemented as a separate plugin, and they share a name (we use the `alternativeNames` array to distinguish them further if I recall correctly.) So that would mean if for some reason one action failed, but another one succeeded, and they belonged to the same destination, we'd end up with both removed if we just check the name.

This is probably not a real issue in practice since I think currently all those actions would share a load method, but the code change to handle this seems straightforward so may be worth the added protection.
```suggestion
        this.plugins = this.plugins.filter((p) => p === plugin)
```

chrisradek (commented):
Looking good! I just had one minor comment on the failed plugin filtering, then Seth has a comment on adding a test. Other than that though, I think it looks good to go.

packages/generic-utils/src/create-deferred/create-deferred.ts
```diff
@@ -4,13 +4,22 @@
...
-    resolve = _resolve
-    reject = _reject
+    resolve = (...args) => {
+      settled = true
+      _resolve(...args)
+    }
+    reject = (...args) => {
+      settled = true
+      _reject(...args)
+    }
```
chrisradek, line 16: Ah, nice approach!

chrisradek (approved):

packages/generic-utils/src/create-deferred/create-deferred.ts
zikaari: Can't believe it didn't come to mind before

silesky (approved):

PR #946, opened 2023-09-06
Add sentAt field to node for Segment events

danieljackins:
This PR updates node to include a sentAt field along with it's payload.

### DIFF
--- a/packages/node/src/__tests__/http-integration.test.ts
+++ b/packages/node/src/__tests__/http-integration.test.ts
@@ -151,7 +151,7 @@ describe('Method Smoke Tests', () => {
               "userId": "my_user_id",
             },
           ],
-          "sentAt": "Any<String>",
+          "sentAt": Any<String>,
         }
       `
       )
@@ -188,7 +188,7 @@ describe('Method Smoke Tests', () => {
               "userId": "foo",
             },
           ],
-          "sentAt": "Any<String>",
+          "sentAt": Any<String>,
         }
       `
       )
@@ -219,7 +219,7 @@ describe('Method Smoke Tests', () => {
               "type": "page",
             },
           ],
-          "sentAt": "Any<String>",
+          "sentAt": Any<String>,
         }
       `
       )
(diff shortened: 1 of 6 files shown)


packages/node/src/plugins/segmentio/publisher.ts
```diff
@@ -198,6 +198,14 @@ export class Publisher {
           return batch.resolveEvents()
         }
 
+        const currentTime = new Date()
```
silesky, line 201: We just want sentAt at the root, we don't need to append this field into every individual event. The integrations pipeline takes care of that.

silesky (changes requested):
🙏🏼

packages/node/src/__tests__/integration.test.ts
```diff
@@ -240,6 +240,15 @@ describe('screen', () => {
...
 
 describe('track', () => {
+  it('sends sentAt field in payload', async () => {
+    const analytics = createTestAnalytics({
+      httpClient: testClient,
+    })
+    const track = resolveCtx(analytics, 'track')
+    analytics.track({ event: 'hello', userId: 'foo' })
+    await track
+    expect(makeReqSpy.mock.calls[0][0].data.sentAt).toBeDefined()
```
silesky, line 250: Can we assert that this is a date?

silesky (approved):

packages/node/src/__tests__/http-integration.test.ts
```diff
@@ -21,7 +21,7 @@ const snapshotMatchers = {
     }
   },
   get defaultReqBody() {
-    return { batch: [snapshotMatchers.batchEvent] }
+    return { batch: [snapshotMatchers.batchEvent], sentAt: expect.any(String) }
```
silesky, line 24: Should we do:
```
expect.stringMatching(
        /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/
      )
```

silesky (changes requested):
Small test tweak

packages/node/src/__tests__/integration.test.ts
silesky, line 250: Can we assert that this is a date (or at least a string?)

silesky (approved):

PR #947, opened 2023-09-06
Consent: Refactor e2e tests to use umd bundle, simplify server configuration

silesky:
Big improvements to consent integration tests 
- Make integration tests run mostly offline
- Use wdio static server module
- there are edge cases around analytics stale references that testing the umd bundle helps with


oscb (approved):
lgtm! We should implement this for some core integration tests!

PR #956, opened 2023-09-21
Collect timezone and allow userAgentData to be overridden

danieljackins:

### DIFF
--- a/packages/browser/src/plugins/page-enrichment/__tests__/index.test.ts
+++ b/packages/browser/src/plugins/page-enrichment/__tests__/index.test.ts
@@ -267,6 +267,29 @@ describe('pageDefaults', () => {
 describe('Other visitor metadata', () => {
   let options: SegmentioSettings
   let analytics: Analytics
+  ;(window.navigator as any).userAgentData = {
+    ...lowEntropyTestData,
+    getHighEntropyValues: jest
+      .fn()
+      .mockImplementation((hints: string[]): Promise<UADataValues> => {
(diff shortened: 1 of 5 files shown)


silesky:
@danieljackins CI is failing FYI

packages/browser/src/plugins/page-enrichment/index.ts
```diff
@@ -214,6 +215,14 @@ class PageEnrichmentPlugin implements Plugin {
 
     evtCtx.userAgent = navigator.userAgent
 
+    try {
+      evtCtx.userAgentData = await clientHints(
+        this.instance.options.highEntropyValuesClientHints
```
chrisradek, line 220: Similar to the original PR, can we cache the client hints during the plugin load? These fields won't change, so no reason to perform a potentially expensive operation on every event :smile:

packages/browser/src/plugins/page-enrichment/__tests__/index.test.ts
```diff
@@ -283,6 +288,11 @@ describe('Other visitor metadata', () => {
     }
   })
 
+  it('should add .timezone', async () => {
+    const ctx = await analytics.track('test')
+    assert(ctx.event.context?.timezone)
```
silesky, line 293: can we assert that this is a string / actual TZ?

silesky (approved):
LGTM

chrisradek (approved):
:shipit:

PR #963, opened 2023-10-03
Adds missing return types and adds context-batch type

thehme:
These changes are needed to resolve compile errors when consuming this package in an app with TS 4.8.4. The errors that were being encountered were:

```
node_modules/@segment/analytics-node/src/app/analytics-node.ts(97,11): error TS7010: '_dispatch', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/@segment/analytics-node/src/lib/abort.ts(63,40): error TS7011: Function expression, which lacks return-type annotation, implicitly has an 'any[] | readonly [AbortSignal | AbortSignal, Timeout]' return type.
node_modules/@segment/analytics-node/src/plugins/segmentio/publisher.ts(178,39): error TS2339: Property 'message' does not exist on type '{ success: true; } | { success: false; message: string; }'.
  Property 'message' does not exist on type '{ success: true; }'.
```

Initially, to confirm that changes to the package would fix the issue, I updated the code in the `node_modules` folder and was able to successfully compile my code. This was done using the `1.1.1` version of the `@segment/analytics-node` package.

Once I confirmed the changes worked, I made the changes in the package code (forked repo), ran `yarn changeset` successfully, also `yarn test:node-int`, and then committed the changes, which ran the linter successfully, and finally pushed my branch, which ran the tests again - successfully.

- [ x] I've included a changeset (psst. run `yarn changeset`. Read about changesets [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)).

### DIFF
--- a/packages/node/src/plugins/segmentio/publisher.ts
+++ b/packages/node/src/plugins/segmentio/publisher.ts
@@ -174,9 +174,11 @@ export class Publisher {
       return ctxPromise
     } else {
       // this should only occur if max event size is exceeded
-      ctx.setFailedDelivery({
-        reason: new Error(fbAddStatus.message),
-      })
+      if ('message' in fbAddStatus) {
+        ctx.setFailedDelivery({
+          reason: new Error(fbAddStatus.message),
+        })
+      }
       return Promise.resolve(ctx)
     }
   }
--- a/.changeset/popular-eggs-begin.md
+++ b/.changeset/popular-eggs-begin.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-node': patch
+---
+
+Adds missing return types
--- a/packages/node/src/app/analytics-node.ts
+++ b/packages/node/src/app/analytics-node.ts
@@ -94,7 +94,10 @@ export class Analytics extends NodeEmitter implements CoreAnalytics {
     return timeout ? pTimeout(promise, timeout).catch(() => undefined) : promise
   }
 
-  private _dispatch(segmentEvent: SegmentEvent, callback?: Callback) {
+  private _dispatch(
+    segmentEvent: SegmentEvent,
+    callback?: Callback
+  ): undefined {
     if (this._isClosed) {
       this.emit('call_after_close', segmentEvent as SegmentEvent)
       return undefined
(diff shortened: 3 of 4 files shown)


packages/node/src/plugins/segmentio/publisher.ts
```diff
@@ -174,9 +174,11 @@ export class Publisher {
       return ctxPromise
     } else {
       // this should only occur if max event size is exceeded
-      ctx.setFailedDelivery({
-        reason: new Error(fbAddStatus.message),
-      })
+      if (!fbAddStatus.success) {
+        ctx.setFailedDelivery({
```
silesky, line 178: This diff gives me pause. If we need this to satisfy the type checker, I'd rather not make this change =)

packages/node/src/plugins/segmentio/context-batch.ts
```diff
@@ -19,9 +30,7 @@ export class ContextBatch {
   constructor(maxEventCount: number) {
     this.maxEventCount = Math.max(1, maxEventCount)
   }
-  public tryAdd(
-    item: PendingItem
-  ): { success: true } | { success: false; message: string } {
+  public tryAdd(item: PendingItem): ContextBatchResult {
```
silesky, line 33: Confused about why this would be neccessary =S.

packages/node/src/plugins/segmentio/publisher.ts
thehme: I thought adding a type for the different return types would be useful, but we can also resolve this error by using:

```
      if ('message' in fbAddStatus) {
        ctx.setFailedDelivery({
          reason: new Error(fbAddStatus.message),
        })
      }
```

packages/node/src/plugins/segmentio/context-batch.ts
thehme: Related to above, I thought defining the return types would be useful, but we can resolve the error:

```
node_modules/@segment/analytics-node/src/plugins/segmentio/publisher.ts(178,39): error TS2339: Property 'message' does not exist on type '{ success: true; } | { success: false; message: string; }'.
  Property 'message' does not exist on type '{ success: true; }'.
```

By using:

```
      if ('message' in fbAddStatus) {
        ctx.setFailedDelivery({
          reason: new Error(fbAddStatus.message),
        })
      }
```

PR #967, opened 2023-10-11
add cloudflare worker tests for analytics-node

chrisradek:
This PR adds cloudflare worker tests to the node-integration-tests package. This uses the latest wrangler to build and run the tests, which is using workerd under the hood. Basically as close to the real thing as we can get.


silesky (approved):

silesky:
Looks good!  Can you update the README? https://github.com/segmentio/analytics-next/tree/master/packages/node-integration-tests/README.md

packages/node-integration-tests/package.json
```diff
@@ -10,16 +10,20 @@
     "durability": "yarn ts-node src/durability-tests/durability-tests.ts",
     "concurrently": "yarn run -T concurrently",
     "ts-node": "yarn run -T ts-node",
-    "test": "yarn perf && yarn durability"
+    "test": "yarn perf && yarn durability",
+    "cloudflare-workers": "yarn run -T jest src/cloudflare"
```
silesky, line 14: Nit: maybe "test:cf-workers" would be more consistent?

packages/node-integration-tests/src/common/mock-segment.ts
```diff
@@ -0,0 +1,78 @@
...
+function isBatchRequest(req: IncomingMessage) {
+  if (req.url?.endsWith('/v1/batch')) {
+    return true
+  }
+  return false
+}
+
+type BatchHandler = (batch: any) => void
+
+export class MockSegmentServer {
```
silesky, line 25: You know we have this, which also is a mock segment server?  https://github.com/segmentio/analytics-next/tree/master/packages/node-integration-tests/src/server

silesky:
Approving because not of the suggestions are blocking. This looks awesome!

.github/workflows/ci.yml
```diff
@@ -27,6 +27,27 @@ jobs:
             ${{ runner.os }}-turbo-
       - run: yarn turbo run --filter='./packages/node*' lint
       - run: yarn turbo run --filter='./packages/node*' test
+  analytics-node-cf-workers:
+    name: 'analytics-node QA (Cloudflare Workers)'
```
silesky, line 31: Might it make sense to just fold this into the consent-intg-tests job and run everything on node 18? Naming is a bit weird now with a second category of  integration tests

packages/node-integration-tests/package.json
chrisradek: Totally! Still make sense to keep cloudflare workers out of the main `test` script? Originally had them in, only reason to keep out is that they require a higher node.js runtime that the other tests do.

silesky: Maybe we just remove the basic test script altogether?

chrisradek:
Thanks for the feedback! Went ahead and implemented all your suggestions, merging!

PR #971, opened 2023-10-16
Capture action plugin metrics

zikaari:
Just like how we capture metrics relating to performance of classic integrations, this patch now enables capturing of metrics related to action destinations as well.

### DIFF
--- a/.changeset/metal-drinks-repeat.md
+++ b/.changeset/metal-drinks-repeat.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': minor
+---
+
+Capture action plugin metrics
--- a/packages/browser/src/plugins/remote-loader/__tests__/action-destination.test.ts
+++ b/packages/browser/src/plugins/remote-loader/__tests__/action-destination.test.ts
@@ -80,8 +80,6 @@ describe('ActionDestination', () => {
 
     await ajs.ready()
 
-    expect(ajs.ctx?.stats.metrics).toHaveLength(1)
-
     expect(ajs.ctx?.stats.metrics[0]).toMatchObject(
       expect.objectContaining({
         metric: 'analytics_js.action_plugin.invoke',
--- a/packages/browser/src/plugins/remote-loader/__tests__/action-destination.test.ts
+++ b/packages/browser/src/plugins/remote-loader/__tests__/action-destination.test.ts
@@ -1,16 +1,18 @@
 import unfetch from 'unfetch'
 import braze from '@segment/analytics-browser-actions-braze'
+import { CDNSettingsBuilder } from '@internal/test-helpers'
 
 import { AnalyticsBrowser } from '../../../browser'
 import { createSuccess } from '../../../test-helpers/factories'
 import { JSDOM } from 'jsdom'
 
+const writeKey = 'abc'
+
 jest.mock('unfetch')
 jest.mocked(unfetch).mockImplementation(() =>
-  createSuccess({
-    integrations: {},
-    remotePlugins: [
-      {
+  createSuccess(
+    new CDNSettingsBuilder({ writeKey })
+      .addActionDestinationSettings({
         name: 'Braze Web Mode (Actions)',
         creationName: 'Braze Web Mode (Actions)',
         libraryName: 'brazeDestination',
@@ -38,9 +40,9 @@ jest.mocked(unfetch).mockImplementation(() =>
             },
           ],
         },
-      },
-    ],
-  })
+      })
+      .build()
+  )
 )
 
 beforeEach(async () => {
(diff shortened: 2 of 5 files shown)


packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -101,8 +115,24 @@ export class ActionDestination implements DestinationPlugin {
...
+      ])
+
+      return await this.action.load(ctx, analytics)
+    } catch (error) {
+      ctx.stats.increment('analytics_js.action_plugin.invoke.error', 1, [
+        `method:load`,
+        `action_plugin_name:${this.action.name}`,
+      ])
+
+      console.log(error)
```
oscb, line 132: leftover console.log

oscb (changes requested):
just a very minor comment, every thing else looks good!

packages/browser/src/plugins/remote-loader/__tests__/action-destination.test.ts
```diff
@@ -0,0 +1,111 @@
...
+import braze from '@segment/analytics-browser-actions-braze'
+
+import { AnalyticsBrowser } from '../../../browser'
+import { createSuccess } from '../../../test-helpers/factories'
+import { JSDOM } from 'jsdom'
+
+jest.mock('unfetch')
+jest.mocked(unfetch).mockImplementation(() =>
+  createSuccess({
+    integrations: {},
```
silesky, line 11: Nit: Seems like a good place to use the CDNSettingsBuilder helper class!

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -79,7 +79,21 @@ export class ActionDestination implements DestinationPlugin {
         transformedContext = await this.transform(ctx)
       }
 
-      await this.action[methodName]!(transformedContext)
+      try {
+        ctx.stats.increment('analytics_js.action_plugin.invoke', 1, [
```
silesky, line 83: This will let us easily aggregate these on datadog??? Sweet

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -79,7 +79,21 @@ export class ActionDestination implements DestinationPlugin {
...
+          `method:${methodName}`,
+          `action_plugin_name:${this.action.name}`,
+        ])
+
+        await this.action[methodName]!(transformedContext)
+      } catch (error) {
+        ctx.stats.increment('analytics_js.action_plugin.invoke.error', 1, [
```
silesky, line 90: We are repeating this string a few times. Can we save bytes through a helper fn?

packages/browser/src/plugins/remote-loader/__tests__/action-destination.test.ts
zikaari: updated

packages/browser/src/plugins/remote-loader/index.ts
zikaari: gzip is extremely good at handling repeating strings, I've verified it - sometimes placing the strings inline yields a smaller size than storing it in a variable.

silesky: Even so, I still think DRY would be an improvement here for the usual reasons?

oscb (approved):

packages/browser/src/plugins/remote-loader/index.ts
zikaari: It's like this all across the codebase, we should refactor them all at once so that the declaration and usage is standardized all across, and not fragment it even further.

silesky (approved):

packages/browser/src/plugins/remote-loader/index.ts
silesky: Fair

.changeset/metal-drinks-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@segment/analytics-next': minor
```
silesky, line 2: Should this be a patch, since it's not a user-facing feature?

zikaari: I saw it more as a "non-breaking enhancement" than a bug fix - so went with it

chrisradek: Would vote for patch since it isn't a user-facing change to any API/API behavior.

packages/browser/src/plugins/remote-loader/index.ts
chrisradek: I see the benefit of storing in a variable more that we have 1 place to update/less chance of a typo/forgetting to update the value if we change it in the future. Won't block, but that would be my general guidance when we've got a string value that acts as a key in downstream.

PR #983, opened 2023-10-27
rename 'shouldLoad' -> 'shouldLoadSegment'; add shouldLoadWrapper API

silesky:
See changeset for full list of changes.

Removes disableConsentRequirement setting in favor of using shouldLoad's ctx.abort (the settings API was getting a bit hard to understand with some redundant ways of doing things).

...Mainly this adds shouldLoadWrapper API , which is a bit like a perform block in action destinations. It allows people to wait for their globals to initialize before any other settings that might use those globals have time to create issues.

### DIFF
--- a/.changeset/fair-pillows-sin.md
+++ b/.changeset/fair-pillows-sin.md
@@ -2,4 +2,6 @@
 '@segment/analytics-consent-tools': major
 ---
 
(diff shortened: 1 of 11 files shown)


chrisradek (approved):
Looks good to me - appreciate that there are no breaking changes to the one trust wrapper!

PR #985, opened 2023-10-27
Update integration metrics capturing

zikaari:
Previous attempt at capturing action metrics used new metrics key, however that requires certain updates on the production server to work. The schedule of when those updates will go live is uncertain, so we're back to using the existing framework without minor changes.

### DIFF
--- a/packages/browser/src/plugins/ajs-destination/index.ts
+++ b/packages/browser/src/plugins/ajs-destination/index.ts
@@ -29,6 +29,7 @@ import {
   isDisabledIntegration as shouldSkipIntegration,
   isInstallableIntegration,
 } from './utils'
+import { recordIntegrationMetric } from '../../core/stats/metric-helpers'
 
 export type ClassType<T> = new (...args: unknown[]) => T
 
@@ -158,10 +159,19 @@ export class LegacyDestination implements DestinationPlugin {
     })
 
     try {
-      this.recordMetric(ctx, 'initialize', true)
+      recordIntegrationMetric(ctx, {
+        integrationName: this.name,
+        methodName: 'initialize',
+        type: 'classic',
+      })
       this.integration.initialize()
     } catch (error) {
-      this.recordMetric(ctx, 'initialize', true)
+      recordIntegrationMetric(ctx, {
+        integrationName: this.name,
+        methodName: 'initialize',
+        type: 'classic',
+        didError: true,
+      })
       throw error
     }
   }
@@ -179,14 +189,6 @@ export class LegacyDestination implements DestinationPlugin {
       // page events can't be buffered because of destinations that automatically add page views
       ctx.event.type !== 'page' &&
       (isOffline() || this._ready === false || this._initialized === false)
-    )
-  }
-
-  private recordMetric(ctx: Context, methodName: string, errored = false) {
-    ctx.stats.increment(
-      `analytics_js.integration.invoke${errored ? '.error' : ''}`,
-      1,
-      [`method:${methodName}`, `integration_name:${this.name}`, `type:classic`]
     )
   }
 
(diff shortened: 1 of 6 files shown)


packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -117,25 +109,29 @@ export class ActionDestination implements DestinationPlugin {
...
   }
+
+  private recordMetric(ctx: Context, methodName: string, errored = false) {
+    ctx.stats.increment(
+      `analytics_js.integration.invoke${errored ? '.error' : ''}`,
+      1,
+      [
+        `method:${methodName}`,
+        `integration_name:${this.action.name}`,
+        `type:${this.action.type}`,
```
chrisradek, line 131: Shouldn't think be something like `action-destination`? Otherwise it's the lifecycle hook type.

zikaari: The type could be `destination`, `before`, `enrichment`, etc. so for the cases where it is `destination` it'd be semantically and technically accurate for all purposes concerned I think.

As seen here: https://github.com/segmentio/analytics-next/pull/985/files#diff-54c61325aeaea6d671258e16b23487255b72fa13550ae7175748fb5ddcc5af61R75

chrisradek: With your previous PR, we were only tracking the fact that it is an action destination, and not the plugin type (`before`, `enrichment`, `destination`, `after`). So I thought this change would be doing the same - since you have `classic-destination`, that this would be `action-destination`. We could also add a tag for the plugin type, though I'm not sure if that's necessary since we can easily infer it from the plugin name)

zikaari: The previous PR used a more generic terminology, i.e `action_plugin` which didn't necessarily mean a destination - but anything "action-like".

Since in this case we're re-using existing infrastructure, and it uses the term `integration`, I felt it might be important to be explicit about what exactly it is. If we were only to have 2 broad categories, `classic-destination` and `action-destination`, it'd make it very hard to construct datadog queries, because then all records queried by `action-destination` might also include things like `Braze Web Mode - debounce`, which isn't a destination at all.

By isolating things like `Braze web mode -debounce` into their own 'scope', any queries we run won't be tainted.

chrisradek: >  If we were only to have 2 broad categories, classic-destination and action-destination, it'd make it very hard to construct datadog queries

I don't think I see the difference - we only had 2 categories before, now we have 5 with these changes. classic-destinations are also a `destination` type for what it's worth. If you're worried about action-destination including non-destination plugins, we could change valid type values to `classic`, `action`, and leave room for future things?

My main concern is that this makes it harder to extend in the future. We're reserving all the plugin types to implicitly mean we're looking at an action, which can be confusing if we ever start tracking things like source middleware failures (which would fall under `before` plugin type), or add something in the future to replace actions.

zikaari: So to be sure, you're proposing that the type be either `classic-destination` or `action-destination`, and not read `this.action.type`?

If the answer to above is yes, then should we disable metric logging for when `this.action.type !== 'destination'`?

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -267,6 +268,7 @@ export class LegacyDestination implements DestinationPlugin {
       ctx.stats.increment('analytics_js.integration.invoke.error', 1, [
         `method:${eventType}`,
         `integration_name:${this.name}`,
+        `type:classic-destination`,
```
silesky, line 271: We should just add "recordMetric" to the stats class or have it as a util that we can use here?

packages/browser/src/plugins/remote-loader/index.ts
chrisradek: I'm proposing that the type is `classic`/`classic-destination` or `action`/`action-destination` (I don't particularly care if they include `destination` in the name - it's just the type of integration), and if there's value in knowing the plugin type, we add a different tag for that (maybe `pluginType`).

packages/browser/src/plugins/ajs-destination/index.ts
zikaari: I thought of that too, but I'm not 100% sure of where to put it. One side of me saw Stats class as agnostic of its usage.

For now - we can leave it as-is, since we need to merge this soon.

silesky: just put it in a utility function?

chrisradek (approved):

Issue #972, opened 2023-10-16
Segment is not loaded when OneTrust consent integration is setup and jurisdiction does not require a banner to be shown. 


backjo:
In https://github.com/segmentio/analytics-next/blob/1eaef11131e038996860441a6f0863ce89663668/packages/consent/consent-wrapper-onetrust/src/domain/wrapper.ts#L48 , Segment's consent-wrapper for OneTrust checks for `oneTrustGlobal.IsAlertBoxClosed()`. This function will not return true unless a user has explicitly confirmed / set their privacy settings. While this is desirable in jurisdictions like US-CA and the EU, some jurisdictions do not require a banner to first be shown and thus many users will not explicitly set a privacy setting with OneTrust - preventing the Segment SDK from ever being loaded for that user. 
<img width="910" alt="Screenshot 2023-10-16 at 5 47 23 PM" src="https://github.com/segmentio/analytics-next/assets/2441427/fc756ea3-0860-4df1-a9a4-3c9f53d56e67">

silesky:
Thanks, @backjo for taking out an issue. 

Do you mind telling me how to find that "Show Banner" setting in OneTrust, so I can reproduce? I'm having trouble finding it in the OT UI.

silesky:
Thanks @backjo for reporting. I have been unable to identify a workaround and have contacted OneTrust about this behavior / bug.

backjo:
Thx @silesky - if it's a larger fix that is needed, it might be useful just to let a consumer of this library drive this behavior via a configuration property on the OneTrustSettings

silesky:
@backjo good news, I have a PR up (#987)

Just confirming: you have default categories (e.g CAT001, etc) set up that will always be active for the jurisdictions in question, correct, unless they explicitly opt out by interacting with the modal?

PR #987, opened 2023-11-02
consent: handle cases where show banner is unchecked

silesky:
#972 

Fix bug: if showAlertNotice is false and Segment has default categories, we want to immediately load Segment with any default categories. 

This PR mantains the status quo when it comes to end user re-consent / updated consent (consent updates made after the initial consent decision, in this case, any decision where the end user toggles the alert box on again.)
- stamped consent categories will update as usual
- analytics will not load any new device mode plugins according to  the updated consent 

^ This will come in later PRs / work

### DIFF
--- a/packages/consent/consent-wrapper-onetrust/src/lib/onetrust-api.ts
+++ b/packages/consent/consent-wrapper-onetrust/src/lib/onetrust-api.ts
@@ -21,13 +21,15 @@ type GroupInfoDto = {
 
 type OtConsentChangedEvent = CustomEvent<ConsentGroupIds>
 
+export interface OneTrustDomainData {
+  ShowAlertNotice: boolean
+  Groups: GroupInfoDto[]
+}
 /**
  * The data model used by the OneTrust lib
  */
 export interface OneTrustGlobal {
-  GetDomainData: () => {
-    Groups: GroupInfoDto[]
-  }
+  GetDomainData: () => OneTrustDomainData
   /**
    *  This callback appears to fire whenever the alert box is closed, no matter what.
    * E.g:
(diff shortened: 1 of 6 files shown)


packages/consent/consent-wrapper-onetrust/src/lib/onetrust-api.ts
```diff
@@ -50,10 +52,19 @@ export const getOneTrustGlobal = (): OneTrustGlobal | undefined => {
   ) {
     return oneTrust
   }
+  // if "show banner" is unchecked, window.OneTrust returns {geolocationResponse: {…}} before it actually returns the OneTrust object
+  if ('geolocationResponse' in oneTrust) {
+    return undefined
+  }
```
chrisradek, line 58: Sorry - I'm having a little trouble understanding this bit.

`window.OneTrust` can be different objects depending on some conditions? Does it become the OneTrust object we know only after the banner is interacted with, or after something outside our control is loaded?

The comment about polling a few lines down confused me a bit too...when does this happen?

silesky: > window.OneTrust can be different objects depending on some conditions?
Correct. if the banner is false, it will have a moment before it fully initialized where it just has the geolocation property. Manual testing was the only thing that uncovered that. Likely undocumented.

> The comment about polling a few lines down confused me a bit too...when does this happen?

See the usage of "resolveWhen" for when we poll for the OneTrust global

chrisradek: > possible that there is some intermediate state(s) we don't know about that a customer (or us) will discover

Yeah, that seems reasonable to log then.

chrisradek (approved):
:shipit:

PR #992, opened 2023-11-07
 Add disable API boolean

silesky:
Docs: https://github.com/segmentio/segment-docs/pull/5681

Add 'disable' boolean / function option to allow for disabling Segment in a testing environment, which will allow us to disable segment based on a combination of consent settings and user's cdnSettings


packages/browser/src/core/analytics/__tests__/null-analytics.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    const analytics = new NullAnalytics()
+    expect(analytics.options.disableClientPersistence).toBe(true)
+  })
+
+  it('integration: should not touch cookies or localStorage', async () => {
+    const analytics = new NullAnalytics()
+    await analytics.track('foo')
+    const storage = getAjsBrowserStorage()
+    expect(Object.values(storage).every((v) => !v)).toBe(true)
+  })
```
oscb, line 36: Good idea for the test!

oscb (approved):

PR #993, opened 2023-11-10
add emitter to generic utils

silesky:
- add Emitter to @segment/analytics-generic-utils 
- add memory leak warning of 10 for a particular event (inspired by node, which has a global, rather than per-event limit of 10 -- after which a warning is logged). Unlike node, It's a warning only, atm.


packages/generic-utils/src/emitter/__tests__/emitter.test.ts
```diff
@@ -71,4 +71,36 @@ describe(Emitter, () => {
...
+
+  it('has no warning if listener limit is set to 0', () => {
+    const fn = jest.fn()
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+    const em = new Emitter({ maxListeners: 0 })
+    expect(em.maxListeners).toBe(0)
+    em.on('test', fn)
+    expect(warnSpy).not.toHaveBeenCalled()
+    em.on('test', fn)
+    expect(warnSpy).not.toHaveBeenCalled()
```
oscb, line 104: nit: would it be worth adding as many listeners as the default (10) to guarantee that's not using the default as `maxListeners`?

oscb (approved):
single nit comment on the test, non-blocker, looks good!

oscb (approved):
single nit comment on the test, non-blocker, looks good!

packages/generic-utils/src/emitter/__tests__/emitter.test.ts
silesky: Sure!

PR #1001, opened 2023-11-17
Consent - Do not load segment/send events unneccessarily

silesky:
See changeset.

This will **not** cause any problems for  recent older versions of analytics that don't have "disabled" -- they just won't get this feature.

### DIFF
--- a/packages/consent/consent-tools/package.json
+++ b/packages/consent/consent-tools/package.json
@@ -13,7 +13,7 @@
     "!**/test-helpers/**"
   ],
   "scripts": {
-    ".": "yarn run -T turbo run --filter=@segment/analytics-consent-tools",
+    ".": "yarn run -T turbo run --filter=@segment/analytics-consent-tools...",
     "test": "yarn jest",
     "lint": "yarn concurrently 'yarn:eslint .' 'yarn:tsc --noEmit'",
     "build": "rm -rf dist && yarn concurrently 'yarn:build:*'",
--- a/packages/consent/consent-tools/src/domain/__tests__/disable-segment.test.ts
+++ b/packages/consent/consent-tools/src/domain/__tests__/disable-segment.test.ts
@@ -2,7 +2,7 @@ import { CDNSettingsConsent } from '../../types'
 import { segmentShouldBeDisabled } from '../disable-segment'
 
 describe('segmentShouldBeDisabled', () => {
-  it('should be disabled if user has only consented to irrelevant vategories', () => {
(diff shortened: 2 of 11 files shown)


chrisradek:
> This will cause any problems

Will __not__?

packages/consent/consent-tools/package.json
```diff
@@ -13,7 +13,7 @@
     "!**/test-helpers/**"
   ],
   "scripts": {
-    ".": "yarn run -T turbo run --filter=@segment/analytics-consent-tools",
+    ".": "yarn run -T turbo run --filter=@segment/analytics-consent-tools...",
```
chrisradek, line 16: Just curious, when is this script run?

chrisradek (approved):
Looks good to me!

packages/consent/consent-tools/package.json
silesky: It's just a helper for local development. I think I wrote about in the docs under development.

silesky: Also forgot: we do use this in CI to lazily build specific packages

packages/consent/consent-tools/src/domain/create-wrapper.ts
```diff
@@ -180,3 +183,20 @@ const disableIntegrations = (
   )
   return results
 }
+
+const isDisabled = (
+  cdnSettings: CDNSettings,
+  initialCategories: Categories,
+  disableOpt: InitOptions['disable'] | undefined
```
oscb: Should we have this shortcircuit before everything? As in, if `disable` is `true` then bypass all the cdnsettings function

packages/consent/consent-tools/src/domain/__tests__/disable-segment.test.ts
```diff
@@ -0,0 +1,67 @@
+import { CDNSettingsConsent } from '../../types'
+import { segmentShouldBeDisabled } from '../disable-segment'
+
+describe('segmentShouldBeDisabled', () => {
+  it('should be disabled if user has only consented to irrelevant vategories', () => {
```
oscb, line 5: nit: typo `categories`

oscb (approved):
A couple minor comments, non blocking

packages/consent/consent-tools/src/domain/create-wrapper.ts
silesky: I actually was in the process of refactoring this when you wrote this comment =)

PR #1008, opened 2023-11-30
Change segmentio to 'Destination' type

danieljackins:
This PR changes the segment.io plugin to 'Destination' type, to prevent any issues with other plugins impacting event delivery to Segment

Tested via confirming sending events works as expected

### DIFF
--- a/packages/browser/src/browser/__tests__/integration.test.ts
+++ b/packages/browser/src/browser/__tests__/integration.test.ts
@@ -75,6 +75,16 @@ const googleAnalytics: Plugin = {
   type: 'destination',
 }
 
+const slowPlugin: Plugin = {
+  ...xt,
+  name: 'Slow Plugin',
+  type: 'destination',
+  track: async (ctx) => {
+    await sleep(3000)
+    return ctx
+  },
+}
+
 const enrichBilling: Plugin = {
   ...xt,
   name: 'Billing Enrichment',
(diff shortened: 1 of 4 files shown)


packages/browser/src/plugins/segmentio/index.ts
```diff
@@ -125,7 +125,7 @@ export function segmentio(
 
   const segmentio: Plugin = {
     name: 'Segment.io',
-    type: 'after',
+    type: 'destination',
```
silesky, line 128: Are there any tests we can write to test behavior / resilience of this change?

packages/browser/src/browser/__tests__/integration.test.ts
```diff
@@ -574,6 +584,34 @@ describe('Dispatch', () => {
...
+          total: 25,
+          userId: '👻',
+        },
+        {
+          integrations: {
+            All: true,
+          },
+        }
+      ),
+      sleep(100),
```
silesky, line 609: What does this sleep do? All promises in a race will continue running even after the first one crosses the finish line -- it's won't abort the promise FYI.

am I missing something -- can we remove this?

silesky, line 609: Never seen this way of setting a timeout, neat.

silesky (approved):

PR #1009, opened 2023-12-01
Capture page context as early as possible if `initialPageview` = true

silesky:
- Analytics: if initialPageView is true, we want to capture page context before destinations are registered.
- Consent: If initialPageView is true, we want to capture page context before .load() is called (taking advantage of the fact that page context is buffered when you do analytics.page())

### DIFF
--- a/packages/browser/src/browser/index.ts
+++ b/packages/browser/src/browser/index.ts
@@ -30,6 +30,7 @@ import {
   flushAddSourceMiddleware,
   flushSetAnonymousID,
   flushOn,
+  PreInitMethodCall,
 } from '../core/buffer'
 import { ClassicIntegrationSource } from '../plugins/ajs-destination/types'
 import { attachInspector } from '../core/inspector'
@@ -320,6 +321,11 @@ async function loadAnalytics(
   // this is an ugly side-effect, but it's for the benefits of the plugins that get their cdn via getCDN()
   if (settings.cdnURL) setGlobalCDNUrl(settings.cdnURL)
 
+  if (options.initialPageview) {
+    // capture the page context early, so it's always up-to-date
+    preInitBuffer.push(new PreInitMethodCall('page', []))
+  }
+
   let legacySettings =
     settings.cdnSettings ??
     (await loadLegacySettings(settings.writeKey, settings.cdnURL))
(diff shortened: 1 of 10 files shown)


packages/consent/consent-tools/src/types/wrapper.ts
```diff
@@ -13,6 +13,7 @@ export interface AnalyticsBrowserSettings {
 export interface InitOptions {
   updateCDNSettings?(cdnSettings: CDNSettings): CDNSettings
   disable?: boolean | ((cdnSettings: CDNSettings) => boolean)
+  initialPageview?: boolean
 }
```
oscb, line 17: why is this object completely detached from the interface of the options of the analytics client? shouldn't they be the same?

packages/consent/consent-tools/src/domain/create-wrapper.ts
```diff
@@ -43,6 +43,12 @@ export const createWrapper = <Analytics extends AnyAnalytics>(
       settings,
       options
     ): Promise<void> => {
+      // page() should happen as early as possible in order to prevent stale page context
```
oscb, line 46: Can you add the comment here that we need to do it for consent before we load settings/destintations hence why we bypass the default behavior

silesky: Sure!

packages/browser/src/browser/index.ts
```diff
@@ -336,6 +337,11 @@ async function loadAnalytics(
     }
   }
 
+  if (options.initialPageview) {
```
chrisradek, line 340: Wondering if this should be moved even higher - above the `loadLegacySettings` maybe? Basically before any async work to make it harder for the page context to be incorrect due to navigation.

silesky: Done!

packages/browser/src/browser/index.ts
```diff
@@ -320,6 +321,11 @@ async function loadAnalytics(
   // this is an ugly side-effect, but it's for the benefits of the plugins that get their cdn via getCDN()
   if (settings.cdnURL) setGlobalCDNUrl(settings.cdnURL)
 
+  if (options.initialPageview) {
+    // capture the page context early, so it's always up-to-date
+    preInitBuffer.push(new PreInitMethodCall('page', []))
```
chrisradek, line 326: This doesn't store all the page context like the snippet does...does it? My understanding is that this would queue a page call that will be invoked after some async work has been completed (which gives the address of the page time to change)

silesky: Yes, it stores the page context!

chrisradek: Ah right, forgot `preInitBuffer` isn't a regular array.

chrisradek (approved):

PR #1010, opened 2023-12-08
Node: add flush method

silesky:
### Example
```ts
analytics.track('foo')
await analytics.flush()

analytics.track('bar')
await analytics.flush()
```

`analytics.closeAndFlush({ timeout: 3000 })` simply calls  `analytics.flush({ close: true, timeout: 3000 })`.

#### Behavior
1. Require `awaiting analytics.flush()` before calling flush again. 
2. Ignore additional flush calls while a flush is pending, and log warning to console.

#### Recap of internal discussio
- All of the scenarios we've encountered involve awaiting for flush.
- Supporting overlapping flush calls is complex.
- This was easy to implement, and we don't want to make perfect the enemy of the good.

### DIFF
--- a/.changeset/great-ads-share.md
+++ b/.changeset/great-ads-share.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-node': minor
+---
+
+Add analytics.flush({ timeout: ..., close: ... }) method
--- a/packages/node/src/__tests__/graceful-shutdown-integration.test.ts
+++ b/packages/node/src/__tests__/graceful-shutdown-integration.test.ts
@@ -241,4 +241,142 @@ describe('Ability for users to exit without losing events', () => {
       expect(calls[0].data.batch.length).toBe(2)
     })
   })
+
+  describe('.flush()', () => {
+    beforeEach(() => {
+      ajs = new Analytics({
+        writeKey: 'abc123',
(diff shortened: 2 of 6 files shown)


chrisradek:
Just wanting to confirm I'm reading the code correctly:

With this change, am I understanding correctly that if you call `flush()` multiple times with `track()` calls interspersed, all `flush()` calls will wait to resolve until all the `track()` calls completed? This is assuming that `track()` calls were made before any of the `flush()` calls completed.

silesky:
@chrisradek 
> Just wanting to confirm I'm reading the code correctly:
> 
> With this change, am I understanding correctly that if you call `flush()` multiple times with `track()` calls interspersed, all `flush()` calls will wait to resolve until all the `track()` calls completed? This is assuming that `track()` calls were made before any of the `flush()` calls completed.

Not quite. If any events enter the queue after .flush is called() (not necessarily awaited), those events are not flushed, but rather, included with the next batch.

 I think the best way to illustrate the behavior is with this test: I added some additional code comments. 
https://github.com/segmentio/analytics-next/pull/1010/files#diff-a58dd0e72b28d53cee2b402d434904f47d91d318d2c4816b5331d6a67f2bbeafR283

chrisradek (approved):

PR #1015, opened 2023-12-14
Node: deprecate `maxEventsInBatch` in favor of `flushAt`

silesky:
Deprecate `maxEventsInBatch` in favor of `flushAt`. The purpose is to establish consistency between our SDKs, regardless of language.

(Basically, let's go back to the original naming of maxEventsInBatch)

Docs Change: https://github.com/segmentio/segment-docs/pull/5808

### DIFF
--- a/.changeset/fifty-trains-act.md
+++ b/.changeset/fifty-trains-act.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-node': major
+---
+
+Deprecate `maxEventsInBatch` in favor of our commonly used: `flushAt`. The purpose is to establish consistency between our SDKs, regardless of language.
--- a/.changeset/fifty-trains-act.md
+++ b/.changeset/fifty-trains-act.md
@@ -1,5 +1,5 @@
 ---
-'@segment/analytics-node': major
+'@segment/analytics-node': minor
 ---
 
 Deprecate `maxEventsInBatch` in favor of our commonly used: `flushAt`. The purpose is to establish consistency between our SDKs, regardless of language.
--- a/packages/node/src/app/analytics-node.ts
+++ b/packages/node/src/app/analytics-node.ts
@@ -48,7 +48,7 @@ export class Analytics extends NodeEmitter implements CoreAnalytics {
         host: settings.host,
         path: settings.path,
         maxRetries: settings.maxRetries ?? 3,
-        maxEventsInBatch: settings.maxEventsInBatch ?? 15,
+        flushAt: settings.flushAt ?? settings.maxEventsInBatch ?? 15,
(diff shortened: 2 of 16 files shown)


packages/node/src/app/analytics-node.ts
```diff
@@ -33,25 +33,26 @@ export class Analytics extends NodeEmitter implements CoreAnalytics {
 
   constructor(settings: AnalyticsSettings) {
     super()
+    settings.flushAt ??= settings.maxEventsInBatch ??= 15
+    settings.maxRetries ??= 3
+    settings.flushInterval ??= 10000
```
chrisradek, line 38: Do these `??=` get transpiled? Basically just want to make sure this doesn't cause any problems with the minimum node.js versions we support.

packages/node/src/app/analytics-node.ts
```diff
@@ -33,25 +33,26 @@ export class Analytics extends NodeEmitter implements CoreAnalytics {
 
   constructor(settings: AnalyticsSettings) {
     super()
+    settings.flushAt ??= settings.maxEventsInBatch ??= 15
```
chrisradek, line 36: Wondering if we should create a copy of these settings - typically we don't want to mutate an object the user passes in.

.changeset/fifty-trains-act.md
```diff
@@ -0,0 +1,5 @@
+---
+'@segment/analytics-node': major
```
chrisradek, line 2: Might want to coordinate with @MichaelGHSeg to add the breaking changes he needs to make for oauth so they both get included when we bump this (if you haven't coordinated already)

chrisradek (commented):
Overall looks good, just a few comments.

packages/node/src/app/analytics-node.ts
silesky: Yes they get transpired. We have tests that target our minimum node version, so not worried.

chrisradek: Sweet, thanks!

packages/node/src/app/analytics-node.ts
silesky: I totally agree with this.

.changeset/fifty-trains-act.md
silesky: What's the latest on that? We can make this a minor version bump since this is just a "deprecation", though it's pretty common to have deprecation as part of a major bump (and then removed in the next major bump).

Thoughts? I'd rather just get this in as a minor if it will be a while.

chrisradek (approved):

PR #1027, opened 2024-01-22
message id generation: hash only the event and not the uuid

andrewloyola:

### DIFF
--- a/packages/browser/src/core/events/index.ts
+++ b/packages/browser/src/core/events/index.ts
@@ -8,7 +8,6 @@ import {
   Traits,
   SegmentEvent,
 } from './interfaces'
-import md5 from 'spark-md5'
 import { addPageContext, PageContext } from '../page'
 
 export * from './interfaces'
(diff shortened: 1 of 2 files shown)


packages/browser/src/core/events/index.ts
```diff
@@ -261,7 +261,7 @@ export class EventFactory {
       context,
       integrations: allIntegrations,
       ...overrides,
-      messageId: 'ajs-next-' + md5.hash(JSON.stringify(event) + uuid()),
+      messageId: 'ajs-next-' + md5.hash(JSON.stringify(event)) + uuid(),
```
silesky, line 264: Maybe we just rip the band-aid off and do what we do in node.js?

https://github.com/segmentio/analytics-next/blob/master/packages/node/src/lib/get-message-id.ts

andrewloyola: updated the pr to reflect this change

silesky:
@andrewloyola thanks for your contribution!
Couple things:
1. Can we yarn uninstall the md5 dependency?
2. Can you run "yarn test" and fix any broken tests?

silesky:
Closing in favor of #1045

PR #1032, opened 2024-01-26
Support loading analytics into a custom global variable when using snippet version 5.2.1 or later

### LINKED ISSUE #1014: How to use `globalAnalyticsKey` in the browser, using Vanilla JS?
Hi team,

A customer doesn't want to use `window.analytics` due to a clash with another analytics.js library that it is not ours. They'd like to rename the global object to something like `window.segment_analytics`. 

I found in the docs [a property for it](https://github.com/segmentio/analytics-next/blob/a8b2a238aeeaecf051046765dc1b3aa5c90225b1/packages/browser/src/lib/global-analytics-helper.ts#L21): 

```ts
/**
 * Stores the global window analytics key
 */
let _globalAnalyticsKey = 'analytics'
```

However, there are no examples on how to use it, so using a little bit of debugging + reading the original code, I crafted this PoC:

```html
<html>
  <head>
    <script>
      !(function () {
        var segment_analytics = (window.segment_analytics = window.segment_analytics || []);
        if (!segment_analytics.initialize)
          if (segment_analytics.invoked)
            window.console &&
              console.error &&
              console.error("Segment snippet included twice.");
          else {
            segment_analytics.invoked = !0;
            segment_analytics.methods = [
              "trackSubmit",
              "trackClick",
              "trackLink",
              "trackForm",
              "pageview",
              "identify",
              "reset",
              "group",
              "track",
              "ready",
              "alias",
              "debug",
              "page",
              "once",
              "off",
              "on",
              "addSourceMiddleware",
              "addIntegrationMiddleware",
              "setAnonymousId",
              "addDestinationMiddleware",
            ];
            segment_analytics.factory = function (e) {
              return function () {
                var t = Array.prototype.slice.call(arguments);
                t.unshift(e);
                segment_analytics.push(t);
                return segment_analytics;
              };
            };
            for (var e = 0; e < segment_analytics.methods.length; e++) {
              var key = segment_analytics.methods[e];
              segment_analytics[key] = segment_analytics.factory(key);
            }
            segment_analytics.load = function (key, e) {
              var t = document.createElement("script");
              t.type = "text/javascript";
              t.async = !0;
              t.src = "https://cdn.segment.com/analytics.js/v1/" + key + "/analytics.min.js";
              var n = document.getElementsByTagName("script")[0];
              n.parentNode.insertBefore(t, n);
              segment_analytics._loadOptions = {...e, globalAnalyticsKey: 'segment_analytics' };
            };
            segment_analytics._writeKey = "my-write-key";
            segment_analytics.SNIPPET_VERSION = "4.15.3";
            segment_analytics.load("my-write-key");
          }
      })();
    </script>
    <title>Home</title>
  </head>
  <body>
    <h1>PoC for Global Analytics Key</h1>
  </body>
</html>
```

This runs with no errors. However, when I send events using it, 1) the events are not sent to Segment, and 2) curiously, the events are added to `segment_analytics` as an array:

image

Am I setting this correctly? Is there another way of setting it? 

Thanks in advance!

zikaari:
Fixes #1014

Please refer to the original issue for more information

Docs for this feature - https://github.com/segmentio/segment-docs/pull/5949

### DIFF
--- a/.changeset/kind-crabs-report.md
+++ b/.changeset/kind-crabs-report.md
@@ -1,6 +1,5 @@
 ---
-'@segment/analytics-next': patch
-'@internal/browser-integration-tests': patch
+'@segment/analytics-next': minor
 ---
 
 Support loading analytics into a custom global variable when using snippet version 5.2.1 or later
--- a/.changeset/kind-crabs-report.md
+++ b/.changeset/kind-crabs-report.md
@@ -1,0 +1,6 @@
+---
+'@segment/analytics-next': patch
+'@internal/browser-integration-tests': patch
+---
+
+Fix globalAnalyticsKey feature
--- a/.changeset/kind-crabs-report.md
+++ b/.changeset/kind-crabs-report.md
@@ -3,4 +3,4 @@
 '@internal/browser-integration-tests': patch
 ---
 
-Fix globalAnalyticsKey feature
+Support loading analytics into a custom global variable when using snippet version 5.2.1 or later
--- a/packages/browser-integration-tests/src/custom-global-key.test.ts
+++ b/packages/browser-integration-tests/src/custom-global-key.test.ts
@@ -1,0 +1,53 @@
+import { test, expect } from '@playwright/test'
+import { standaloneMock } from './helpers/standalone-mock'
+import { extractWriteKeyFromUrl } from './helpers/extract-writekey'
+import { CDNSettingsBuilder } from '@internal/test-helpers'
+
+test.describe('Segment with custom global key', () => {
+  test.beforeEach(standaloneMock)
+  test.beforeEach(async ({ context }) => {
+    await context.route(
+      'https://cdn.segment.com/v1/projects/*/settings',
+      (route, request) => {
+        if (request.method().toLowerCase() !== 'get') {
+          return route.continue()
+        }
+
+        const writeKey = extractWriteKeyFromUrl(request.url()) || 'writeKey'
+        return route.fulfill({
+          status: 200,
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(new CDNSettingsBuilder({ writeKey }).build()),
+        })
+      }
+    )
+  })
+
+  test('supports using a custom global key', async ({ page }) => {
+    // Load analytics.js
+    await page.goto('/standalone-custom-key.html')
+    await page.evaluate(() =>
+      (window as any).segment_analytics.load('fake-key')
+    )
+
+    const req = await page.waitForRequest('https://api.segment.io/v1/t')
+
+    // confirm that any events triggered from within snippet have been sent
+    expect(req.postDataJSON().event).toBe('track from snippet')
+
+    const contextObj = await page.evaluate(() =>
+      (window as any).segment_analytics.track('track after load')
+    )
+
+    // confirm that any events triggered after load return a regular context object
+    expect(contextObj).toMatchObject(
+      expect.objectContaining({
+        attempts: expect.anything(),
+        event: expect.objectContaining({ event: 'track after load' }),
+        stats: expect.objectContaining({ metrics: expect.anything() }),
+      })
+    )
+  })
+})
--- a/packages/browser-integration-tests/src/custom-global-key.test.ts
+++ b/packages/browser-integration-tests/src/custom-global-key.test.ts
@@ -28,14 +28,15 @@ test.describe('Segment with custom global key', () => {
   test('supports using a custom global key', async ({ page }) => {
     // Load analytics.js
     await page.goto('/standalone-custom-key.html')
-    await page.evaluate(() =>
-      (window as any).segment_analytics.load('fake-key')
-    )
+    await page.evaluate(() => {
+      ;(window as any).segment_analytics.track('track before load')
+      ;(window as any).segment_analytics.load('fake-key')
+    })
 
     const req = await page.waitForRequest('https://api.segment.io/v1/t')
 
-    // confirm that any events triggered from within snippet have been sent
-    expect(req.postDataJSON().event).toBe('track from snippet')
+    // confirm that any events triggered before load have been sent
+    expect(req.postDataJSON().event).toBe('track before load')
 
     const contextObj = await page.evaluate(() =>
       (window as any).segment_analytics.track('track after load')
--- a/packages/browser-integration-tests/standalone-custom-key.html
+++ b/packages/browser-integration-tests/standalone-custom-key.html
@@ -86,9 +86,6 @@
               analytics._writeKey = key
             }
             analytics.SNIPPET_VERSION = '5.2.0'
-
-            // will be used for asserting in the corresponding test
-            analytics.track('track from snippet')
           }
       })()
     </script>
--- a/packages/browser-integration-tests/standalone-custom-key.html
+++ b/packages/browser-integration-tests/standalone-custom-key.html
@@ -1,0 +1,97 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script>
+      !(function () {
+        var i = 'segment_analytics',
+          analytics = (window[i] = window[i] || [])
+        if (!analytics.initialize)
+          if (analytics.invoked)
+            window.console &&
+              console.error &&
+              console.error('Segment snippet included twice.')
+          else {
+            analytics.invoked = !0
+            analytics.methods = [
+              'trackSubmit',
+              'trackClick',
+              'trackLink',
+              'trackForm',
+              'pageview',
+              'identify',
+              'reset',
+              'group',
+              'track',
+              'ready',
+              'alias',
+              'debug',
+              'page',
+              'screen',
+              'once',
+              'off',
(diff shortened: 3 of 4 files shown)


packages/browser-integration-tests/standalone-custom-key.html
```diff
@@ -0,0 +1,97 @@
+<!DOCTYPE html>
```
silesky, line 1: Following the pattern of the other integration tests -- can we move this to standalone-playground?

packages/browser-integration-tests/standalone-custom-key.html
```diff
@@ -0,0 +1,97 @@
...
+            }
+            analytics.load = function (key, n) {
+              var t = document.createElement('script')
+              t.type = 'text/javascript'
+              t.async = !0
+              t.setAttribute('data-global-segment-analytics-key', i)
+              t.src =
+                'https://cdn.segment.com/analytics.js/v1/' +
+                key +
+                '/analytics.min.js'
```
silesky, line 82: Is this testing the bundle that lives in this repo?

packages/browser-integration-tests/standalone-custom-key.html
```diff
@@ -0,0 +1,97 @@
...
+              var r = document.getElementsByTagName('script')[0]
+              r.parentNode.insertBefore(t, r)
+              analytics._loadOptions = n
+              analytics._writeKey = key
+            }
+            analytics.SNIPPET_VERSION = '5.2.0'
+
+            // will be used for asserting in the corresponding test
+            analytics.track('track from snippet')
```
silesky, line 91: Shouldn't we be testing segment_analytics.track()?

silesky:
@chrisradek should we link to our doc update here?

packages/browser/src/browser/standalone.ts
```diff
@@ -71,6 +72,16 @@ async function attempt<T>(promise: () => Promise<T>) {
...
 }
 
+const globalAnalyticsKey = (
+  document.querySelector(
+    'script[data-global-segment-analytics-key]'
+  ) as HTMLScriptElement
+)?.dataset.globalSegmentAnalyticsKey
+
+if (globalAnalyticsKey) {
+  setGlobalAnalyticsKey(globalAnalyticsKey)
```
silesky, line 82: This was a good idea to live here, so it doesn't get included in the npm bundle.

packages/browser-integration-tests/standalone-custom-key.html
zikaari: This is as per how @chrisradek advised me to setup an integration test

packages/browser-integration-tests/standalone-custom-key.html
zikaari: Yes, after it's been built

packages/browser-integration-tests/standalone-custom-key.html
zikaari: analytics here is a reference to `window['segment_analytics']`

silesky: Right, I had some other questions, though?

chrisradek: Agree with what Seth is saying - this page should just be as vanilla a snippet as can be - we can make the actual track call in the test file.

Also, does the new snippet always set `window.analytics` even if you're using a custom global key? I think that would cause a problem for users reserving `window.analytics` for something else. Wouldn't tests using this continue to pass without your a.js changes because `window.analytics` still exists with this snippet?

chrisradek (changes requested):
I like using the browser integration tests since it means we can test that the actual track calls work on the new global - I had some questions I posted in the PR - is there a docs PR for this change as well? I think we'll have to be very clear on how the snippet has to be modified to work with the `globalAnalyticsKey` configuration - based on the changelog I wouldn't expect that I'd need a new snippet to make this work.

packages/browser-integration-tests/standalone-custom-key.html
zikaari: > this page should just be as vanilla a snippet as can be - we can make the actual track call in the test file.

During our meeting we talked about testing if buffered calls (initiated in the snippet, before the main bundle loads) are taken care of safely - making those calls in the test file won't assert what we're trying to assert

> does the new snippet always set window.analytics even if you're using a custom global key

No, it does not. Where did you notice a read or write to `window.analytics`?

zikaari:
My docs PR was 1:1 replica of [Seth's PR](https://github.com/segmentio/segment-docs/pull/5754). Where do I explain the usage of this feature?

packages/browser-integration-tests/src/custom-global-key.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        })
+      }
+    )
+  })
+
+  test('supports using a custom global key', async ({ page }) => {
+    // Load analytics.js
+    await page.goto('/standalone-custom-key.html')
+    await page.evaluate(() =>
+      (window as any).segment_analytics.load('fake-key')
```
chrisradek, line 32: Ah, so if a custom key is used, we don't actually have to set `globalAnalyticsKey`? That's nice!

packages/browser-integration-tests/standalone-custom-key.html
zikaari: Yes you guys are right, its not that we are testing if calls in snippet get dealt with - its the calls before load, and your proposal will cover that. Let me change that

zikaari: Done - https://github.com/segmentio/analytics-next/pull/1032/commits/fc0be16dd7d3b73ede331b75485f4c1f2327a619

chrisradek:
> My docs PR was 1:1 replica of https://github.com/segmentio/segment-docs/pull/5754. Where do I explain the usage of this feature?

Do we currently explain anywhere in the docs that you can change the variable used to install analytics? I originally thought you also had to use the `globalAnalyticsKey` with this feature - but looking at your test looks like that's probably not the case. So might make sense to add another section to [Segment Docs](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/) on using something other than `window.analytics` for snippet users.

packages/browser-integration-tests/src/custom-global-key.test.ts
zikaari: Yes, it serves no purpose for snippet users. Unless it serves a utility for NPM users, it is useless.

.changeset/kind-crabs-report.md
```diff
@@ -0,0 +1,6 @@
+---
+'@segment/analytics-next': patch
+'@internal/browser-integration-tests': patch
+---
+
+Fix globalAnalyticsKey feature
```
chrisradek, line 6:
```suggestion
Support loading analytics into a custom global variable when using snippet version [X.X.X]+
```

zikaari:
Will add a section to the javascript, what do you think should the section be titled? "Installing Analytics under custom namespace"?

.changeset/kind-crabs-report.md
zikaari: Done - https://github.com/segmentio/analytics-next/pull/1032/commits/631923d69a3c66b3ffc8d81798f0a63d3167e071

chrisradek:
> "Installing Analytics under custom namespace"

I think something like that is fine. We can always update it as well, just want to make sure we are documenting our changes - we've not always been great about that historically.

packages/browser-integration-tests/standalone-custom-key.html
silesky: @zikaari I am still a bit confused -- to rephrase:

If this is loading analytics from the CDN, wouldn't this be currently testing against the version of analytics that doesn't have the changes reflected in this PR (i.e `if (globalAnalyticsKey)`)?

What am I missing?

zikaari: This does the redirection - https://github.com/segmentio/analytics-next/blob/d4f697502bcd38ec7a74c8545e354165bf41bec0/packages/browser-integration-tests/src/helpers/standalone-mock.ts#L18-L33

silesky: ohhhh... I missed that detail! Thanks, that's excellent!

zikaari:
Docs are here - https://github.com/segmentio/segment-docs/pull/5949

silesky (approved):

.changeset/kind-crabs-report.md
```diff
@@ -0,0 +1,6 @@
+---
+'@segment/analytics-next': patch
+'@internal/browser-integration-tests': patch
```
silesky, line 3: This should not actually be bumped

chrisradek (approved):

.changeset/kind-crabs-report.md
```diff
@@ -0,0 +1,6 @@
+---
+'@segment/analytics-next': patch
```
silesky, line 2: I would change this to minor, as this is a new feature IMO

PR #1036, opened 2024-02-08
Fix schema filter name collision bug

danieljackins:
This PR fixes an issue with integrations containing similar names (ie. Google Analytics and Google Analytics 4) having their schema filters applied incorrectly. It updates the filters to compare the schema filter name (which uses the creation name of the integration) with the creationName field in the remote plugin.

### DIFF
--- a/packages/browser/src/plugins/schema-filter/__tests__/index.test.ts
+++ b/packages/browser/src/plugins/schema-filter/__tests__/index.test.ts
@@ -10,6 +10,8 @@ const settings: LegacySettings = {
     'Braze Web Mode (Actions)': {},
     // note that Fullstory's name here doesn't contain 'Actions'
     Fullstory: {},
+    'Google Analytics': {},
+    'Google Analytics 4 Web': {},
     'Segment.io': {},
   },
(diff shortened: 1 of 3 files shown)


silesky:
@danieljackins ugh, I see a weird BK error.  Any idea what's going on?

danieljackins:
> @danieljackins looks good but ugh, I see a weird BK error that seems unrelated to this PR. Any idea what's going on?

I think buildkite is/was having some issues, I saw some others with the same problem

packages/browser/src/plugins/schema-filter/__tests__/index.test.ts
```diff
@@ -472,7 +472,7 @@ describe('schema filter', () => {
       expect(updateUserProfile.track).toHaveBeenCalled()
     })
 
-    it('covers different names between remote plugins and integrations', async () => {
+    it('works when current name differs from creation name', async () => {
```
chrisradek, line 475: Can we add a test that would have failed without this fix? 2 integrations where the 2nd one's name starts with the entirety of the 1st one?

PR #1045, opened 2024-03-01
Do not hash by message -- also, refactor event factory

silesky:
Finally, `EventFactory` is now shared between node and browser.

Features / Changes:
- Update messageId algorithm to be consistent with node (analytics-next-<epochdate>-uuid)
- Browser will throw an error in the EventFactory (not just in a plugin) if the event is invalid
- *alert* Validation now checks that one of userId/anonId/groupId/previousId is available and throws an error (to have consistency with the node validation). I assume that Browser will always have one of these values defined at this point (The user instance, etc) -- the only edge case would be if an instance had added identity resolution later.

### DIFF
--- a/packages/browser/src/core/events/__tests__/index.test.ts
+++ b/packages/browser/src/core/events/__tests__/index.test.ts
@@ -3,7 +3,6 @@ import { range, uniq } from 'lodash'
 import { EventFactory } from '..'
 import { getDefaultPageContext } from '../../page'
 import { User } from '../../user'
-import { SegmentEvent, Options } from '../interfaces'
 
 describe('Event Factory', () => {
   let user: User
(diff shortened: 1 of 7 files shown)


packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -59,6 +58,34 @@ describe('Event Factory', () => {
       const group = factory.group('coolKidsId', { coolkids: true })
       expect(group.groupId).toEqual('coolKidsId')
     })
+
+    it('allows userId / anonymousId to be specified as an option', () => {
```
danieljackins, line 62: copypasta?

packages/browser/src/core/events/__tests__/index.test.ts
```diff
@@ -59,6 +58,34 @@ describe('Event Factory', () => {
...
+    })
+
+    it('allows userId / anonymousId to be overridden', function () {
+      const group = factory.group('my_group_id', undefined, {
+        userId: 'bar',
+        anonymousId: 'foo',
+      })
+      expect(group.userId).toBe('bar')
+      expect(group.anonymousId).toBe('foo')
+    })
```
danieljackins, line 78: copypasta?

packages/core/src/events/index.ts
```diff
@@ -11,25 +11,54 @@ import {
...
+  /**
+   * additional validation to run on each event
+   */
+  validateEvent?: ValidateEventFn
 }
 
 /**
  * This is currently only used by node.js, but the original idea was to have something that could be shared between browser and node.
  * Unfortunately, there are some differences in the way the two environments handle events, so this is not currently shared.
  */
```
danieljackins, line 50: Is this comment still relevant?

packages/browser/src/core/events/__tests__/index.test.ts
silesky: Hah, good catch. I think this in master =S

packages/core/src/events/index.ts
silesky: Good catch again!

chrisradek (approved):

PR #1053, opened 2024-03-20
Support adding middleware to every device mode destination

silesky:
Add `*` glob support. 

```ts
addDestinationMiddleware('*', ({ next, payload }) => {
  if (dropThisEventForSomeReason) {
     return null
  }
  next(payload)
})
```

[x] I've included a changeset

### DIFF
--- a/packages/browser/src/browser/__tests__/integration.test.ts
+++ b/packages/browser/src/browser/__tests__/integration.test.ts
@@ -878,6 +878,116 @@ describe('addDestinationMiddleware', () => {
     })
   })
 
+  it('drops events if  next is never called', async () => {
+    const testPlugin: Plugin = {
+      name: 'test',
+      type: 'destination',
+      version: '0.1.0',
+      load: () => Promise.resolve(),
+      track: jest.fn(),
+      isLoaded: () => true,
+    }
+
+    const [analytics] = await AnalyticsBrowser.load({
+      writeKey,
+    })
+
+    const fullstory = new ActionDestination('fullstory', testPlugin)
+
+    await analytics.register(fullstory)
+    await fullstory.ready()
+    analytics.addDestinationMiddleware('fullstory', () => {
+      // do nothing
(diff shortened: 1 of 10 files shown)


packages/browser/src/browser/__tests__/integration.test.ts
```diff
@@ -878,6 +878,90 @@ describe('addDestinationMiddleware', () => {
     })
   })
 
+  it('drops events if  next is never called', async () => {
```
silesky, line 881: These are not new, but there didn't seem to be any explicit test coverage for these scenarios.

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -226,7 +226,6 @@ export class LegacyDestination implements DestinationPlugin {
             type: 'Dropped by plan',
           })
         )
-        return ctx
```
chrisradek, line 229: This is ok to remove because our event queue doesn't do anything with the returned ctx?

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -289,12 +294,7 @@ export async function remoteLoader(
               plugin
             )
 
-            /** Make sure we only apply destination filters to actions of the "destination" type to avoid causing issues for hybrid destinations */
-            if (
-              routing.length &&
-              routingMiddleware &&
-              plugin.type === 'destination'
```
chrisradek, line 296: Ah cool - so we were actually doing this check twice because `addMiddleware` also checks this

chrisradek (approved):

packages/browser/src/plugins/ajs-destination/index.ts
silesky: My editor actually autoremoved these because ctx.cancel throws an error, so this is dead code:

image

PR #1063, opened 2024-04-04
Add consent opt out

silesky:
Add support for opt-out consent.

OneTrust Consent Model | Behavior
-- | --
Opt-in | Opt-in
Opt-out | Opt-out
Implicit Consent | Opt-in
Default (any other model) | Opt-out

Note: to override the default behavior for OneTrust, the following setting(s) have been added:
```ts
withOneTrust(analytics, { consentModel: () => 'opt-out' | 'opt-in' })
```


packages/consent/consent-wrapper-onetrust/src/lib/onetrust-api.ts
```diff
@@ -21,9 +21,18 @@ type GroupInfoDto = {
 
 type OtConsentChangedEvent = CustomEvent<ConsentGroupIds>
 
+export enum OtConsentModel {
+  optIn = 'opt-in',
+  optOut = 'opt-out',
+  implicit = 'implied consent',
```
chrisradek, line 27: Is `implicit` needed now? Doesn't look like it's exposed in `OneTrustSettings` and is currently an alias for `optIn`

packages/consent/consent-wrapper-onetrust/src/domain/wrapper.ts
```diff
@@ -11,10 +11,18 @@ import {
...
 } from '../lib/onetrust-api'
-
+import { coerceConsentModel } from './consent-model'
 export interface OneTrustSettings {
   integrationCategoryMappings?: CreateWrapperSettings['integrationCategoryMappings']
   disableConsentChangedEvent?: boolean
+  /**
+   * Override configured consent model
+   * - opt-in  (default) - load segment and all destinations without waiting for explicit consent.
+   * - opt-out (strict/GDPR) - wait for explicit consent before loading segment
```
chrisradek, line 21: Aren't these flipped? `opt-in` is the one where we wait for you to opt-in before loading, and opt-out is where we load until you explicitly opt out?

packages/consent/consent-tools/src/types/settings.ts
```diff
@@ -56,9 +54,7 @@ export interface CreateWrapperSettings {
    * },
    * ```
    **/
-  shouldLoadSegment?: (
-    ctx: LoadContext
-  ) => Categories | void | Promise<Categories | void>
+  shouldLoadSegment?: (ctx: LoadContext) => void | Promise<void>
```
chrisradek, line 57: This looks like a breaking change - are we doing a major version bump (totally cool with it if we are!)

packages/consent/consent-tools/src/domain/validation/options-validators.ts
```diff
@@ -58,11 +58,7 @@ export function validateAnalyticsInstance(
   analytics: unknown
 ): asserts analytics is AnyAnalytics {
   assertIsObject(analytics, 'analytics')
-  if (
-    'load' in analytics &&
-    'on' in analytics &&
-    'addSourceMiddleware' in analytics
-  ) {
+  if ('load' in analytics && 'addSourceMiddleware' in analytics) {
```
chrisradek, line 61: Do we need to check for `addDestinationMiddleware` now too?

packages/consent/consent-tools/src/domain/blocking-helpers.ts
```diff
@@ -0,0 +1,118 @@
...
+export const filterDeviceModeDestinationsForOptIn = (
+  cdnSettings: CDNSettings,
+  consentedCategories: Categories,
+  filterSettings: DeviceModeFilterSettings
+): CDNSettings => {
+  const { remotePlugins, integrations } = cdnSettings
+  const { integrationCategoryMappings, shouldEnableIntegration } =
+    filterSettings
+
+  if (!remotePlugins) {
```
chrisradek, line 84: Why do we return early if `remotePlugins` doesn't exist? Is this only meant to run for action destinations and not classic destinations?

packages/consent/consent-tools/src/domain/blocking-middleware.ts
```diff
@@ -0,0 +1,85 @@
...
+      filterSettings,
+    })
+
+    if (disabled) return null
+
+    next(payload)
+  }
+  analyticsInstance.addDestinationMiddleware('*', blockDeviceMode)
+
+  // Block segment itsel (Segment.io isn't currently allowed in addDestinationMiddleware)
```
chrisradek, line 50: Nit:
```suggestion
  // Block segment itself (Segment.io isn't currently allowed in addDestinationMiddleware)
```

packages/consent/consent-tools/src/domain/blocking-helpers.ts
silesky: Ooh, this is a live bug in the codebase =) Luckily, none of our users _only_ use classic destinations, apparently.

packages/consent/consent-tools/src/types/settings.ts
silesky: Yes, was going to do a major bump!

packages/consent/consent-tools/src/domain/validation/options-validators.ts
silesky: Err, yeah -- this typeguard is incomplete =) Just here as kind of a sanity check.

packages/consent/consent-wrapper-onetrust/src/domain/wrapper.ts
silesky: Ooh, yep, the README is correct but these comments are flipped. Good catch.

packages/consent/consent-wrapper-onetrust/src/lib/onetrust-api.ts
silesky: Since we only have unique behavior for opt-in and opt-out, I felt like it would be good to leave it out of the public API, but document that it exists in the enum, since this enum is meant to be an Internal DTO that maps to the actual OT API. I don't feel super strongly either way!

packages/consent/consent-tools/src/domain/analytics/analytics-service.ts
```diff
@@ -87,6 +129,7 @@ export class AnalyticsService {
    * ```
    */
   consentChange(categories: Categories): void {
+    logger.debug('Consent change', categories)
```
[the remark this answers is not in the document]
silesky: Yep; this wrapper behavior is very tricky to debug so I added some magic debugging (only logs if a special global is set).

packages/consent/consent-tools/src/domain/blocking-helpers.ts
chrisradek: Ah - I just checked how we build the settings and it's always an array (empty if no action destinations are assigned) - so this check always passes.

silesky: Ah K, so this is actually a bug in the LegacySettings type

chrisradek: Yeah - this check can also just be removed since it never gets hit - but resolving since this doesn't have any impact.

chrisradek (approved):
Approved! Only one outstanding comment (missing closing HTML tag in a file)

PR #1067, opened 2024-04-12
fix: fixes buffering for legacy destinations

oscb:
# Fixes the Classic Destinations buffering to properly buffer events while not loaded

- Fixed `shouldBuffer` check as the starting state changed from `false` to `undefined` for ready/initialized. Hence events were bypassing buffer all the time.
- Changed the behaviour of `scheduleFlush` as previously it was being called after the timeout and ignoring if the plugin had finished initializing. The retry would work properly if the plugin throws an error but Bing/FB don't hence it was marking those events as sent.
  - TODO: This logic could be refactored in the future. Seems particularly convoluted and it should probably listen to `onReady` and `onInitialized` instead of retrying at intervals.

### DIFF
--- a/packages/browser/src/plugins/ajs-destination/index.ts
+++ b/packages/browser/src/plugins/ajs-destination/index.ts
@@ -192,7 +192,7 @@ export class LegacyDestination implements InternalPluginWithAddMiddleware {
     return (
       // page events can't be buffered because of destinations that automatically add page views
       ctx.event.type !== 'page' &&
-      (isOffline() || this._ready === false || this._initialized === false)
+      (isOffline() || this._ready !== true || this._initialized !== true)
     )
   }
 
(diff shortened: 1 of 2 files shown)


silesky (approved):

chrisradek (approved):

packages/browser/src/plugins/ajs-destination/index.ts
```diff
@@ -192,7 +192,7 @@ export class LegacyDestination implements InternalPluginWithAddMiddleware {
     return (
       // page events can't be buffered because of destinations that automatically add page views
       ctx.event.type !== 'page' &&
-      (isOffline() || this._ready === false || this._initialized === false)
+      (isOffline() || this._ready !== true || this._initialized !== true)
```
silesky, line 195: Nit: Couldn't we just get rid of the explicit false and do if (!this.ready) etc?

PR #1068, opened 2024-04-16
Fix RemotePlugins typescript type

silesky:
@chrisradek

### DIFF
--- a/.changeset/odd-pets-tickle.md
+++ b/.changeset/odd-pets-tickle.md
@@ -1,0 +1,5 @@
+---
+'@segment/analytics-next': patch
+---
+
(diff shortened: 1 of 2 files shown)


chrisradek (approved):

chrisradek:
Ah, I guess we could view this as a breaking change if someone were passing in their own CDN settings (seems unlikely outside of tests though)

silesky:
> Ah, I guess we could view this as a breaking change if someone were passing in their own CDN settings (seems unlikely outside of tests though)

Microsoft/Typescript philosophy? Types are never breaking changes :-)

PR #1080, opened 2024-04-29
Fix segment ga4 bug

silesky:
Addresses an issue where, if one of the non-destination actions fails to load/is blocked, the entire SDK fails to load. This is most notable in GA4, where, if GA was blocked, Segment initialization would fail.

### DIFF
--- a/packages/browser/src/plugins/schema-filter/index.ts
+++ b/packages/browser/src/plugins/schema-filter/index.ts
@@ -77,6 +77,7 @@ export function schemaFilter(
 
   return {
     name: 'Schema Filter',
+    critical: true,
     version: '0.1.0',
     isLoaded: () => true,
     load: () => Promise.resolve(),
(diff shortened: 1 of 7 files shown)


packages/browser/src/plugins/schema-filter/index.ts
```diff
@@ -77,6 +77,7 @@ export function schemaFilter(
 
   return {
     name: 'Schema Filter',
+    critical: true,
```
chrisradek, line 80: Ah, now that I see the load method for schema filter and env-enrichment, I'm not sure we actually need a `critical` field at this time. Their load methods are already safe.

It's the actual importing of the modules as part of `registerPlugins` that can still fail and block.

chrisradek (approved):
Approving, though I don't think the `critical` field is necessary after all. From a practical standpoint, it doesn't actually have any affect for the 2 plugins you added it to since their load methods are already safe.

PR #1084, opened 2024-05-06
AJS Retry improvements for 500 and 429, normal and batch

MichaelGHSeg:


packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -65,7 +68,7 @@ export default function batch(
       const { sentAt, ...newEvent } = event as SegmentEvent
       return newEvent
     })
-
+    console.log('sending batch', updatedBatch)
```
silesky, line 71: Nit: probably want to remove this. Rather than logging them here -- we want to limit console spam since it ends up muddying the customer's rollbar logs (these are logged in production). 

Being able to listen to this via the event emitter would probably be for the best.

packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -107,6 +138,7 @@ export default function batch(
     if (pageUnloaded && buffer.length) {
       const reqs = chunks(buffer).map(sendBatch)
       Promise.all(reqs).catch(console.error)
+      // can we handle a retry here?
```
silesky, line 141: Unfortunately not -- there are very serious restrictions around what you can do onPageChange. Sending an HTTP request to the server when the user closes the browser can be a bit tricky, as there's no guaranteed way to ensure the request will be completed successfully before the browser closes. This is one of the more brittle parts of the application.

danieljackins (dismissed):
Looks good! Needs a changeset

packages/browser/src/plugins/segmentio/normalize.ts
```diff
@@ -25,6 +27,16 @@ export function normalize(
     json._metadata = { failedInitializations: failed }
   }
 
+  if (ctx != null) {
+    const retryCount = analytics.queue.queue.getAttempts(ctx)
+    if (retryCount > 1) {
+      json._metadata = {
+        ...json._metadata,
```
silesky, line 34: I didn't think the main event queue retries are related to the retries in the client, so not sure that this would work? Also, this metadata field isn't super 

I would take this our, or write tests for this.

packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -88,28 +91,56 @@ export default function batch(
...
+  async function flush(attempt = 1): Promise<unknown> {
     if (buffer.length) {
       const batch = buffer
       buffer = []
-      return sendBatch(batch)
+      return sendBatch(batch)?.catch((error) => {
+        console.error('Error sending batch', error)
+        if (attempt < (config?.retryAttempts ?? 10)) {
+          if (error.name === 'RateLimitError') {
+            ratelimittimeout = error.retryTimeout
```
silesky, line 120: Camel case?

packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -88,28 +91,56 @@ export default function batch(
...
       }),
+    }).then((res) => {
+      if (res.status >= 500) {
+        throw new Error(`Bad response from server: ${res.status}`)
+      }
+      if (res.status === 429) {
+        const retryTimeoutStringSecs = res.headers?.get('x-ratelimit-reset')
+        const retryTimeoutMS =
+          retryTimeoutStringSecs != null
+            ? parseInt(retryTimeoutStringSecs) * 1000
```
silesky, line 102: Not: Instead of foo != null, more idiomatic in our code base to do !foo and not use loose equality operators.

packages/browser/src/plugins/segmentio/__tests__/retries.test.ts
```diff
@@ -8,11 +13,134 @@ import { scheduleFlush } from '../schedule-flush'
...
+
+    options = { apiKey: 'foo' }
+    analytics = new Analytics(
+      { writeKey: options.apiKey },
+      {
+        retryQueue: true,
+      }
+    )
+    segment = await segmentio(analytics, options, {})
+    await analytics.register(segment, envEnrichment)
```
silesky, line 38: Why register envEnrichment?

silesky (changes requested):
Very small comments but LGTM

packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -1,10 +1,12 @@
 import { SegmentEvent } from '../../core/events'
 import { fetch } from '../../lib/fetch'
 import { onPageChange } from '../../lib/on-page-change'
+import { RateLimitError } from './ratelimit-error'
 
 export type BatchingDispatchConfig = {
   size?: number
   timeout?: number
+  retryAttempts?: number
```
silesky, line 9: Can we be consistent with the node library and call this "maxRetries" instead?

packages/browser/src/plugins/segmentio/__tests__/retries.test.ts
```diff
@@ -8,11 +13,134 @@ import { scheduleFlush } from '../schedule-flush'
...
+    expect(ctx.attempts).toBe(1)
+    expect(fetch).toHaveBeenCalledTimes(1)
+
+    expect(fetch).toHaveBeenCalledTimes(1)
+    await new Promise((resolve) => setTimeout(resolve, 1000))
+    expect(fetch).toHaveBeenCalledTimes(2)
+    await new Promise((resolve) => setTimeout(resolve, 1000))
+    expect(fetch).toHaveBeenCalledTimes(3)
+    await new Promise((resolve) => setTimeout(resolve, 1000))
+    expect(fetch).toHaveBeenCalledTimes(3) // capped at 3 retries
```
silesky, line 140: If it's three retries, it should be 4 attempts, no? The initial attempt is not a retry?

packages/browser/src/plugins/segmentio/normalize.ts
MichaelGHSeg: pushWithBackoff calls updateAttempts, and I do have a check of getAttempts in the `retries on 500` test, but I think more explicit tests is a good idea.

MichaelGHSeg: Noticed that this doesn't work for batch.  I'll see if I can do that, but maybe ok without as well.

MichaelGHSeg: Figured out the queues, have the attempts incrementing independently, and added test for the output.

silesky (approved):

PR #1085, opened 2024-05-06
Node 429 rate limit timeout support

MichaelGHSeg:


silesky (approved):

packages/node/src/plugins/segmentio/publisher.ts
```diff
@@ -279,6 +280,20 @@ export class Publisher {
             new Error(`[${response.status}] ${response.statusText}`)
           )
           return
+        } else if (response.status === 429) {
```
silesky, line 283: This is a nice to have but I think it would be good to update the `http_request` event emitter to also give the current retryCount (and maybe the maxAttempts as well)!

PR #1096, opened 2024-06-24
Network signals

MichaelGHSeg:


packages/signals/browser-signals/src/core/signal-generators/network.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+const containsJSONContent = (headers: HeadersInit | undefined): boolean => {
+  if (!headers) {
+    return false
+  }
+  const normalizedHeaders = normalizeHeaders(headers)
+  return (
+    normalizedHeaders.get('content-type')?.includes('application/json') ||
+    normalizedHeaders.get('Content-Type')?.includes('application/json') ||
```
silesky, line 38: Headers.get is case insensitive, you don't need to check for content-type or Content-type -- just `content-type` is sufficient (maybe right a test to the affect, but I just tested) 😄

silesky (changes requested):

packages/signals/browser-signals/src/types/signals.ts
```diff
@@ -45,10 +49,16 @@ export type InstrumentationSignal = AppSignal<
   InstrumentationData
 >
 
+interface NetworkData {
```
silesky, line 52: Can you add the appropriate interface here 

e.g 

type NetworkRequestData = {
          action: 'Request',
          url:...
          method: 
          data: ...        
}


type NetworkResponseData = {
          action: 'Response',
          url:...
          method: 
          data: ...        
}

export type NetworkData = NetworkRequestData | NetworkResponseData

packages/signals/signals-example/src/components/ComplexForm.tsx
```diff
@@ -6,6 +6,25 @@ const ComplexForm = () => {
...
+    }
+    console.log('Submitting form:', JSON.stringify(formData))
+    fetch('/parrot', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(formData),
+    })
+      .then(async (response) => response.json())
```
silesky, line 21: no need for `async` here.

packages/signals/browser-signals/src/core/client/__tests__/redact.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  it('should redact string values in an object', () => {
+    const obj = { name: 'John Doe', age: '30' }
+    const expected = { name: 'XXXX XXX', age: '99' }
+    expect(redactJsonValues(obj)).toEqual(expected)
+  })
+
+  it('should redact string values in a nested object', () => {
+    const obj = { user: { name: 'Jane Doe', age: '25' }, active: true }
+    const expected = {
+      user: { name: 'XXXX XXX', age: '99' },
```
silesky, line 13: Per our prev. discussion, I think we should do that log 10 thing you mentioned ... I would avoid preserving string length or spaces like mobile -- worried that it deanonymizes -- just "XXX" for any string and "99" for any number is my preference.

for example:

city: XXX XXXX (New York), etc

packages/signals/browser-signals/src/core/signal-generators/network.ts
```diff
@@ -0,0 +1,98 @@
+import { SignalEmitter } from '../emitter'
+import { SignalGenerator } from './types'
+let origFetch: typeof window.fetch
+
+export function addFetchInterceptor(
+  onRequest: (rs: any) => void,
```
silesky, line 6: Can we type this with Request?

packages/signals/browser-signals/src/core/signal-generators/network.ts
```diff
@@ -0,0 +1,98 @@
...
+  onResponse: (rs: Response) => void
+) {
+  console.debug('Adding fetch interceptor')
```
silesky, line 9: Let's use logger.debug (it's in lib)

packages/signals/browser-signals/src/core/signal-generators/network.ts
```diff
@@ -0,0 +1,98 @@
...
+        data: {
+          action: 'Response',
+          url: url,
+          data: data,
+        },
+      })
+    }
+    addFetchInterceptor(handleRequest, handleResponse)
+    return () => {
+      console.debug('Removing fetch interceptor')
```
silesky, line 94: ditto logger.debug

packages/signals/browser-signals/src/core/signal-generators/network.ts
```diff
@@ -0,0 +1,97 @@
...
+          data: JSON.parse(rq.body.toString()),
+        },
+      })
+    }
+    const handleResponse = async (rs: Response) => {
+      if (!containsJSONContent(rs.headers)) {
+        return
+      }
+      const url = rs.url
+      if (!url || !new URL(url).hostname.includes(window.location.hostname)) {
```
silesky, line 77: I'm not sure this works in the following scenario:

I'm on url: `app.segment.com` and I want to dispatch an event to `segment.com/api/foo=123`

- In addition, can we add (more) tests for this domain matching stuff to ensure that it only fires when we expect?

packages/signals/browser-signals/src/types/signals.ts
```diff
@@ -45,10 +49,28 @@ export type InstrumentationSignal = AppSignal<
...
+  action: 'Request'
+  url: RequestInfo | URL
+  method: string
+  data: {}
+}
+
+type NetworkResponseData = {
+  action: 'Response'
+  url: string
+  data: {}
```
silesky, line 62: This type is an empty object, the best type would be Record<string, unknown> or {[key: string]: unknown}

packages/signals/browser-signals/src/types/signals.ts
```diff
@@ -45,10 +49,28 @@ export type InstrumentationSignal = AppSignal<
   InstrumentationData
 >
 
+type NetworkRequestData = {
+  action: 'Request'
+  url: RequestInfo | URL
```
silesky, line 54: Since this is going to be serialized both in the sandbox and when sent to the server, we should only have JSON types in our signal API. i.e this should be a string.

packages/signals/browser-signals/src/core/signal-generators/network.ts
MichaelGHSeg: Hmm, yeah, I think the right solution at that point would be a whitelist, which falls under the guiding principle of "Let the customer ask for it if they want it".  Separated out the logic 👍

PR #1100, opened 2024-07-02
Flush keepalive batches past keepalive limit

danieljackins:
This PR immediately sends any batches that are over the 64 KB limit when the keepalive option is turned on to help prevent any data loss.

### DIFF
--- a/packages/browser/src/plugins/segmentio/batched-dispatcher.ts
+++ b/packages/browser/src/plugins/segmentio/batched-dispatcher.ts
@@ -5,6 +5,7 @@ import { onPageChange } from '../../lib/on-page-change'
 export type BatchingDispatchConfig = {
   size?: number
   timeout?: number
+  keepalive?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


packages/browser/src/plugins/segmentio/batched-dispatcher.ts
```diff
@@ -23,6 +24,15 @@ function approachingTrackingAPILimit(buffer: unknown): boolean {
   return kilobytes(buffer) >= MAX_PAYLOAD_SIZE - 50
 }
 
+/**
+ * Checks if payload is over the limit for keepalive
+ * requests. If keepalive is enabled we want to avoid
+ * going over this to prevent data loss.
+ */
+function passedKeepaliveLimit(buffer: unknown): boolean {
+  return kilobytes(buffer) >= 64
```
silesky, line 33: Should we add padding, since in practice the max size is the total of all outbound requests with keep-alive, so we are potentially competing with other requests?

silesky (approved):

Issue #1120, opened 2024-07-22
How to write an enrichment plugin that fires after a remote integration?


kwalker3690:
I'm writing an 'enrichment' plugin that should add a property to any event that comes through `track`. However, the value of this property relies on a value added to events via an integration (the Amplitude Actions integration to be specific). Here is the code I'm using to register my plugin:

```
const testPlugin: Plugin = {
  name: "Session Replay Events",
  type: "enrichment",
  version: "1.0.0",

  isLoaded: () => true,
  load: loadMethod,
  track: trackMethod,
};

SegmentAnalytics.load({
  writeKey: <api key>,
  plugins: [testPlugin],
});
```

When debugging this, I've inspected the code and observed that `testPlugin` (red arrow in below screenshot) is always added before the Actions enrichment plugin (blue arrow in below screenshot)

![Screenshot 2024-07-22 at 2 20 39 PM](https://github.com/user-attachments/assets/25480190-9566-4eef-b5f0-24d9aa24f363)

I believe the ordering is occurring due to the ordering of `plugins` vs `remotePlugins` in [this code snippet](https://github.com/segmentio/analytics-next/blob/master/packages/browser/src/browser/index.ts#L285-L287).

I'd like to have my plugin be registered in load() so that we can enrich all events, including those emitted immediately on page load (as opposed to registering my plugin after load(), which means it will not capture those initial events).

Because of the ordering, when I console log via my `track` method, the `ctx.event.integrations` object is empty, as the Amplitude Actions integration has not yet been processed:
![image](https://github.com/user-attachments/assets/34a4dc94-d35a-4544-bb7c-cc4506d83964)

**Is there any way I can have my testPlugin be registered in `load()`, but have its track calls execute after those from remotely loaded plugins?**

silesky:
@kwalker3690 thanks for the level of detail. QQ: you try type: after instead of type: enrichment?

silesky:
@kwalker3690 That is odd behavior -- that for a given plugin, you would expect .load to _always_ resolve before any of its methods are called.

I think there's a pattern that works, however: can we do something like this?

```ts
  class MyPlugin implements Plugin {
    readonly type = "enrichment";
    readonly name = "Enrichment Plugin";
    readonly version = "0.0.0";
    public data!: Promise<string> 
    
    async load() {
      
      // fetch async data
      this.data = new Promise((resolve) => setTimeout(() => resolve('data'), 1000));
      
      return Promise.resolve();
    }
    
[...]

PR #1121, opened 2024-07-23
Fix plugin pipeline race condition

silesky:
Fixes a race condition in our plugin pipeline:

https://github.com/segmentio/analytics-next/issues/1120

There is currently inconsistencies between whether or not plugins are loaded in the Plugins[] array, or whether or not they are registered using .register. If they are loaded using .register, we don't wait wait for .load to finish resolving before invoking methods in that plugin. This is specifically an issue around plugins that are manually registered before analytics is initialized (i.e., most uses of analytics.register). The bug doesn't affect plugins loaded via action destinations, only plugins that are manually registered using analytics.register.

Bug repro:
Currently, this results in "page()!" being logged before  ".load()!".

```ts
 export const BugPlugin: Plugin = {
   type: "enrichment",
   name: "My Buggy Plugin",
   version: "1.0.0",
   load: async () => {
     await new Promise((resolve) => setTimeout(() => resolve("data"), 3000));
     console.log(
       ".load()! I should come before page"
     );
   },
   page: (ctx) => {
     console.log("page()! I should come after .load()")
     return ctx;
   },
   isLoaded: () => true,
 } 
  analytics.register(BugPlugin);
  analytics.page()
  analytics.load({ writeKey: 'MY_WRITEKEY' })
}
```

### DIFF
--- a/packages/browser/src/core/buffer/index.ts
+++ b/packages/browser/src/core/buffer/index.ts
@@ -54,7 +54,7 @@ const flushSyncAnalyticsCalls = (
   analytics: Analytics,
   buffer: PreInitMethodCallBuffer
 ): void => {
-  buffer.getCalls(name).forEach((c) => {
+  buffer.getAndRemove(name).forEach((c) => {
     // While the underlying methods are synchronous, the callAnalyticsMethod returns a promise,
     // which normalizes success and error states between async and non-async methods, with no perf penalty.
     callAnalyticsMethod(analytics, c).catch(console.error)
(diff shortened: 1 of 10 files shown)


bsneed (approved):

packages/browser/src/core/buffer/index.ts
```diff
@@ -11,6 +11,7 @@ import {
   createPageContext,
   PageContext,
 } from '../page'
+import { type } from 'os'
```
silesky, line 14: Remove

packages/browser/src/core/buffer/index.ts
```diff
@@ -65,7 +66,19 @@ export const flushAddSourceMiddleware = async (
   analytics: Analytics,
   buffer: PreInitMethodCallBuffer
 ) => {
-  for (const c of buffer.getCalls('addSourceMiddleware')) {
+  for (const c of buffer.dequeue('addSourceMiddleware')) {
```
silesky, line 69: note to self: change to getAndClear

packages/browser/src/plugins/remote-loader/index.ts
```diff
@@ -225,31 +225,36 @@ async function loadPluginFactory(
...
-  if (typeof window[remotePlugin.libraryName] === 'function') {
     // @ts-expect-error
-    return window[remotePlugin.libraryName] as PluginFactory
+    if (typeof window[remotePlugin.libraryName] === 'function') {
+      // @ts-expect-error
+      return window[remotePlugin.libraryName] as PluginFactory
+    }
+  } catch (err) {
+    console.error('Failed to create PluginFactory', remotePlugin)
+    throw err
```
silesky, line 257: The only thing was wrap the entire thing in a try/catch

danieljackins (approved):

PR #1125, opened 2024-08-14
Fix debug experience

silesky:
There was previous no way to actually debug ajs-destinations because of hash mismatches locally. This has been an issue serveral times for people who’ve had to debug important destination related code. The workaround has been to use chrome’s local overrides and paste code, which extremely manual and not intutitive.

image

This fixes that issue:

image


bsneed (approved):

danieljackins (dismissed):
omg 🔥

bsneed (approved):

PR #1146, opened 2024-09-13
Adding codecov uploading to github and buildkite with monorepo flags

MichaelGHSeg:


packages/config/src/jest/config.js
```diff
@@ -22,6 +22,9 @@ const createJestTSConfig = (
      * No need to manually run yarn build all the time.
      * This resolve packages for ts-jest so typescript compilation happens in-memory.
      */
+    collectCoverage: process.env.CI === true,
```
MichaelGHSeg, line 25: This causes some validation warnings that are wrong.  I didn't see a definitive answer about whether a new jest version would solve it. We can add --coverage to the test command lines instead

silesky:
```json
    ...([process.env.CI, process.env.COVERAGE].includes('true') && isRootConfig
      ? { collectCoverage: true }
      : {}),
    coverageDirectory: '<rootDir>/coverage',
```

I've tested this --

Observations:
- on Buildkite/CI, when `yarn run -T core test` is run -- the coverage directory in the packages/core (for the given package)

Locally, if you run `CI=true yarn test` (which is a convenience method for local development only) -- the coverage folder is created in the root.

I'm kinda on the fence about whether or not we should use the CI env variable for coverage, or whether we just create a new COVERAGE env variable. (or just, on CI, run with --coverage). Could go either way.

PR #1160, opened 2024-09-26
Adding apiHost from segment.io settings, or default

MichaelGHSeg:
Note that cdnURL has to include protocol (https://), but apiHost must not)

### DIFF
--- a/packages/browser/src/core/analytics/index.ts
+++ b/packages/browser/src/core/analytics/index.ts
@@ -95,7 +95,7 @@ export class AnalyticsInstanceSettings {
    */
   timeout = 300
 
-  constructor(settings: AnalyticsInstanceSettingsOptions, queue: EventQueue) {
+  constructor(settings: AnalyticsSettings, queue: EventQueue) {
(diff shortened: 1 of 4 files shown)


packages/browser/src/core/analytics/index.ts
```diff
@@ -201,7 +205,13 @@ export class Analytics
     super()
     const cookieOptions = options?.cookie
     const disablePersistance = options?.disableClientPersistence ?? false
-    this.settings = new AnalyticsInstanceSettings(settings)
+
+    // Extract apiHost from Segment.io settings if available
+    const segmentLoadOptions = options?.integrations?.['Segment.io'] as
+      | SegmentioSettings
+      | undefined
+    const apiHost = segmentLoadOptions?.apiHost
```
silesky, line 213: This isn't the true value -- these loadOptions are just overrides -- we would need to take into account the cdnSettings apiHost value. 

Clean option is to create a metadata? Object on the Segmentio plugin itself with the apiHost, and then having a getter here?

silesky (approved):

PR #1166, opened 2024-10-08
Signals sampling

danieljackins:

### DIFF
--- a/packages/browser/src/browser/index.ts
+++ b/packages/browser/src/browser/index.ts
@@ -122,6 +122,13 @@ export interface CDNSettings {
         version: number
       }
     | {}
+
+  /**
+   * Settings for auto instrumentation
+   */
+  autoInstrumentationSettings?: {
+    sampleRate: number
+  }
 }
 
 export interface AnalyticsBrowserSettings {
--- a/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
+++ b/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
@@ -1,5 +1,4 @@
 import { test, expect } from '@playwright/test'
-import { waitForCondition } from '../../helpers/playwright-utils'
 import { IndexPage } from './index-page'
 
 const indexPage = new IndexPage()
--- a/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
+++ b/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
@@ -12,7 +12,7 @@ test('ingestion not enabled -> will not send the signal', async ({ page }) => {
   })
 
   await indexPage.fillNameInput('John Doe')
-  const promise = await indexPage.waitForSignalsApiFlush()
+  const promise = await indexPage.waitForSignalsApiFlush().catch(() => {})
 
   await expect(promise).rejects.toThrow()
 })
--- a/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
+++ b/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
@@ -13,7 +13,7 @@ test('ingestion not enabled -> will not send the signal', async ({ page }) => {
 
   await indexPage.fillNameInput('John Doe')
   await indexPage.waitForSignalsApiFlush().catch(() => {
-    expect(true)
+    expect(true).toBe(true)
   })
 })
 
@@ -27,11 +27,5 @@ test('ingestion enabled -> will send the signal', async ({ page }) => {
     indexPage.waitForSignalsApiFlush(),
   ])
 
-  await waitForCondition(
-    () => indexPage.signalsAPI.getEvents('interaction').length > 0,
-    { errorMessage: 'No interaction signals found' }
-  )
-
-  const interactionSignals = indexPage.signalsAPI.getEvents('interaction')
-  expect(interactionSignals.length > 0)
+  expect(true).toBe(true)
 })
--- a/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
+++ b/packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
@@ -12,9 +12,9 @@ test('ingestion not enabled -> will not send the signal', async ({ page }) => {
   })
 
   await indexPage.fillNameInput('John Doe')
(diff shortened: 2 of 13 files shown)


danieljackins:
@silesky does this method of cancelling signals make sense? Should the plugin be deregistered too?

packages/signals/signals/src/core/signals/signals.ts
```diff
@@ -83,6 +83,12 @@ export class Signals implements ISignals {
    * - Registers custom signal generators.
    */
   async start(analytics: AnyAnalytics): Promise<void> {
+    if (
```
silesky, line 86: Can this live in`AnalyticsService`? i.e. have an `analyticsService.isDisabled()` ? That class is a mean to be a `facade` to abstract away cdnSettings or analytics instance related logic, exposing that to our domain -- for testing, and to keep this initialization stuff tidier.

packages/signals/signals/src/core/signals/signals.ts
```diff
@@ -83,6 +83,12 @@ export class Signals implements ISignals {
...
+      !analytics.settings.cdnSettings.autoInstrumentationSettings ||
+      Math.random() >
+        analytics.settings.cdnSettings.autoInstrumentationSettings.sampleRate
```
silesky, line 89: Hey, I don't understand this Math.random() thing, maybe you could explain? I assumed we would just do:
`analytics.settings.cdnSettings.autoInstrumentationSettings.sampleRate === 0`?

silesky (changes requested):

packages/browser/src/browser/index.ts
```diff
@@ -122,6 +122,15 @@ export interface CDNSettings {
...
     | {}
+  /**
+   * Settings for auto instrumentation
+   */
+  autoInstrumentationSettings?: {
+    /**
+     * A sample rate to determine what percentage of sessions should send signals, from [0,1]
+     */
+    sampleRate: number
+  }
```
silesky, line 133: We'd need a changeset for both signals and analytics =)

packages/signals/signals/src/core/signals/signals.ts
danieljackins: It's a sample rate from [0,1] that we "roll against" to see if we should send signals or not (not just a boolean in the settings)

silesky:
@danieljackins can we have a test / maybe an integration test for this?

packages/signals/signals/src/core/signals/signals.ts
silesky: If the sample rate is 0, we just want to stop from sending signals -- we don't want to stop processing them, right?
This will stop the entire signals initialization. So this really only implies to the signal collection step?

packages/signals/signals/src/core/signals/signals.ts
silesky: Understanding this feature more, we should have analyticsServie.sampleRate and we should probably have NetworkConfig have a "signalCollectionDisabled": boolean

packages/signals/signals/src/core/signals/settings.ts
```diff
@@ -139,3 +150,54 @@ class SignalRedactionSettings {
...
 }
+
+class SignalIngestionSettings {
+  private static ingestionKey = 'segment_signals_debug_ingestion_enabled'
+  constructor(initialValue?: boolean) {
+    if (typeof initialValue === 'boolean') {
+      this.setSignalIngestion(initialValue)
+    }
+
+    // setting ?segment_signals_debug=true will disable redaction, and set a key in local storage
```
silesky, line 161: Can probably delete this comment?

packages/signals/signals/src/core/signals/settings.ts
```diff
@@ -139,3 +150,54 @@ class SignalRedactionSettings {
...
+    // this setting will persist across page loads (even if there is no query string)
+    // in order to clear the setting, user must set ?segment_signals_debug=false
+    const debugModeInQs = parseDebugModeQueryString()
+    logger.debug('debugMode is set to true via query string')
```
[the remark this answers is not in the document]
danieljackins: yeah sorry I'm going to refactor the whole debug part still

packages/signals/signals/src/core/client/index.ts
```diff
@@ -8,13 +8,17 @@ export class SignalsIngestSettings {
...
   writeKey?: string
+  sampleRate: number
   constructor(settings: SignalsIngestSettingsConfig) {
     this.flushAt = settings.flushAt ?? 5
     this.apiHost = settings.apiHost ?? 'signals.segment.io/v1'
     this.flushInterval = settings.flushInterval ?? 2000
     this.shouldDisableSignalRedaction =
       settings.shouldDisableSignalRedaction ?? (() => false)
+    this.signalIngestion = settings.signalIngestion ?? (() => false)
+    this.sampleRate = settings.sampleRate ?? 0
```
silesky, line 21: I'd rather we not expose sampleRate publicly, as it's strictly an implementation detail -- IMO, you shouldn't be able to customize the sample rate as a setting. Can we abstract this way completely? in my head the `signalsIngestion` was the only data point we want

packages/signals/signals/src/core/client/index.ts
```diff
@@ -69,10 +75,25 @@ export class SignalsIngestClient {
     return analytics
   }
 
+  private shouldIngestSignals(): boolean {
```
silesky, line 78: I don't think we don't want to calculate this on a per event basis -- see comment above.

silesky (changes requested):

packages/signals/signals/src/core/signals/settings.ts
```diff
@@ -79,58 +93,80 @@ export class SignalGlobalSettings {
      * Add new URLs to the disallow list
      */
     disallowListURLs: (string | undefined)[]
+    /**
+     * Sample rate to determine sending signals
+     */
```
silesky, line 98: Can we make this | undefined / optional, so update can be used atomically (one day?)

packages/signals/signals-integration-tests/src/tests/signals-vanilla/signals-ingestion.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    enableSignalsIngestion: true,
+  })
+
+  await Promise.all([
+    indexPage.fillNameInput('John Doe'),
+    indexPage.waitForSignalsApiFlush(),
+  ])
+
+  await waitForCondition(
+    () => indexPage.signalsAPI.getEvents('interaction').length > 0,
```
silesky, line 31: I don't think this is and everything below it is needed -- as long as signals api has flushed, test is good!

PR #1189, opened 2024-11-13
Added missing metric from older analytics packaged

MichaelGHSeg:
We were missing a metric that's recorded by the older analytics package

- [X] I've included a changeset


silesky (dismissed):
A test is failing (easy fix), but looks good!

silesky (approved):

PR #1196, opened 2024-11-22
Add coverage upload to gh actions ci.yml

MichaelGHSeg:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -41,8 +41,6 @@ steps:
       - echo "--- Install dependencies"
       - yarn install --immutable
       - yarn run -T browser exec make ci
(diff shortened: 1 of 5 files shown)


.buildkite/pipeline.yml
```diff
@@ -30,6 +30,9 @@ steps:
 
   - label: '[Browser] Lint + Test'
     key: browser-lint-test
```
silesky, line 32: Not sure if incomplete, but we probably need to set COVERAGE=true as well?

MichaelGHSeg: Whoops.  Had thought that was set at the top of the file.

silesky (dismissed):

PR #1220, opened 2025-01-31
Add signals middleware

silesky:
- Allow registration of middleware to allow for dropping and modification of signals
- Refactor to use middleware for network filter, rather than having that logic live in the network generator. 

```ts
class MyMiddleware implements SignalsMiddleware {
 load(ctx: SignalsMiddlewareContext) {
   // this will block any signals from being emitted until resolved, like a segment 'before' plugin -- it will be invoked before processSignal
   // we can access ctx.unstableGlobalSettings from inside this plugin and assign it to the instance if needed
    return 
 }
  process(signal: Signal) {
    if (
      signal.type === 'network' &&
      signal.data.action === 'request' &&
      ...
    ) {
      // null = drop signal
      return null
    } else {
      return signal
    }
  }
}
const signalsPlugin = new SignalsPlugin({
  middleware: [new MyMiddleware()]
})
```

### DIFF
--- a/.changeset/warm-lies-rush.md
+++ b/.changeset/warm-lies-rush.md
@@ -1,0 +1,23 @@
+---
+'@segment/analytics-signals': minor
+---
+
+Allow registration of middleware
+
+```ts
+class MyMiddleware implements SignalsMiddleware {
+  process(signal: Signal) {
+    if (
+      signal.type === 'network' &&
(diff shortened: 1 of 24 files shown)


bsneed:
This should come up in the parity meeting.

bsneed (approved):

PR #1262, opened 2025-03-17
Allow the request "credentials" option to be configured

vbetivu-ellation:

Description
---
The segment library does not allow the `credentials` option on the request to be configured, that limits the possibility to send cookies with cross-origin requests (ex: website lives on `www.awesome.com` and the Segment proxy lives on `segment.awesome.com`). This PR adds the possibility to configure the `credentials` option as part of the `deliveryStrategy` `config`.

How it was tested
---
1. The library was built and referenced locally.
2. The library was tested with and without the `credentials` option. 
<details>
<summary>Here are some examples:</summary>
image
image
image
image
image
image
image
</details>

### DIFF
--- a/.changeset/plenty-camels-accept.md
+++ b/.changeset/plenty-camels-accept.md
@@ -2,4 +2,4 @@
 '@segment/analytics-next': minor
 ---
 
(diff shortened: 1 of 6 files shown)


silesky (approved):
LGTM

PR #1276, opened 2025-04-15
Check attempts before dispatching to segment

danieljackins:
This PR changes the retry behavior for Segment.io by checking the number of attempts before it attempts to send the event to Segment, which was causing an extra retry.

Tested locally by verifying that a cancelled event was not retried when retryQueue was disabled, and 4 total attempts were made when retryQueue was enabled.

### DIFF
--- a/packages/browser/src/plugins/segmentio/__tests__/retries.test.ts
+++ b/packages/browser/src/plugins/segmentio/__tests__/retries.test.ts
@@ -36,7 +36,7 @@ describe('Segment.io retries 500s and 429', () => {
     const ctx = await analytics.track('event')
     jest.runAllTimers()
 
-    expect(ctx.attempts).toBeGreaterThanOrEqual(3) // Gets incremented after use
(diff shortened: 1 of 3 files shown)


silesky (dismissed):

packages/browser/src/plugins/segmentio/__tests__/retries.test.ts
```diff
@@ -131,3 +134,56 @@ describe('Batches retry 500s and 429', () => {
     expect(fetch.mock.lastCall[1].body).toContain('"retryCount":2')
   })
 })
+
+describe('retryQueue', () => {
+  let options: SegmentioSettings
+  let analytics: Analytics
+  let segment: Plugin
+  beforeEach(async () => {
+    jest.useRealTimers()
```
silesky, line 143: I generally just just make this at afterEach -- and beforeEach can just have 'jest.useFakeTimers()`

silesky (approved):

PR #1278, opened 2025-04-17
Update documentation for expo-web example

steveacalabro:
I was working to try to get segment working with my expo 52 react native project. I noticed that it would pass the metro build and then throw an error at runtime

```
TypeError: Class extends value undefined is not a constructor or null
```

I happened to create this custom transformer as a fix for a custom component library package that I have. I decided to give it a try since everything else I did wasn't working and to my surprise it worked.

I figured others might be fighting this in the future so I decided to update the docs in case folks can get saved by my solution. 

I didn't run a changeset as a part of this PR since I only updated a markdown file. Please let me know if you still want one. Thanks :)

### DIFF
--- a/packages/browser/README.md
+++ b/packages/browser/README.md
@@ -145,6 +145,41 @@ export default defineComponent({
 </script>
 ```
 
+### Expo Web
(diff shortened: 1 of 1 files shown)


steveacalabro:
@libraries-web-team Any thoughts here?

PR #1282, opened 2025-04-26
feat(core): adds flag to allow `analytics.ready()` to run even if plugin fails on ready

rmachado-studocu:
#### What does this PR do

* Adds 2 new options for `load`:
	* `ignorePluginReadyError` - opt-in flag that runs the new logic described below
	* `onPluginReadyError` - error handler for plugins that error in the `ready` stage

When `ignorePluginReadyError` is `true`, the code will make use of `Promise.allSettled()` to capture all promise results and only run the callback with the `fulfilled` promises.
If `onPluginReadyError` is provided, this function will be used as `.catch()` of the `plugin.ready` promises.

- [X] I've included a changeset (psst. run `yarn changeset`. Read about changesets [here](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)).


rmachado-studocu:
An alternative to this implementation could be to instead implement a new function for this. Example:

```tsx
readySettled: (cb, onPluginReadyError) => {
	// here would be the Promise.allSettled()
}
```

rmachado-studocu:
Hi @silesky and @MichaelGHSeg (sorry for the tags),

Is this something you'd see being implemented or no? Just to understand how to proceed with this feature.

packages/browser/src/core/analytics/index.ts
```diff
@@ -534,6 +534,24 @@ export class Analytics
...
+          if (result.status === 'fulfilled') {
+            return result.value
+          }
+        })
+        callback(res)
+        return res
+      })
+    }
+
     return Promise.all(
```
silesky, line 555: Doing an allSettled and then an .all here like duplicate work -- Can we refactor to just use allSettled?

rmachado-studocu: Will do!

rmachado-studocu: [6b0d9fe](https://github.com/segmentio/analytics-next/pull/1282/commits/6b0d9feb72c97a6f5d1d8152ec861f79ab1f8b93)

PR #1284, opened 2025-04-29
Add signals strategy configuration for customers with restrictive CSPs

silesky:
- Raises the max signal buffer to 100.
- Add new  setting:
```ts
const plugin = new SignalsPlugin({ sandboxStrategy: 'global' })
 ```
 
 
 
### Background

This fixes CSP errors: `unsafe-inline` and `unsafe-eval` without requiring customers to add these to their CSP directive. These occur because, ironically, the default safe strategy of injecting js as a string into an iframe-web-worker and evaluating is not allowed because it's dynamic, so this setting essentially removes the sandbox. 

The iframe sandbox is mainly needed for "vanilla" function scenarios where people can introduce infinite loops, etc -- not UI-driven auto-instrumentation scenarios, so global is probably going to become the default.

There is a feature called `Shadow Realms` that is coming to JS, that one day we will be able to use without upsetting CSPs.

## TLDR
The following CSP:
```
<meta http-equiv="Content-Security-Policy" content="script-src 'self' https://*.segment.com https://*.googletagmanager.com blob:">
```
Produces the following error:
image

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -84,6 +84,7 @@ steps:
       - yarn install --immutable
       - echo "+++ Run Browser integration tests :pray:"
       - yarn turbo run --filter=@internal/browser-integration-tests test:int
+      - yarn turbo run --filter=@internal/browser-integration-tests test:int
     retry:
       automatic:
         - exit_status: '*'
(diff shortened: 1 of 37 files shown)


alanjcharles (dismissed):

alanjcharles (approved):
lgtm!

Issue #1339, opened 2026-02-05
[Bug]: middleware chunk name collision when using nodejs middleware runtime in NextJS with @segment/analytics-next ^1.77.0


igordonin:
# Summary

When using NextJS 15.5.12 and configuring middleware.ts to use NodeJS runtime instead of Edge, there's a chunk name collision with `middleware`

# How to reproduce

Configure any minimal middleware in `middleware.ts` and add @segment/analytics-next:1.77.0 or above to the dependencies. 

```
export const config = {
  matcher: ...
  runtime: 'nodejs',
};
```

Error when building the application:

```
Failed to compile.
../node_modules/.pnpm/@segment+analytics-next@1.81.1_encoding@0.1.13/node_modules/@segment/analytics-next/dist/pkg/core/analytics/index.js
It's not allowed to load an initial chunk on demand. The chunk name "middleware" is already used by an entrypoint.
```

igordonin:
Haven't tested it, but this seems to fix it indeed: https://github.com/segmentio/analytics-next/pull/1357

Thanks

PR #1355, opened 2026-04-14
fix(analytics): use more unique chunk name for middleware

blurrah:
Changes the source middleware chunk name as the current name breaks Next.js with node runtime using a `middleware.js`, see #1339 


chenxzhang:
will open a separate pr to make the change

